PR #7, opened 2014-11-09
Keep line-breaks in descriptions

Sophist-UK:
Some descriptions contain &lt;pre&gt; sections - so we need to keep the line breaks in descriptions so that &lt;pre&gt; sections display correctly.

I have tested that the JSON is created as expected, but cannot test that the JSON will display in the wiki as expected.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -40,7 +40,7 @@ def get_data(filepath):
             if 'author' not in data:
                 author = re.match(re_author, line)
                 if author:
-                    data['author'] = author.group(2)
+                    data['author'] = re.sub(r'([^,])[ ]', r'\1&nbsp;', author.group(2))
 
             if 'description' not in data:
                 if re.match(re_desc_start, line):
--- a/generate.py
+++ b/generate.py
@@ -47,7 +47,7 @@ def get_data(filepath):
                     desc_flag = True
                 elif re.match(re_desc_end, line):
                     desc_flag = False
-                    desc = re.match(re_desc, re.sub(r'[\\\n]', '', "".join(desc_lines)))
+                    desc = re.match(re_desc, re.sub(r'[\\\n]', '', "\r".join(map(lambda s: s.strip(), desc_lines))))
                     if desc:
                         data['description'] = desc.group(2)
 


generate.py
```diff
@@ -40,14 +40,14 @@ def get_data(filepath):
             if 'author' not in data:
                 author = re.match(re_author, line)
                 if author:
-                    data['author'] = author.group(2)
+                    data['author'] = re.sub(r'([^,])[ ]', r'\1&nbsp;', author.group(2))
```
zas: This change causes an issue on the website:

capture du 2015-01-31 11 09 01

Would you mind to provide a fix ?

phw: I would just revert this line, what's the idea here? What this line does is that it will any space character not prceded by a comma with a `&nbsp;`, but why? I can see that it makes sense to allow HTML in the description, but why should the author field contain HTML? That doesn't make much sense to me, it even looks wrong to me having the HTML encoded characters in the JSON.

```
"discnumber": {
      "api_version": "0.9.0\", \"0.10\", \"0.15", 
      "author": "Lukas&nbsp;Lalinsky", 
      "description": "Moves disc numbers and subtitles....", 
      "downloads": 0, 
      "files": {
        "discnumber.py": "324fa53c1a81386fae1840430f95fb11"
      }, 
      "name": "Disc Numbers", 
      "version": "0.1"
    }, 
```

The other possible fix is on the website to treat the author field as HTML and not HTML encode it on display.

By the way the second character set in the regex, the `[ ]`, is superfluous. There is no need for a character set when there is only a single character in it, so `'([^,]) '` would be fine.

Sophist-UK: The purpose of this was to prevent names being split across lines like this:
```
RifRaf, Lukas
Lalinsky, Brian
Toth
```
But I don't know how to set up a test environment to be able to test this, and hadn't realised that the output would be html escaped.

We can revert this line if we can't easily stop the html escaping (or if that would open us to CSS vulnerabilities).

phw: > The purpose of this was to prevent names being split across lines like this:

For me this is really a display issue, that should be solved on the display side (the website) and not inside the JSON. I think the purpose of the JSON is to provide the data (and not the formatting) and eventually might be reused in different ways (e.g. Picard plugin downloader). The description is somewhat special as it surely is meant to have formatting included.

If we want to structure this better it would make more sense to store an array of authors in the JSON (and probably also in the plugins themselves, but that would be a breaking change). But for now I would just revert this change and eventually implement this very same logic on the website.

If you want to give it a go, the website is pretty easy to setup for local development, have a look at the documentation.

zas: > For me this is really a display issue, that should be solved on the display side (the website) and not inside the JSON.

Yes, exactly. Formatting shouldn't be done here.

Sophist-UK: Since this PR is merged / closed, someone needs to revert this line.

I can do it, but won't have time for a day or two.

zas: > Since this PR is merged / closed, someone needs to revert this line.
> 
> I can do it, but won't have time for a day or two.

Done in https://github.com/musicbrainz/picard-plugins/commit/1b529822496450da4563e332556b194f12f0c71f

zas: Website-side equivalent solution proposed in https://github.com/musicbrainz/picard-website/pull/23

Issue #5, opened 2014-11-07
No license information


warpr:
Many plugins in this repository have no information on what license they can be used and redistributed under.  Preferably that information should be included in the generated `plugins.json`.

See also http://tickets.musicbrainz.org/browse/PICARD-326

phw:
We have started to add standardized license variables to the plugins. It's now added for those plugins where it was easy to do.

PR #8, opened 2015-01-05
Created two new plugins

dstrohl:
one to remove and save anything that is at 80+% accurate, the other to remove anything that is less than 80% (without saving)

In both cases they will also remove albums that are empty (leaving ones that still have unmatched files).

both have a number of debug messages in them, which can be removed it it makes sense.

### DIFF
--- a/plugins/remove_80_plus/save_80_plus.py
+++ b/plugins/remove_80_plus/save_80_plus.py
@@ -1,0 +1,69 @@
+PLUGIN_NAME = u'Save good files'
+PLUGIN_AUTHOR = u'Dan Strohl'
+PLUGIN_DESCRIPTION = u'''Save all files at more than 80+ percent.'''
+PLUGIN_VERSION = '0.4'
+PLUGIN_API_VERSIONS = ['1.3.0']
+
+from picard.album import Album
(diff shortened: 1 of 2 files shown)


phw:
You should add some source code license information to your code. We already have too many plugins which don't specify a license, which makes it hard to distribute and reuse the code. See issue #5 .

Otherwise I haven't looked too close into the code, but I like the functionality.

zas:
Imho it would be better to create 2 separated PRs for those plugins, with explicit title for each.
And, as said by @phw , please add : 
```python 
PLUGIN_LICENSE = "your license"
PLUGIN_LICENSE_URL = "your license url (if any)"
```
Format isn't yet normalized, but i think it is a good start (see http://tickets.musicbrainz.org/browse/PICARD-703)

zas:
@dstrohl ?

dstrohl:
sorry, I missed the notification of these.. I will take a look and add the license info and update the pull requests.

zas:
> sorry, I missed the notification of these.. I will take a look and add the license info and update the pull requests.

@dstrohl : ping ? ;)

zas:
No update from author, closing.

PR #9, opened 2015-01-09
New plugin: AcousticBrainz

Freso:
This is a simple plugin which has a number of small problems though. However, it seems that only @ariscop (the author of the plugin) and I have actually poked at it and played around with it, so [we agreed to make a pull request for it](http://chatlogs.musicbrainz.org/musicbrainz/2015/2015-01/2015-01-09.html#T21-32-26-265982) to get some more eyeballs on it.


plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -0,0 +1,48 @@
...
+def result(album, metadata, release, track, data, reply, error):
+    moods = []
+    genres = []
+    bpm = False
+    try:
+        data = loads(data)["highlevel"]
+        for k, v in data.items():
+            if k.startswith("genre_") and not v["value"].startswith("not_"):
+                genres.append(v["value"])
+            if k.startswith("mood_") and not v["value"].startswith("not_"):
```
zas: I would make this more generic, using prefixes as keys, something like (untested, just an example):
```python
prefixes = {
'genre': 'genre_',
'mood':  'mood_',
}

for k, v in data.items():
   for pk, pv in prefixes.items():
        if k.startswith(pv) and not v["value"].startswith("not_"):
             metadata.setdefault(pk, []).append(v["value"])
```

phw:
You should add a source code license for this plugin. We have already too many plugins that don't specify a license and an open bug report about this (which I doubt we can really fix for all existing ones).

Otherwise this looks good to me, except that I support zas' suggestion.

zas:
> You should add a source code license for this plugin.

It is perhaps time to define a normalized way to add license info to plugins.

What about adding something like
```python
PLUGIN_LICENSE = "text or url"
```
to the plugin API for 1.4 ?

Sophist-UK:
I agree - we need a standard way to state license i.e. 
```
PLUGIN_LICENSE = "GPL v2"
```

a. To try to keep it simple for users (i.e. minimise the number of different licenses used for plugins), I think we should encourage plugin writers to use a specific recommended license (like GPL v2).

b. I wondered whether we should require plugins hosted on MB to be licensed using e.g. GPL, but on reflection decided that was perhaps somewhat limiting. However, we could add a Ts&Cs statement that (after say 1 July 2015) plugins listed / hosted on MB without a stated license will default to GPL v2, and we can be proactive in contacting authors to get them to add the license, possibly (though this needs careful thought) adding the default license ourselves on 1 July if the author hasn't done it themselves.

Just a thought...

zas:
http://tickets.musicbrainz.org/browse/PICARD-703

zas:
I proposed to add 2 new variables, `PLUGIN_LICENSE` and `PLUGIN_LICENSE_URL`, even if we decide to do another way, you can add them to this plugin.

phw:
> I wondered whether we should require plugins hosted on MB to be licensed using e.g. GPL, but on reflection decided that was perhaps somewhat limiting. However, we could add a Ts&Cs statement that (after say 1 July 2015) plugins listed / hosted on MB without a stated license will default to GPL v2, and we can be proactive in contacting authors to get them to add the license, possibly (though this needs careful thought) adding the default license ourselves on 1 July if the author hasn't done it themselves.

We are a bit off-topic here, but I wanted to add my opinion on those very important points. First I think the GPL v2 is the natural choice, since this is the license of Picard. I am not sure if a strict interpretation of the GPL allows much variation in the plugins, but I think from the developers point of view we are fine with any license that is free and allows us to redistribute the plugin. However, just changing the license for existing plugins may not be so easy and requires finding the original author.

zas:
@Freso , @ariscop : any chance to have comments addressed so we can merge it ?

zas:
@ariscop : travis-ci build failed, just a python 3 syntax issue, but it would be nice if it was fixed (well, not that Picard supports Python 3 yet, but i wish we keep plugins compatible with Python 2.6, 2.7, 3.3, 3.4 syntax at least, it will ease the move later, and doesn't imply too much work). In this case, 'as' instead of ',' should be enough.

plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -0,0 +1,70 @@
...
+from picard import webservice
+from picard.metadata import register_track_metadata_processor
+
+ACOUSTICBRAINZ_HOST = "acousticbrainz.org"
+ACOUSTICBRAINZ_PORT = 80
+
+webservice.REQUEST_DELAY[(ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT)] = 100
+
+def result(album, metadata, release, track, data, reply, error):
+    moods = []
```
Sophist-UK: No error handling.

plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -0,0 +1,70 @@
...
+            if k.startswith("mood_") and not v["value"].startswith("not_"):
+                moods.append(v["value"])
+        metadata["genre"] = genres
+        metadata["mood"] = moods
+        log.debug("%s: Track %s (%s) Parsed response (genres: %s, moods: %s)", PLUGIN_NAME, metadata["musicbrainz_recordingid"], metadata["title"], str(genres), str(moods))
+    except Exception as e:
+        log.error("%s: Track %s (%s) Error parsing response: %s", PLUGIN_NAME, metadata["musicbrainz_recordingid"], metadata["title"], str(e))
+    finally:
+        album._requests -= 1
+        album._finalize_loading(None)
```
Sophist-UK: You should only call _finalize_loading if _requests == 0.

plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -0,0 +1,70 @@
...
+
+def process_track(album, metadata, release, track):
+    album.tagger.xmlws.download(
```
Sophist-UK: Are we sure that all MB release tracks have a recording-id?

plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -0,0 +1,70 @@
...
+        ACOUSTICBRAINZ_HOST,
+        ACOUSTICBRAINZ_PORT,
+        u"/%s/high-level" % (metadata["musicbrainz_recordingid"]),
+        partial(result, album, metadata, release, track),
```
Sophist-UK: Why are we passing release and track when they are never used by the result function?

plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -0,0 +1,70 @@
...
+from json import loads
+from functools import partial
+from picard import log
+from picard import webservice
+from picard.metadata import register_track_metadata_processor
+
+ACOUSTICBRAINZ_HOST = "acousticbrainz.org"
+ACOUSTICBRAINZ_PORT = 80
+
+webservice.REQUEST_DELAY[(ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT)] = 100
```
Sophist-UK: This would be better written as:
```
from webservice import REQUEST_DELAY
REQUEST_DELAY[(ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT)] = 100
```

Also if we assume that MB provides release information at 1 per second, and that there are on average 20 tracks on an album, then I would suggest that you use 50ms as the delay so that the AB calls are processed at approximately the same rate that they will be created.

Sophist-UK:
I have renamed my AcousticBrainz plugin at #41 AcousticBrainz Tonal-Rhythm in order to avoid conflicts with this one.

Can I suggest that this be named AcousticBrainz Mood-Genre to assist users in differentiating between them?

plugins/acousticbrainz/acousticbrainz.py
mineo: No, the second thing finalize_loading does is exactly that check.

plugins/acousticbrainz/acousticbrainz.py
mineo: Yes.

plugins/acousticbrainz/acousticbrainz.py
Sophist-UK: We should probably provide standardised routines for this rather than have people accessing variables.

We can also use this to note if there is an error in any of the webservice calls and if so flag it with an icon or similar.

ariscop:
@Sophist-UK Might be better to merge the two

Sophist-UK:
They make different calls to AcousticBrainz.
#9 makes a call for high-level data.
#41 makes a call for low-level data.

So merging would result in double the number of calls - hence my suggestion to keep them separate and rename them appropriately.

ariscop:
@Sophist-UK Yes but imo that would be better as a setting in one plugin

I'll rename it for now

ariscop:
Plus automating submission for tracks without acousticbrainz data would be really nice, i'll play around with merging the plugins and adding that

Sophist-UK:
Yes - I agree, settings would be better - it would also allow for selecting specific tags.

It might also be possible to include the ability to submit data, but (having tried it) the submission tools is very CPU intensive and automated submission would be very time-consuming for a large set of files PLUS you would not want to run it on files which had already been submitted.

ariscop:
[My attempt at a merge](https://github.com/ariscop/picard-plugins/blob/acousticbrainz-merge/plugins/acousticbrainz/acousticbrainz.py)
Never used PyQT4 (or QT tbh), so no ui yet

ariscop:
Could just expose acousticbrainz data to the tagger script, at least initially

PR #13, opened 2015-01-30
Save and rewrite header

NCenerar:
This is a simple plugin that adds a context menu action to save the files and rewrite their header. It is usefull when there is a padding in the header (http://tickets.musicbrainz.org/browse/PICARD-695).


zas:
Please add license info, see http://tickets.musicbrainz.org/browse/PICARD-703

NCenerar:
I'm no expert in licensing. I hope I did it the right way...

NCenerar:
Ok, so, I picked up the GNU GPL 3... I hope it is good...

PR #23, opened 2015-02-16
soundtrack: add new plugin

Ram-Z:


plugins/soundtrack/soundtrack.py
```diff
@@ -0,0 +1,24 @@
...
+# Copyright © 2015 Samir Benmendil <me@rmz.io>
+# This work is free. You can redistribute it and/or modify it under the
+# terms of the Do What The Fuck You Want To Public License, Version 2,
+# as published by Sam Hocevar. See http://www.wtfpl.net/ for more details.
+
+PLUGIN_NAME = 'Soundtrack'
+PLUGIN_AUTHOR = 'Samir Benmendil'
+PLUGIN_LICENSE = 'WTFPL'
+PLUGIN_LICENSE_URL = 'http://www.wtfpl.net/'
+PLUGIN_DESCRIPTION = '''Sets the albumartist to <em>"Soundtrack"</em> if releasetype is a soundtrack.'''
```
zas: Please don't use tags in plugin description, it is used by various tools, some of them will not handle '<em>'.

Ram-Z: This was inspired by other plugins, which use them. I wasn't too keen to put them in in the first place, but added them for consistency.

Ram-Z:
Is there any documentation on what plugin API version supports what? The version I put down was chosen randomly, but I'm sure older API version will work as well.

PR #29, opened 2015-08-20
Keep the folder structure for multi-plugin files in the zipfile

mineo:
Fixes PW-19

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -140,11 +140,26 @@ def zip_files():
         archive = zipfile.ZipFile(archive_path + ".zip", "w")
 
         dirpath = os.path.join(plugin_dir, dirname)
+        plugin_files = []
+
         for root, dirs, filenames in os.walk(dirpath):
             for filename in filenames:
                 file_path = os.path.join(root, filename)
-                archive.write(file_path,
-                              file_path.split(os.path.join(dirpath, ''))[1],
+                plugin_files.append(file_path)
+
+        if len(plugin_files) == 1:
+            # There's only one file, put it directly into the zipfile
+            archive.write(plugin_files[0],
+                          os.path.basename(plugin_files[0]),
+                          compress_type=zipfile.ZIP_DEFLATED)
+        else:
+            for filename in plugin_files:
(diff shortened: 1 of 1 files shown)


generate.py
```diff
@@ -140,11 +140,26 @@ def zip_files():
...
+        plugin_files = []
+
         for root, dirs, filenames in os.walk(dirpath):
             for filename in filenames:
                 file_path = os.path.join(root, filename)
-                archive.write(file_path,
-                              file_path.split(os.path.join(dirpath, ''))[1],
+                plugin_files.append(file_path)
+
+        if len(plugin_files) == 1:
```
zas: I wonder about that. I think we should always have a directory, it would make things simpler.

mineo: Hallo Laurent Monin:
> >          for root, dirs, filenames in os.walk(dirpath):
> >              for filename in filenames:
> >                  file_path = os.path.join(root, filename)
> > -                archive.write(file_path,
> > -                              file_path.split(os.path.join(dirpath, ''))[1],
> > +                plugin_files.append(file_path)
> > +
> > +        if len(plugin_files) == 1:
> 
> I wonder about that. I think we should always have a directory, it would make things simpler.

Does importing a plugin in a file that's not called __init__.py in an
extra folder work? I thought it doesn't and that's why I made it so only
multi-file plugins get put into a directory - you can now drop
whatever's in the .zip directly into the plugins folder and it just
works.

zas: > Does importing a plugin in a file that's not called __init__.py in an extra folder work? I thought it doesn't and that's why I made it so only multi-file plugins get put into a directory - you can now drop whatever's in the .zip directly into the plugins folder and it just works.

Hmmm, yes, you're right.
I'll look at that in details, perhaps we could change that so it is simpler: one directory per plugin, looking for <name_of_the_dir>.py if there is no __init__.py (and perhaps later enforce dir/__init__.py for all plugins)

zas:
@mineo : anyway, the proposed PR is ok for me as is.

PR #31, opened 2015-08-22
Improve plugin metadata parser and simplify generate.py

zas:
Replace regex-based parser by one based on ast
Remove extra code related to the historical 'downloads' counter

It fixes https://github.com/musicbrainz/picard-plugins/issues/30

### DIFF
--- a/get_plugin_data.py
+++ b/get_plugin_data.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+import ast
+
+known_data = [
+    'PLUGIN_NAME',
+    'PLUGIN_AUTHOR',
+    'PLUGIN_VERSION',
+    'PLUGIN_API_VERSIONS',
+    'PLUGIN_LICENSE',
+    'PLUGIN_LICENSE_URL',
+    'PLUGIN_DESCRIPTION',
+]
+
+
+def get_plugin_data(filepath):
+    """Parse a python file and return a dict with plugin metadata"""
+    data = {}
+    with open(filepath, 'r') as plugin_file:
+        source = plugin_file.read()
+        try:
+            root = ast.parse(source, filepath)
+        except Exception as e:
+            print "Cannot parse " + filepath
+            raise e
+        for node in ast.iter_child_nodes(root):
+            if isinstance(node, ast.Assign) and len(node.targets) == 1:
+                target = node.targets[0]
+                if (isinstance(target, ast.Name)
+                    and isinstance(target.ctx, ast.Store)
+                    and target.id in known_data):
+                    name = target.id.replace('PLUGIN_', '', 1).lower()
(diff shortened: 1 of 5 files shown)


zas:
I have few questions about `api_version`:
- about name, why `api_version` instead of `api_versions` ?
- the old code was generating weird list:
`"api_version": "0.15.0\", \"0.15.1\", \"0.16.0\", \"1.0.0\", \"1.1.0\", \"1.2.0\", \"1.3.0",`
- new code generate a list as a string without all those useless escaped double quotes to please tests:
`"api_version": "0.15.0, 0.15.1, 0.16.0, 1.0.0, 1.1.0, 1.2.0, 1.3.0",`
- what about having a javascript list of strings here instead ?

dufferzafar:
Wow! I totally didn't know about the `ast`module. The name `api_version` was probably an error on my part. It should be `api_versions`. And yeah, a js list of strings would be better, I guess.

zas:
> Wow! I totally didn't know about the astmodule. The name api_version was probably an error on my part. It should be api_versions. And yeah, a js list of strings would be better, I guess.

I wasn't aware of it either since today ;)
I think we can change `api_version` to `api_versions` without much harm since it is not used by picard website anyway, but if someone is using the plugins api and this field ... i don't know if anyone uses it but picard website at the moment. What do you think guys ?

dufferzafar:
Um, I don't know anyone using it either, and even if they do, I think this will be a trivial issue to fix on their side too.

zas:
I made the changes about `api_versions`.
I'm for merging it as is, and let users (if any...) fix things on their side. Let's wait to see what others think about it.

mwiencek:
The format of the generated JSON was never a documented API, it's fine to break it. :shipit:

get_plugin_data.py
```diff
@@ -0,0 +1,39 @@
...
+                if (isinstance(target, ast.Name)
+                    and isinstance(target.ctx, ast.Store)
+                    and target.id in known_data):
+                    name = target.id.replace('PLUGIN_', '', 1).lower()
+                    if name not in data:
+                        try:
+                            data[name] = ast.literal_eval(node.value)
+                        except ValueError as e:
+                            print filepath + ':' + ast.dump(node)
+                            pass
```
mineo: Can the website actually handle missing keys in the JSON data?

zas: > Can the website actually handle missing keys in the JSON data?

Website only uses `name`, `version`, `description`, `author`.
And plugins tests check for `name`, `api_versions`, `author`, `description`.
I think test could be ensure that `version` is present too.

zas: I added test for `version` in 1d27cefa0cdca2cabd248d27e8b0466231c11ece

PR #32, opened 2015-08-22
Minor changes for tests to work also in python 3 environment

zas:


plugins/abbreviate_artistsort/abbreviate_artistsort.py
```diff
@@ -25,7 +25,7 @@
 PLUGIN_LICENSE = "GPL-2.0"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
 
-
+from __future__ import printfunction
```
mineo: Missing _

plugins/abbreviate_artistsort/abbreviate_artistsort.py
```diff
@@ -89,10 +89,11 @@ def abbreviate_artistsort(tagger, metadata, track, release):
         for i in range(0,min(len(sorts),len(unsorts))):
             sort = sorts[i]
             if _debug_level > 1:
-                print "%s: Trying to abbreviate '%s'." % (PLUGIN_NAME, sort)
```
mineo: While we're touching this: why not replace all the print statements with `picard.log.info` (or something else from picard.log)?

zas:
I'm actually unsure about this anyway, the purpose was to make the tests pass on 3.x too, but it seems to me than since Picard itself hardly works on anything but 2.x, it doesnt make much sense.
I guess we could just close this.

plugins/abbreviate_artistsort/abbreviate_artistsort.py
zas: > While we're touching this: why not replace all the print statements with picard.log.info (or something else from picard.log)?

It would make sense for plugins.
We should perhaps provide few guidelines on how to write plugins (about mandatory/optionnal metadata, api versions, and helper functions).

For generate/test code shouldn't depend on picard.

mwiencek:
Well, I think it's a good patch, I imagine we'll have to move to python 3 eventually. :)

zas:
> Well, I think it's a good patch, I imagine we'll have to move to python 3 eventually. :)

I committed and pushed commits to master, that includes those changes plus few others.
Tests are now working on 2.6, 2.7, 3.3 and 3.4, 3.2 still failing because of lack of support for u"" strings (it would require much more changes, so i disabled it).

PR #35, opened 2015-09-22
Plugin to calculate BPM from a file

madmouser1:
A plugin to calculate BPM from a file with a options page.


plugins/bpm/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+
+class BMPOptionsPage(OptionsPage):
+
+    NAME = "bpm"
+    TITLE = "BPM"
+    PARENT = "plugins"
+
+    options = [
+        TextOption("setting", "bpm_win_s_parameter", "1024"),
```
zas: You should use `IntOption()` here instead imho

plugins/bpm/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+    def load(self):
+        self.ui.win_s_parameter.setText(self.config.setting["bpm_win_s_parameter"])
+        self.ui.hop_s_parameter.setText(self.config.setting["bpm_hop_s_parameter"])
+        self.ui.samplerate_parameter.setText(self.config.setting["bpm_samplerate_parameter"])
+
+    def save(self):
+        self.config.setting["bpm_win_s_parameter"] = unicode(self.ui.win_s_parameter.text())
+        self.config.setting["bpm_hop_s_parameter"] = unicode(self.ui.hop_s_parameter.text())
+        self.config.setting["bpm_samplerate_parameter"] = unicode(self.ui.samplerate_parameter.text())
+
```
zas: Add a whiteline here (PEP8 conformance)

plugins/bpm/__init__.py
```diff
@@ -0,0 +1,136 @@
+# -*- coding: utf-8 -*-
+
+# Changelog:
+#   [2015-09-15] Initial version
+# Dependancies: 
+#   aubio, numpy
+#
+# TODO:
+#   See how dependancies are isntalled automatically or added to Picard if not part of it already
```
zas: s/isntalled/installed/

plugins/bpm/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+        path: path to the file
+    """
+
+    samplerate = int(float(BMPOptionsPage.config.setting["bpm_samplerate_parameter"]))
+    win_s = int(float(BMPOptionsPage.config.setting["bpm_win_s_parameter"]))
+    hop_s = int(float(BMPOptionsPage.config.setting["bpm_hop_s_parameter"]))
+    
+    #    self.tagger.log.debug('%s' % (win_s))
+    
+    s = source(path, samplerate, hop_s)
```
zas: i would rename `s` to something more explicit

plugins/bpm/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+    samplerate = s.samplerate
+    o = tempo("specdiff", win_s, hop_s, samplerate)
```
zas: I would rename `o` to something more explicit.

plugins/bpm/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+            #    break
+        total_frames += read
+        if read < hop_s:
+            break
+
+    # Convert to periods and to bpm 
+    bpms = 60./diff(beats)
+    b = median(bpms)
+    return b
+
```
zas: Insert a whiteline here (PEP 8).

plugins/bpm/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+            #if o.get_confidence() > .2 and len(beats) > 2.:
+            #    break
+        total_frames += read
+        if read < hop_s:
+            break
+
+    # Convert to periods and to bpm 
+    bpms = 60./diff(beats)
+    b = median(bpms)
+    return b
```
zas: `b` is of no use here, just `return median(bpms)` (but if you want to keep it for debugging, in this case, use a longer variable name).

zas:
Overall, looks good, i'll give it a try soon.

zas:
Are you aware that AcousticBrainz provide such info ?
See ie. http://acousticbrainz.org/976362d6-2a82-4f26-920b-b525b1a021f0 (BPM)

madmouser1:
Many thanks for all the comments will change accordingly, I had a look at acousticbrainz but decided to go ahead with the Picard plugin as I am not aware of any integration between Picard and acousticbrainz (iow. using acousticbrainz info in Picard)

zas:
> Many thanks for all the comments will change accordingly, I had a look at acousticbrainz but decided to go ahead with the Picard plugin as I am not aware of any integration between Picard and acousticbrainz (iow. using acousticbrainz info in Picard)

Well, there is a plugin in development for AcoustiBrainz at https://github.com/musicbrainz/picard-plugins/pull/9 but it doesn't retrieve the BPM info afaik

madmouser1:
PEP8 fixes done and complete options page changed to a slider (Default to super fast) so that a user can make a easy decision and not be bothered with the technical parameters.
I do display the technical parameters but don't know how to automatically update it when the slider is moved.

zas:
@madmouser1 : i tested it, i provided a PR for things i found (https://github.com/madmouser1/picard-plugins/pull/1) but i'm wondering where is stored the BPM value, it doesn't appear in the metadata box, is there a standard bpm tag ?

phw:
> it doesn't appear in the metadata box, is there a standard bpm tag ?

Strange, according to the source it uses Picard's own "bpm" tag which gets mapped to all formats supported by Picard. Are you sure it did not show up? I just tried a simple tagger script `$set(bpm,42)` and saw the BPM info in the metadata box for all loaded recordings. Maybe the issue is that the plugin attaches the tag to the files and not the recordings?

plugins/bpm/__init__.py
```diff
@@ -0,0 +1,162 @@
+# -*- coding: utf-8 -*-
+
+# Changelog:
+#   [2015-09-15] Initial version
+# Dependancies:
+#   aubio, numpy
```
phw: One thing to note is that these imports turn this plugin pretty much into a Linux only plugin (maybe OSX, not sure). That's because the Windows version ships only a limited set of required Python modules which can not be easily extended by the user and there is no mechanism for plugins to load their own dependencies (other than including them directly).

The videotools plugin is also affected by this. Just saying as there is not much one can currently do about this.

madmouser1: noted many thanks phw I was about to ask how one would go about doing that.

zas: Should we add a meta to the header about platform compatibility ?
Something like `PLUGIN_INCOMPATIBLE_PLATFORMS`  = [_list of plaforms_]
where _list of plaforms_ consists of a list of values from https://docs.python.org/2/library/sys.html#sys.platform

I prefer to list platforms known as incompatible, so plugins not having such field will be considered as compatible and eventually emit an error.
We can do the check at load time, excluding any plugin marked as incompatible with the platform Picard is running on.

madmouser1:
what would b the best / standard way to add either a licence and/or contributors to the plugin? Or do I just list under authors ? 
@zas can I add you in there?

zas:
> what would b the best / standard way to add either a licence and/or contributors to the plugin? Or do I just list under authors ?
> @zas can I add you in there?

There are two fields you can use, see following exemple:
```python
PLUGIN_LICENSE = "GPL-2.0"
PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
```

PR #39, opened 2015-10-11
Updated Last.fm API used to v2.0, added option for Last.fm API key.

ghost:
Given that the v1.0 API used previously appears to be deprecated (or just not working), this update shifts the API to v2.0. This will mean, however, that users will need to provide an API key for this to work.

### DIFF
--- a/plugins/lastfm/ui_options_lastfm.py
+++ b/plugins/lastfm/ui_options_lastfm.py
@@ -1,14 +1,26 @@
 # -*- coding: utf-8 -*-
 
-# Automatically generated - don't edit.
-# Use `python setup.py build_ui` to update it.
+# Form implementation generated from reading ui file 'options_lastfm.ui'
+#
+# Created by: PyQt4 UI code generator 4.11.4
+#
+# WARNING! All changes made in this file will be lost!
 
 from PyQt4 import QtCore, QtGui
 
 try:
     _fromUtf8 = QtCore.QString.fromUtf8
 except AttributeError:
-    _fromUtf8 = lambda s: s
+    def _fromUtf8(s):
+        return s
+
+try:
+    _encoding = QtGui.QApplication.UnicodeUTF8
+    def _translate(context, text, disambig):
+        return QtGui.QApplication.translate(context, text, disambig, _encoding)
+except AttributeError:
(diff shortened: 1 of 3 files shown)


AdrianLC:
Really missing the Lastfm tags. I hope this gets released as soon as possible.
Thank you for the effort OP!

plugins/lastfm/ui_options_lastfm.py
```diff
@@ -86,24 +98,42 @@ def setupUi(self, LastfmOptionsPage):
...
+        self.use_track_tags.setText(_translate("LastfmOptionsPage", "Use track tags", None))
+        self.use_artist_tags.setText(_translate("LastfmOptionsPage", "Use artist tags", None))
+        self.rename_files_2.setTitle(_translate("LastfmOptionsPage", "Tags", None))
+        self.ignore_tags_2.setText(_translate("LastfmOptionsPage", "Ignore tags:", None))
+        self.ignore_tags_4.setText(_translate("LastfmOptionsPage", "Join multiple tags with:", None))
+        self.join_tags.setItemText(1, _translate("LastfmOptionsPage", " / ", None))
+        self.join_tags.setItemText(2, _translate("LastfmOptionsPage", ", ", None))
+        self.label_4.setText(_translate("LastfmOptionsPage", "Minimal tag usage:", None))
+        self.min_tag_usage.setSuffix(_translate("LastfmOptionsPage", " %", None))
+        self.label_5.setText(_translate("LastfmOptionsPage", "Last.fm API key:", None))
```
zas: I wonder about i18n here, main picard project is using a trick to get the string in plugins to be parsed by gettext stuff for translation (see https://github.com/musicbrainz/picard/blob/master/setup.py#L315-L351). But we don't have any i18n support in this repo.
Historically, plugins were in contrib/ directory in picard main repo, so translatable strings were in the same domain as main picard code. Since now we moved third-party plugins i wonder about it. We could set this repo as a sub-module of main picard repo (replacing contrib/plugins). Really not sure about what to do.
Any idea ?

samj1912:
Can you please rebase this against 2.0 and 1.0? for the appropriate versisons for API 2.0 and 1.0
API 2.0 is strictly py3/qt5

PR #40, opened 2015-10-14
Avoid potential API_KEY name clash

Sophist-UK:
in global namespace.


phw:
Not sure with what this could clash inside this file but I'm fine with this change ;)

Sophist-UK:
Hmmm - I thought it was a global variable, but upon reflection I am not sure that Python has such things - as you need to import a reference to make it available.

So merge or cancel as you wish.

phw:
Yeah, there is not really something global. It can still give conflicts if one uses the "from module import *" syntax, but that's why it is considered bad practice to use this.

Anyway, I would merge it just for the clearer naming. I just don't have merge permission for this repository :)

PR #41, opened 2015-10-15
Add Acousticbrainz Tonal/Rhythm plugin

Sophist-UK:
This plugin gets the musical key and the beats per minute from AcousticBrainz.

There is no rate limiting for AcousticBrainz, so this allows 20 calls per second on the basis that:

a. Too many more might flood broadband routers; and
b. This is the right order of magnitude for track by track AcousticBrainz calls to keep pace with MB album lookups which are rate limited to 1 per second.

This plugin requires Picard 1.4.0.

### DIFF
--- a/plugins/acousticbrainz/acousticbrainz.py
+++ b/plugins/acousticbrainz/acousticbrainz.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+
+PLUGIN_NAME = _(u'AcousticBrainz')
+PLUGIN_AUTHOR = u'Sophist'
+PLUGIN_DESCRIPTION = u'''Add's the following tags:
+<ul>
+<li>Key (in ID3v2.3 format)</li>
+<li>Beats Per Minute (BPM)</li>
+</ul>
+from the AcousticBrainz database providing:
+<ol type="a">
+<li>the musicbrainz_recordingid is in the track's metadata; and</li>
+<li>the key is in the AcousticBrainz database.</li>
+</ol>'''
+PLUGIN_VERSION = '0.1'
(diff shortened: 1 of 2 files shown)


mineo:
Is this in any way related to #9 ?

Sophist-UK:
No - I wasn't aware of #9. This is a completely separate plugin.

As far as I can tell, #9 is looking for Genres in the AcousticBrainz High-Level call.

This one is looking for musical Key and BPM in the Low-Level call.

Sophist-UK:
I have made some name changes.

Sophist-UK:
No idea why Travis checks are not working.

mineo:
My guess is that Travis doesn't work because the plugins name is wrapped in a gettext call, which isn't available in the test script.

Sophist-UK:
Ah - easy to change.

zas:
> My guess is that Travis doesn't work because the plugins name is wrapped in a gettext call, which isn't available in the test script.

In fact, this is due to the fact i switched to `ast` to parse plugins metadata, and i didn't implemented support for _() and N_() in plugin metadata: https://github.com/musicbrainz/picard-plugins/commit/c1041f8e9a41252cd550d376fcf2ac1a0c6ec8af

It was to solve other issues we had with complex regex parsing (ie. multiline arrays), and is much simpler.

I think it should be re-introduced at some point though.
It wasn't much an issue since most plugins weren't using gettext for metadata.
I removed 2 occurences in https://github.com/musicbrainz/picard-plugins/commit/b4acc530c6020ae06fd4f23f02313639ce8325a8

So, for now, don't use gettext (or any other function) in plugin metadata fields, they only support simple types (strings, arrays, etc).

PR #42, opened 2015-10-24
Don't sort artists tags...

Sophist-UK:
... because position is important for these tags.

### DIFF
--- a/plugins/sort_multivalue_tags/sort_multivalue_tags.py
+++ b/plugins/sort_multivalue_tags/sort_multivalue_tags.py
@@ -15,7 +15,14 @@
 
 PLUGIN_NAME = u"Sort Multi-Value Tags"
 PLUGIN_AUTHOR = u"Sophist"
-PLUGIN_DESCRIPTION = u'Sort Multi-Value Tags e.g. Release Type, Lyrics alphabetically.'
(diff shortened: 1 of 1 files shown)


mineo:
This looks similar to https://github.com/musicbrainz/picard/pull/272 but with less exceptions?

Sophist-UK:
Actually [musicbrainz/picard#272](https://github.com/musicbrainz/picard/pull/272) **is** a better version - but of course, that was submitted before we had a separate plugins infrastructure, and as I understand it Contributed plugins are due to be removed from main Picard repo shortly.

I will port over the changes to this PR.

plugins/sort_multivalue_tags/sort_multivalue_tags.py
```diff
@@ -16,19 +16,29 @@
 PLUGIN_NAME = u"Sort Multi-Value Tags"
 PLUGIN_AUTHOR = u"Sophist"
 PLUGIN_DESCRIPTION = u'Sort Multi-Value Tags e.g. Release Type, Lyrics alphabetically.'
```
mineo: This isn't a good description because release type is in the exception list and lyrics are not set automatically.

Sophist-UK: I agree.

PR #44, opened 2015-11-19
implement a basic cyrillic de-mangling plugin

Aeon:
This plugin helps you quickly convert mis-encoded cyrillic Windows-1251 tags 
to proper UTF-8 encoded strings. If your track/album names look something like
"Àëèñà â ñò›àíå ÷óäåñ", run this plugin from the context menu
before running the "Lookup" or "Scan" tools

### DIFF
--- a/plugins/decode_cyrillic/decode_cyrillic.py
+++ b/plugins/decode_cyrillic/decode_cyrillic.py
@@ -1,0 +1,103 @@
+# -*- coding: utf-8 -*-
+
+# This is the Decode Cyrillic plugin for MusicBrainz Picard.
+# Copyright (C) 2015 aeontech
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import print_function
+PLUGIN_NAME = u"Decode Cyrillic"
+PLUGIN_AUTHOR = u"aeontech"
+PLUGIN_DESCRIPTION = u'''
+This plugin helps you quickly convert mis-encoded cyrillic Windows-1251 tags 
+to proper UTF-8 encoded strings. If your track/album names look something like
+"Àëèñà â ñò›àíå ÷óäåñ", run this plugin from the context menu
+before running the "Lookup" or "Scan" tools
+'''
+PLUGIN_VERSION = "1.0"
+PLUGIN_API_VERSIONS = ["1.0"]
+PLUGIN_LICENSE = "MIT"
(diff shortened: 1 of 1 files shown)


Sophist-UK:
IMO this is something that needs to be implemented in Picard itself.

I have a PR already submitted some time ago [musicbrainz/pricard#221](https://github.com/musicbrainz/picard/pull/221)  to handle character sets better, which I will be reworking after my current PR, so I will see whether this can be included or whether it needs a separate PR.

Can you please supply details of the use case where this occurs and / or some MBIDs which load incorrectly and / or some test files which don;t load correctly.

Thx.

Sophist-UK:
Upon reading the code comments, it appears that this is designed to correct files which have been mis-tagged in the wrong encoding by another tool. Is that correct? If so, then please disregard my previous comment as this would indeed fit into a plugin.

Aeon:
@Sophist-UK correct, this is to make handling mis-encoded tags easier. This is a very frequent occurrence when moving old files tagged on Russian Windows machines into the modern age.

plugins/decode_cyrillic/decode_cyrillic.py
```diff
@@ -0,0 +1,103 @@
...
+    'album', 
+    'artistsort'
+]
+# _from_encoding = "latin1"
+# _to_encoding   = "cp1251"
+
+
+# TODO:
+# - extend to support multiple codepage decoding, not just cp1251->latin1
+#   instead, try the common variations, and show a dialog to the user,
```
Sophist-UK: I guess if this is run from a context menu, then a future dialog would be OK. But it would be better to avoid this if you can.

plugins/decode_cyrillic/decode_cyrillic.py
```diff
@@ -0,0 +1,103 @@
...
+                    file.orig_metadata[tag] = unmangled_tag
+                    file.metadata[tag] = unmangled_tag
+
+                    file.orig_metadata.changed = True
+                    file.metadata.changed = True
+                    file.update(signal=True)
+
+            cluster.update()
+
+register_cluster_action(DecodeCyrillic())
```
Sophist-UK: Is cluster_action the only place you want to run this plugin?

How about register_file_action, register_clusterlist_action etc.?

Also, is this something that can be run automatically when you load a file i.e. can it tell when a file needs converting and whether the conversion works or not (or is this something that only a human can decide)? If it can be made completely automated, that would be fantastic.

plugins/decode_cyrillic/decode_cyrillic.py
Aeon: It's more of a nice-to-have note for the future, the cp1251-latin1 misencoding is the most common in my experience.

plugins/decode_cyrillic/decode_cyrillic.py
Aeon: > Is cluster_action the only place you want to run this plugin?

generally I want to decode the entire album. This seems to work fine now, even when I select multiple clusters, the code iterates over them and demangles each one correctly.

> How about register_file_action, register_clusterlist_action etc.?

I have no objection, just not sure what changes are required to make them work, or what the benefit will be. This is my first picard plugin.

> Also, is this something that can be run automatically when you load a file i.e. can it tell when a file needs converting and whether the conversion works or not (or is this something that only a human can decide)? 

I don't think so, since the user is just as likely to be tagging latin or chinese or some other character set. I suppose some black magic "likely language" detection with frequency analysis could be implemented, but that would be both too unreliable, and too magical for my taste – especially considering the fairly narrow use case for this plugin, I fear trying to make it more automagic would cause it to do more damage than good by mangling perfectly good tags when guessing wrong.

With that in mind, I think it's reasonable to make it a user-invoked action.

Sophist-UK: @Aeon: 

1. Suggest you experiment with the other actions to see if needed.

2. I agree that automatic analysis needs to be accurate and not translate false positives. So best to keep it manual.

Aeon: @Sophist-UK 

I haven't had much time to spend on this, can we maybe merge it as is? I haven't seen any documentation for `register_file_action` and `register_clusterlist_action` so I'm not sure what the benefit of those would be?

plugins/decode_cyrillic/decode_cyrillic.py
```diff
@@ -0,0 +1,103 @@
...
+
+                    unmangled_tag = self.unmangle(tag, file.metadata[tag])
+
+                    file.orig_metadata[tag] = unmangled_tag
+                    file.metadata[tag] = unmangled_tag
+
+                    file.orig_metadata.changed = True
+                    file.metadata.changed = True
+                    file.update(signal=True)
+
```
Aeon: by the way, I came up with lines 94-99 by looking at other plugins and trial and error, it works but I'm not sure what exactly `file.update` etc actually _do_, so if there's a better way to do it, please comment.

mwiencek:
Thanks. Feel free to submit another PR if you have improvements to make.

PR #45, opened 2016-01-05
FanartTv plugin runs even when it's disabled

rahul-raturi:
If one has installed and activated fanarttv plugin, but disabled it in cover art settings, picard still fetches the coverart from fanarttv. In my situation,
 
fanart

I suppose this can be considered a bug.


phw:
Thanks for this, the change itself looks good. Can you also bump the plugin version?

That's one point less on my todo list :)

rahul-raturi:
@phw Sure. Done.

phw:
From me a +1 for merging. We just need someone with write access to actually do it ;)

mineo:
There you go.

PR #48, opened 2016-02-28
Add non-ASCII equivalents plugin

andersonvom:
This plugin tries to do a better job than Picard's native "Replace
non-ASCII characters" by replacing such characters with the most similar
ASCII one. This allows old devices to be able to display artists and
song titles somewhat correctly instead of using default blank or weird
characters.

### DIFF
--- a/plugins/non_ascii_equivalents/non_ascii_equivalents.py
+++ b/plugins/non_ascii_equivalents/non_ascii_equivalents.py
@@ -1,0 +1,149 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (C) 2016 Anderson Mesquita <andersonvom@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


plugins/non_ascii_equivalents/non_ascii_equivalents.py
```diff
@@ -0,0 +1,149 @@
...
+}
+
+FILTER_TAGS = [
+    "album",
+    "artist",
+    "title",
+]
+
+
+def sanitize(char):
```
mineo: This is the same as `CHAR_TABLE.get(char, char)`.

mineo:
There seems to be some overlap between the replacements in this plugin and the currently existing replacement tables in https://github.com/metabrainz/picard/blob/master/picard/util/textencoding.py (like Æ, ß, possibly more). Do you know which ones are only in this plugin or where the replacement's here are different?

Sophist-UK:
See also #221 which I wrote but has never been merged which attempts to do a comprehensive conversion.

PR #49, opened 2016-03-08
Added picard-tangoinfo plugin

ix5:
Hey Picard devs,

we tango folks really like MusicBrainz. But sometimes, it does not have enough information on the really old recordings. And in tango music, there is usually little doubt about the genre. And most importantly, we do not really care about the release date of a compilation of old songs, we *have* to know the original release date(30s, 40s, 50s are all very different eras to us).
So that is why tagging tango music involves "cleaning up" with Picard, and then tagging genre, release date and the singers manually by cross-referencing tango.info or other sites.

I have written this plugin to ease that process in Picard. It ain't pretty, as it scrapes the website and only works of MusicBrainz provides a barcode, but I believe it will help many tango DJs organize their collections better and save a lot of time.

Let me know if there is a problem with approving plugins that utilize scraping(tango.info has shut down their opendata initiative, sadly) or if the code could be improved. There seems to be very little information on the plugin API; this is my best shot.
I can confirm the plugin works on Ubuntu 15.10 with Picard version 1.3.2 from the repositories.

### DIFF
--- a/plugins/picard-tangoinfo/README.md
+++ b/plugins/picard-tangoinfo/README.md
@@ -1,0 +1,34 @@
+# tango.info plugin for MusicBrainz Picard
+The [old one](http://simbatango.com/2010/07/23/updated-musicbrainz-picard-plugins-for-tango-info/)
+is not working anymore, the opendata initiative has stalled,
+so we are back to good old web scraping.
+
+# Usage
+You will need MusicBrainz Picard and these python libraries:
+
+- urllib2
+- bs4(BeautifulSoup)
+- re(should be preinstalled)
+
+Put `tangoinfo.py` into your plugin directory. On \*nix, this
+would be `~/.config/MusicBrainz/Picard/plugins`, consult the
+Picard documentation for other operating systems.  
+Then open Picard, navigate to `Options -> Options -> Plugins` and enable
+the one called `ix5 - tango.info`, then restart and use Picard
+like you usually would.
+
+If(and **only if**) there is a valid `barcode` tag for your files,
+the plugin will try to look up the corresponding album on *tango.info*
+and fill in the `genre`(capitalized) and `year` tags for you.
+
+# What it does
+It looks up the barcode, constructs the url for the corresponding 
+*tango.info* album('product'/'TINP') page, parses the html,
+looks for the `tracks` table and extracts `genre` and `Perf date`.
+
+# Future
+Provided [Tobias Conradi](https://tango.info/tangoinfo/eng/contact)
+does not change the page structure in a manner that breaks this plugin,
+you might be able to map instrumentalists and vocalists to the id3 tags
+of your liking. Let us hope he does not ban urllib2 clients,
+or we will have to fumble with user agents.
--- a/plugins/picard-tangoinfo/tangoinfo.py
+++ b/plugins/picard-tangoinfo/tangoinfo.py
@@ -1,0 +1,71 @@
+"""
+Get genre, performed date and (soon) vocalists and instrumentalists from tango.info
+Uses web scraping since opendata is disabled for now
+"""
+
+PLUGIN_NAME = 'ix5 tango.info'
+PLUGIN_AUTHOR = 'ix5'
+PLUGIN_DESCRIPTION = 'change <em>genre</em> and <em>year</em> fields'
+PLUGIN_VERSION = '0.1'
+PLUGIN_API_VERSIONS = ["0.9.0", "0.12", "1.0.0"]
+
+from picard import log
+from picard.metadata import register_track_metadata_processor
+import re,os
+
+from bs4 import BeautifulSoup
+import urllib2
+
+scraped_data = {}
+
+def scrape_page(barcode):
+    url = ("https://tango.info/0" + str(barcode))
+    page = urllib2.urlopen(url).read()
+    soup = BeautifulSoup(page, "lxml")
+    table = soup.findAll("table")[3]
+    x = (len(table.findAll('tr')) -1)
+
+
+    for row in table.findAll('tr')[1:]:
+        col = row.findAll('td')
+        genre = col[3].getText().capitalize()
+        perf_date = col[6].getText()
+        # checks
+        info = col[-1].find('a').get('href').split('/')[1] # get TINT
+        #track_tint = "0" + str(int(metadata["barcode"])) + "-" + str(mydiscnumber) + "-" + metadata["tracknumber"]
+        scraped_data[info] = {"genre": genre, "year": perf_date}
+        #: Data expected to be in this format:
+        # ["Side", "Tracknr", "Title", "Genre", "Instrumentalists", "Vocalists", "Perf Date", "Duration", "Info"]
+
+_discnumber_re = re.compile(r"\s+\(disc (\d+)(?::\s+([^)]+))?\)")
+
+def get_discnumber(metadata):
+    retval = 1
+    if (is_integer(metadata["discnumber"])):
+        if metadata["discnumber"]>0:
+            retval = metadata["discnumber"]
+        else:	
+            matches = _discnumber_re.search(metadata["album"])
+            if matches:
+                retval = matches.group(1)
+    return retval
+
+def is_integer(testinteger):
+	try:
+		int(testinteger)
+		return True
+	except:
+		return False	
+
+
+def set_data(tagger, metadata, release, track):
+    mydiscnumber = get_discnumber(metadata)
+    tangoinfo_tint = str("00" + str(int(metadata["barcode"])) + "-" + str(mydiscnumber) + "-" + metadata["tracknumber"])
+    #: prefixing 00 is weird, but should work
+    scrape_page(metadata["barcode"])
+    metadata['genre'] = scraped_data[tangoinfo_tint]['genre']
+    log.debug("Set genre %s" % scraped_data[tangoinfo_tint]['genre'])
+    metadata['year'] = scraped_data[tangoinfo_tint]['year']
+    log.debug("Set year: %s" % scraped_data[tangoinfo_tint]['year'])
+
+register_track_metadata_processor(set_data)


Sophist-UK:
My cousin is a Tango fan, so I know how many people will find this useful. I will add a few constructive code comments separately.

plugins/picard-tangoinfo/README.md
```diff
@@ -0,0 +1,34 @@
+# tango.info plugin for MusicBrainz Picard
+The [old one](http://simbatango.com/2010/07/23/updated-musicbrainz-picard-plugins-for-tango-info/)
+is not working anymore, the opendata initiative has stalled,
+so we are back to good old web scraping.
+
+# Usage
+You will need MusicBrainz Picard and these python libraries:
+
+- urllib2
```
Sophist-UK: If you want this plugin to work on Windows for people who run the executable then you need to try to avoid using libs that Picard doesn't already include.

Picard already makes web calls, so try to use the routines already in Picard webservice.py to send web calls. And whilst bs4 may make screen scraping easy, it might be worth using native facilities like `re` to screen scrape - or ask tango.info to provide a REST XML API for TINP and TINT searches.

plugins/picard-tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,71 @@
...
+    mydiscnumber = get_discnumber(metadata)
+    tangoinfo_tint = str("00" + str(int(metadata["barcode"])) + "-" + str(mydiscnumber) + "-" + metadata["tracknumber"])
+    #: prefixing 00 is weird, but should work
+    scrape_page(metadata["barcode"])
+    metadata['genre'] = scraped_data[tangoinfo_tint]['genre']
+    log.debug("Set genre %s" % scraped_data[tangoinfo_tint]['genre'])
+    metadata['year'] = scraped_data[tangoinfo_tint]['year']
+    log.debug("Set year: %s" % scraped_data[tangoinfo_tint]['year'])
+
+register_track_metadata_processor(set_data)
```
Sophist-UK: This is a Track Metadata processor rather than an Album metadata processor, so it runs individually for each track. But scrape_page makes a call based on the barcode i.e. the same call for every track on the album.

So for performance reasons you should screen scrape all the album info and cache it. See my plugin for Artist Official Web Page for examples of how to cache and (because it will queue the calls for every track before it gets the response from the first) avoid making the same call 20x in parallel.

plugins/picard-tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,71 @@
...
+PLUGIN_API_VERSIONS = ["0.9.0", "0.12", "1.0.0"]
+
+from picard import log
+from picard.metadata import register_track_metadata_processor
+import re,os
+
+from bs4 import BeautifulSoup
+import urllib2
+
+scraped_data = {}
```
Sophist-UK: This is a global variable - so you need to be careful that the index is unique.

plugins/picard-tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,71 @@
...
+		return False	
+
+
+def set_data(tagger, metadata, release, track):
+    mydiscnumber = get_discnumber(metadata)
+    tangoinfo_tint = str("00" + str(int(metadata["barcode"])) + "-" + str(mydiscnumber) + "-" + metadata["tracknumber"])
+    #: prefixing 00 is weird, but should work
+    scrape_page(metadata["barcode"])
+    metadata['genre'] = scraped_data[tangoinfo_tint]['genre']
+    log.debug("Set genre %s" % scraped_data[tangoinfo_tint]['genre'])
```
Sophist-UK: You may want to prefix debug messages with `plugin-name: ` so as to identify where the messages have come from.

ix5:
Alright, thank you for your input. Let me work on this for a while and update the PR.
By global variable you mean "scraped_data"? The keys are TINTs(0+barcode+discnr+tracknr), which are always unique.

Sophist-UK:
Yes - I thought that was the case and that they were unique - but just wanted to make sure that this wasn't expected to store only the values for one album and that it would not be overwritten by a second album.

plugins/picard-tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,71 @@
...
+from picard.metadata import register_track_metadata_processor
+import re,os
+
+from bs4 import BeautifulSoup
+import urllib2
+
+scraped_data = {}
+
+def scrape_page(barcode):
+    url = ("https://tango.info/0" + str(barcode))
```
Sophist-UK: Not all MB releases have barcodes. You do not check anywhere that the release has a barcode before issuing a web call where the URL includes the barcode.

ix5: It used to fail silently before. I have added a check now.
`if (len(str(metadata["barcode"]))>0):`

zas: [**@ix5**](https://github.com/ix5) commented on [9 mars 2016 01:04 UTC+1](https://github.com/metabrainz/picard-plugins/pull/49#discussion_r55451700 "2016-03-09T00:04:03Z - Replied by Github Comment Enhancer"):
> It used to fail silently before. I have added a check now.
> `if (len(str(metadata["barcode"]))>0):`

```python
if metadata["barcode"]:
```
should be enough (if "barcode" key always exists).
https://www.python.org/dev/peps/pep-0008/#programming-recommendations

plugins/picard-tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,71 @@
...
+
+PLUGIN_NAME = 'ix5 tango.info'
+PLUGIN_AUTHOR = 'ix5'
+PLUGIN_DESCRIPTION = 'change <em>genre</em> and <em>year</em> fields'
+PLUGIN_VERSION = '0.1'
+PLUGIN_API_VERSIONS = ["0.9.0", "0.12", "1.0.0"]
+
+from picard import log
+from picard.metadata import register_track_metadata_processor
+import re,os
```
zas: Please split  those:
```python
import os
import re
```

plugins/picard-tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,71 @@
...
+        #track_tint = "0" + str(int(metadata["barcode"])) + "-" + str(mydiscnumber) + "-" + metadata["tracknumber"]
+        scraped_data[info] = {"genre": genre, "year": perf_date}
+        #: Data expected to be in this format:
+        # ["Side", "Tracknr", "Title", "Genre", "Instrumentalists", "Vocalists", "Perf Date", "Duration", "Info"]
+
+_discnumber_re = re.compile(r"\s+\(disc (\d+)(?::\s+([^)]+))?\)")
+
+def get_discnumber(metadata):
+    retval = 1
+    if (is_integer(metadata["discnumber"])):
```
zas: Useless outer parenthesis

plugins/picard-tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,71 @@
...
+        info = col[-1].find('a').get('href').split('/')[1] # get TINT
+        #track_tint = "0" + str(int(metadata["barcode"])) + "-" + str(mydiscnumber) + "-" + metadata["tracknumber"]
+        scraped_data[info] = {"genre": genre, "year": perf_date}
+        #: Data expected to be in this format:
+        # ["Side", "Tracknr", "Title", "Genre", "Instrumentalists", "Vocalists", "Perf Date", "Duration", "Info"]
+
+_discnumber_re = re.compile(r"\s+\(disc (\d+)(?::\s+([^)]+))?\)")
+
+def get_discnumber(metadata):
+    retval = 1
```
zas: If `retval` is a discnumber, please just use `discnumber`.

plugins/picard-tangoinfo/tangoinfo.py
zas: Isn't `barcode` already a string ?
```
url = "https://tango.info/0" + barcode
```
should be enough.

plugins/picard-tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,71 @@
...
+	try:
+		int(testinteger)
+		return True
+	except:
+		return False	
+
+
+def set_data(tagger, metadata, release, track):
+    mydiscnumber = get_discnumber(metadata)
+    tangoinfo_tint = str("00" + str(int(metadata["barcode"])) + "-" + str(mydiscnumber) + "-" + metadata["tracknumber"])
```
zas: What about using formats ?
```
tangoinfo_tint = "00%d-%d-%d" % (metadata["barcode"], mydiscnumber, metadata["tracknumber"])
```

ix5:
@zas the whole thing is a mess, I blindly copied a lot from another plugin without thinking. I am working on implementing the queue system @Sophist-UK suggested and rewriting without bs4 and urllib2 at the moment, will prettify the code in the process.

I am encountering a weird bug that on certain albums, if I paste in the barcode manually using Picard, restart the program and re-add the files, it will show the correct "barcode" information in the user interface, but the plugin will think there is now barcode tag field present.
I checked using `log.debug(metadata)`, and the resulting dict does not contain the barcode field for any file.
So it will take some time to iron out that.

ix5:
Alright, some progress has been made. 
I think the issue was that the `barcode` field is not uniformly in one case, i.e. sometimes `BARCODE`, sometimes `barcode`, at least for some of my albums.

Picard will display either as `Barcode` in the UI - my files are mp3s, so I think [id3.py](https://github.com/metabrainz/picard/blob/9847a9c96cd421cd7928d4af03be943580cebc0e/picard/formats/id3.py#L174) is doing the conversion, but the plugin will only recognize `barcode`. So with
```
if metadata["barcode] or metadata["BARCODE"]:
    ...
```
it should work.

Regarding the queue system from @Sophist-UK : Does `tagger.xmlws.get()` also download normal html files or can it only retrieve and parse xml from MusicBrainz? I do not want to use urllib since you said the plugin should run on Windows, too.

I am using this: `<h2><a href="\/tracks">Tracks<\/a><\/h2>(?!<\/table>)(.+?<\/table>)` to get the table from tango.info (an album page looks like [this](https://tango.info/00094636268620)) and then this regex: `<tr>((?!<\/tr>).+?)</tr>` for the individual tracks, with some further parsing to be done

I have another question for @Sophist-UK : When you extract the `albumartistids` from `track_metadata` in your [albumartist_website plugin](https://github.com/Sophist-UK/sophist-picard-plugins/blob/master/albumartist_website.py#L65), do you expect there to be multiple ids per track, because you go over them in a for loop?
To me it seems that `track_metadata['musicbrainz_albumartistid']` can only return one value? Or is this done for all tracks in an album? I am confused.
This is what I for checking the barcode:
```
if track_metadata["barcode"]:
            barcode = track_metadata["barcode"]
        elif track_metadata["BARCODE"]:
            barcode = track_metadata["BARCODE"]
        else:
            barcode = []
        #barcodes = dict.get(track_metadata, 'barcode', [])
        #for barcode in barcodes:
        if barcode in self.website_cache:
            if self.website_cache[barcode]:
	        # Checks, as not to overwrite with empty data
	        if self.website_cache[barcode][title]:
                	track_metadata['title'] = self.website_cache[barcode][title]
	        if self.website_cache[barcode][genre]:
                	track_metadata['genre'] = self.website_cache[barcode][genre]
	        if self.website_cache[barcode][year]:
                	track_metadata['year'] = self.website_cache[barcode][year]
	        if self.website_cache[barcode][vocal]:
                	track_metadata['vocal'] = self.website_cache[barcode]vocal]
        else:
            self.website_add_track(album, album._new_tracks[-1], barcode)
```

The more I trouble myself with this, the more questions pop up!

[This](https://github.com/ix5/picard-tangoinfo/blob/a247f79cd94ce82c555853ffd8d6fec6541b83d8/tangoinfo.py) is what I think the plugin should look like, some pseudocode still in there, but I hope you can tell me whether the thinking behind it is sound.

ix5:
Okay, I have worked on it and it should be stable for now. Latest version is [here](https://github.com/ix5/picard-tangoinfo/blob/master/tangoinfo.py)
The only thing is that it now has this weird bug with a duplicated entry for the first item.
duplication

@Sophist-UK , you could help me sort this out, since I borrowed heavily from your albumartist plugin(also, there is a slight mistake in [your new version](https://github.com/Sophist-UK/sophist-picard-plugins/commit/4a29a9ad5dbc2f62436111426566bb0bdc7c7191#commitcomment-16982685)). It looks like I might have missed a call to remove something from the queue and that is why it does this? I cannot find my mistake somehow.
Your code is very elegant, but I have trouble understanding it.

Regarding the other issues with this plugin you and @zas pointed out:
- I have dropped `bs4` and `urllib2` and done everything with `re`, `urllib`(the standard module) and string formatting. I hope it works for other users now.
- It is using a queue system, only downloading the web page once, very fast and does not lock up the interface anymore
- Log messages have the `PLUGIN_NAME` prefixed
- Using formats more, it is so much easier to read with them, thank you @zas 
- It should work with both python2 and python3
- It checks whether a barcode exists and acts accordingly
- It can now get `date`, `vocal`(Singers) and `genre` information
- It has a license and license information embedded in the plugin, see here #5 

That's it, I hope. If you have any more suggestions, I'd be happy to hear about them.
Once this bug is worked out, I will update the PR and picard will hopefully be more attractive to tango enthusiasts

Sophist-UK:
I have quite a few comments:

a. You should try to use `album.tagger.xmlws.get` (which is async) rather than `urlopen` (which is thread blocking) for performance reasons. Have you tried `album.tagger.xmlws.get` with XML=False?

(That said, the impact is probably small, and as far as I can tell, your code should still work ok.)

b. On code style, your logging messages should ideally be more explanatory - e.g. "No barcode in track" begs the question which track - or really which album. Also, e.g. a lack of a barcode is NOT an error - it is at worst only a warning. 

Otherwise, I can see no reason for a duplicated track. I assume you don't get the duplication if your plugin is not loaded.

Sophist-UK:
P.S. I can't see the reason for it in your code, but the fact that it is the first track that is duplicated is suggestive of some sort of issue with multi-threading and blocking, as it is the first track which will create the web request. The issue could be e.g. elsewhere in Picard.

ix5:
I have no idea how to debug xmlws, it just fails and after a while gives unhelpful output like
`D: 22:23:57 Last request to ('tango.info', '80') was 1000 ms ago, starting another one`
and doesn't call the `handler` I have defined. Ideally, I'd like to get a string object of the html response that I can parse further, don't really care whether it is from xmlws or urllib.

I have created a gist to demonstrate the issue: [xmltest.py](https://gist.github.com/ix5/a7b8258b7593888a93daf91fa6a5a505)

Looking at `webservice.py` from picard, `xmlws.get()` seems to do an `add_task` to the picard webservice queue and then call some QT networking stuff. I just wish it was more verbose in telling me what is going on internally.

--
The log messages are hopefully all clear now and only at `debug` level.

Sophist-UK:
The stuff you are commenting on is the rate limiter - which sends (by default) only  request per second (though you can change this per web site).

You need to understand what the `partial` function does in order to understand how the callback system works.

If you set xml=false the you should get the response as a string sent to the handler as the first parameter that is not pre-set by the `partial` function. See `_handle_reply` in webservice.py.

ix5:
If I understand `partial` correctly, it runs another function, but with a default value set, no?

ix5:
Regarding the rate limiter, it shouldn't even have to send more than one request to tango.info...

Sophist-UK:
Partial create a temporary function which has some of the arguments pre-applied - but it doesn't run the function at that time. So the handler function is still passed as a callable function - but some of the arguments are already set.

You should send one request to tango.info **per album** - so if you load multiple albums at the same time, your plugin will send multiple requests - if you send too many simultaneous calls, you may overload your router or you may trigger rate limits or DoS protection on the server. Unless you are doing one call per track, rate limiting is not normally a major issue for Picard.

The reason that the requests number in the bottom right hand corner does not decrease is because we are not processing the response. This is linked to the handler not being called - but this number is decreased by Picard when the request completes and before the handler is called - so it looks like the request is not completing for some reason. Can you please:

a. Post your current code which uses webservice.py;
b. Set debug on and post the debug log when you load the album.

Thx.

Sophist-UK:
P.S. No idea why the gist doesn't work. Code looks ok to me.

ix5:
Using xmlws: https://gist.github.com/ix5/89da3e6ec19ce76684f8eb624cba9e41
The log: https://gist.github.com/ix5/91f2e2fcff20cd61195f10e9fe639523

I have only added one file for this test(Brazo de oro), so there really should not be more than one request.

Sorry for not being clear enough earlier, my comment about the rate limiter log entry not being helpful was more out of frustration that it didn't tell me what happened to the first request.

In theory, this code *should* work, yet somehow there's always some hidden error that I am not aware of. It *kinda* works now, but I have to remove and re-add the tracks every time for it to work. But that is not acceptable for other users.
I have since included puddletag into my workflow, using a tiny script that scrapes tango.info and puts the data into my clipboard formatted in a special way that puddletag can work with.

Maybe it wasn't meant to be with picard; given my unfamiliarity with the codebase and very limited skill with python (and lack of spare time), I doubt I will produce anything useful if I continue working on this.

@Sophist-UK, thank you so much for all the help so far, even if it might have been in vain.

Sophist-UK:
Ok - the reason that you are not getting a response is because Picard is not sending the request.

If you look further up the debug log you see it says:
```
D: 17:29:16 Last request to (u'musicbrainz.org', 80) was 1000 ms ago, starting another one
D: 17:29:16 GET http://musicbrainz.org/ws/2/release/1d1e95e9-06e8-48e3-b7b5-478b433e7d49?inc=release-groups+media+recordings+artist-credits+artists+aliases+labels+isrcs+collections+artist-rels+release-rels+url-rels+recording-rels+work-rels
D: 17:29:17 Received reply for http://musicbrainz.org/ws/2/release/1d1e95e9-06e8-48e3-b7b5-478b433e7d49?inc=release-groups+media+recordings+artist-credits+artists+aliases+labels+isrcs+collections+artist-rels+release-rels+url-rels+recording-rels+work-rels: HTTP 200 (OK) 
```
But at the end of the log it just says:
```
D: 17:29:17 Last request to ('tango.info', '80') was 1000 ms ago, starting another one
```
So, it looks to me like the request was never sent - the question is why?

My guess (and you might kick yourself if it turns out to be the reason) is that you have specified the port as a string rather than as a number.

Let me know if this fixes it.

Sophist-UK:
P.S. Please reopen this issue and keep trying - I don't think you are far from a working plugin now.

ix5:
Whoooohooo, it's working! Will update asap!

Sophist-UK:
And did you kick yourself? ;-)

ix5:
Okay, all clear now. I just opened a new PR, because I deleted my old fork repo...

https://github.com/metabrainz/picard-plugins/pull/51

ix5:
Again, thank you so much! No more duplication, it's beautiful!
I was so close to giving up, that string/int mistake was vicious!

PR #51, opened 2016-04-10
Added plugin tangoinfo

ix5:
See the [old one](https://github.com/metabrainz/picard-plugins/pull/49), can't reopen so I created a new one.
All works now.

### DIFF
--- a/plugins/lastfmplus/__init__.py
+++ b/plugins/lastfmplus/__init__.py
@@ -580,7 +580,8 @@ class LastfmOptionsPage(OptionsPage):
                 else:  # data lines
                     for column, value in zip(columns, data):
                         if value:
-lists[column].append(value)
+                            lists[column].append(value)
+
         self.ui.genre_major.setText(', '.join(lists['Major']))
         self.ui.genre_minor.setText(', '.join(lists['Minor']))
         self.ui.genre_country.setText(', '.join(lists['Country']))


plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+from picard.util import LockableObject
+from picard.metadata import register_track_metadata_processor
+
+from functools import partial
+
+# python2/python3 compatibility
+try:
+    from urllib.request import urlopen as urlopen
+except ImportError:
+    from urllib import urlopen
```
mineo: These imports are not used anywhere.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+
+        def __contains__(self, name):
+            return name in self.queue
+
+        def __iter__(self):
+            return self.queue.__iter__()
+
+        def __getitem__(self, name):
+            self.lock_for_read()
+            value = self.queue[name] if name in self.queue else None
```
mineo: It doesn't matter much in this case, but that's exactly what `value = self.queue.get(name)` would have done

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+            #         )
+            return
+
+        if track_metadata["discnumber"]:
+            discnr = str(track_metadata["discnumber"])\
+                    .split("/")[0]\
+                    .lstrip("0")
+                    # For discnumbers like 02/12"
+        else:
+            discnumber = "1"
```
mineo: `discnr` at one point and `discnumber` at another can't be correct.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+            #log.debug("%s: No barcode found for %s" % (PLUGIN_NAME, \
+            #        trackXmlNode) # No __getitem__ function? too verbose...
+            #         )
+            return
+
+        if track_metadata["discnumber"]:
+            discnr = str(track_metadata["discnumber"])\
+                    .split("/")[0]\
+                    .lstrip("0")
+                    # For discnumbers like 02/12"
```
mineo: I think the discnumber tag in Picard will always be just an integer and not contain the number of total discs.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+        else:
+            discnumber = "1"
+        tracknr = str(track_metadata["tracknumber"])\
+                .split("/")[0]\
+                .lstrip("0")
+                # For tracknumbers like 02/12"
```
mineo: I think this will never happen.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+                        track_metadata['vocal'] = self.albumpage_cache[barcode][tint]['vocal']
+            else:
+                log.debug("%s: No information on tango.info for barcode %s" % (PLUGIN_NAME, barcode))
+        else:
+            #log.debug("%s: Adding to queue: %s, new track: %s" % (PLUGIN_NAME, barcode, album._new_tracks[-1]))
+            self.website_add_track(album, album._new_tracks[-1], barcode)
+
+
+    def website_add_track(self, album, track, barcode):
+        #: returns the tango.info website, raw
```
mineo: It doesn't.

plugins/tangoinfo/tangoinfo.py
Sophist-UK: @mineo is correct - in Picard the values are already stored separately - see e.g. `totaltracks`. They are combined only when saved to some file formats and split when you load metadata from these same formats.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+        self.album_add_request(album) # increase counter, gets decreased later
```
mineo: It doesn't get descreased if the condition on the next line is False because `website_process` will never be called.

plugins/tangoinfo/tangoinfo.py
Sophist-UK: See similar comment below - see also `totaldiscs`.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+            #log.debug("%s: Album requests: %s" % (PLUGIN_NAME, album._requests))
+            if tangoinfo_album_data:
+                tm = track.metadata
+                if tm["discnumber"]:
+                    discnr = str(tm["discnumber"])\
+                            .split("/")[0]\
+                            .lstrip("0")
+                            # For discnumbers like 02/12"
+                else:
+                    discnumber = "1"
```
mineo: Again, `discnumber` vs. `discnr`.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+                        xml=False, priority=False, important=False)
+
+
+    def website_process(self, barcode, response, reply, error):
+        # for using xmlws
+
+        if error:
+            log.error("%s: %s: Network error retrieving info for barcode", PLUGIN_NAME, barcode)
+            tuples = self.albumpage_queue.remove(barcode)
+            for track, album in tuples:
```
mineo: This code (and the same one in lines 158 & 159) looks suspicious to me: `album._requests` is incremented exactly once per album (/barcode), but is decremented for every track on the album, which is going to be more than once in most cases.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+                    if self.albumpage_cache[barcode][tint]['vocal']:
+                        fm['vocal'] = self.albumpage_cache[barcode][tint]['vocal']
+
+
+    def album_add_request(self, album):
+        album._requests += 1
+
+    def album_remove_request(self, album):
+        album._requests -= 1
+        if album._requests == 0:
```
mineo: `_finalize_loading` already performs this test.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+            album._finalize_loading(None)
+
+
+    def barcode_process_metadata(self, barcode, response):
+
+        #: Check whether we have a concealed 404 and get the homepage
+        if response.splitlines()[1] == "<title>Contents - tango.info</title>":
+            log.debug("%s: No album with barcode %s on tango.info" % (PLUGIN_NAME, barcode))
+            return None
+        else:
```
mineo: No need to put this into an else branch and the additional indentation level of that, the line before is a `return`.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+            regexp = """<h2><a href="\/tracks">Tracks<\/a><\/h2>(?!<\/table>)(.+?)<\/table>"""
+            trs = re.compile("<tr>((?!<\/tr>).+?)</tr>")
```
mineo: At the moment, these are recompiled on each call to `barcode_process_metadata` (ok, that's a lie, the `re` module caches them, but this still means a cache lookup somewhere). If you make these global constant variables, compiling would only happen once and they wouldn't be buried in this function.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+            return
+
+        tangoinfo_album_data = self.barcode_process_metadata(barcode, response)
+        #log.debug("Album data: %s" % tangoinfo_album_data)
+
+        self.albumpage_cache[barcode] = tangoinfo_album_data
+        tuples = self.albumpage_queue.remove(barcode)
+        #log.debug("%s: %r: tango.info page = http://tango.ingo/0%s", PLUGIN_NAME, barcode, barcode)
+        for track, album in tuples:
+            self.album_remove_request(album)
```
mineo: This means that `album._finalize_loading` will be called before the data from tango.info has been written on all tracks.

plugins/tangoinfo/tangoinfo.py
Sophist-UK: This code is based on my Artist Official Website plugin - so I understand how it works.

Because this is a Track Metadata processor, album._requests is incremented once per track. To avoid duplicated requests for the same metadata, the tracks are queued against each request and when the request is completed, it is decremented once per track.

plugins/tangoinfo/tangoinfo.py
Sophist-UK: Ah yes - that is probably also an issue in Artist Website plugin I guess.

plugins/tangoinfo/tangoinfo.py
mineo: Oh, now I see how it works. Seems a bit weird to count the number of not-really-started requests, but as long as the other direction is aware of it...

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
+PLUGIN_NAME = 'Tango.info Adapter'
+PLUGIN_AUTHOR = 'Felix Elsner'
+PLUGIN_DESCRIPTION = '<p>Get genre, performed date and vocalists from tango.info</p>\
+<p>Uses web scraping since opendata is disabled for now</p>'
+PLUGIN_VERSION = '0.1.3'
+PLUGIN_API_VERSIONS = ["0.15.0", "0.15.1", "0.16.0", "1.0.0", "1.1.0", "1.2.0", "1.3.0"]
+PLUGIN_LICENSE = 'ISC License'
```
mineo: Is relicensing GPL code under the ISC license allowed?

plugins/tangoinfo/tangoinfo.py
Sophist-UK: It's complicated. An artist website request can be for some but not all tracks on an album (e.g. Various Artists), and for tracks across multiple albums. So its easier to increment and decrement once per track than to work out whether you need to increment / decrement because e.g. this is the first track on the album etc.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+    def barcode_process_metadata(self, barcode, response):
+
+        #: Check whether we have a concealed 404 and get the homepage
+        if response.splitlines()[1] == "<title>Contents - tango.info</title>":
+            log.debug("%s: No album with barcode %s on tango.info" % (PLUGIN_NAME, barcode))
+            return None
+        else:
+            regexp = """<h2><a href="\/tracks">Tracks<\/a><\/h2>(?!<\/table>)(.+?)<\/table>"""
```
Sophist-UK: You should also compile `regexp` and not used it in a raw `re` call.

plugins/tangoinfo/tangoinfo.py
ix5: Wouldn't it throw a keyerror if name wasn't in self.queue?

plugins/tangoinfo/tangoinfo.py
ix5: Fixed

plugins/tangoinfo/tangoinfo.py
ix5: Will picard always return a discnumber? Or could the field also be empty? I have only added a check if discnumber exists, else it gets "1", because I need to construct the TINT.

plugins/tangoinfo/tangoinfo.py
ix5: Yes, I just copied it. Maybe it does not make as much sense as for official websites, because each album can only have on tango.info page.

plugins/tangoinfo/tangoinfo.py
ix5: Fixed, I put them outside the whole class, hope that is right.

plugins/tangoinfo/tangoinfo.py
ix5: Also fixed that in my plugin.

plugins/tangoinfo/tangoinfo.py
ix5: I had previously chosen ISC for my rough draft, but on the other PR @Sophist-UK told me to take inspiration from his `albumartist_website` plugin. Changed to GPLv2 since now it's mostly his code.

plugins/tangoinfo/tangoinfo.py
ix5: Done. Does it have that much of an impact on performance though? Would be interesting to know for future projects.

ix5:
Alright, thank you so much everyone. Updated version is here: https://github.com/ix5/picard-tangoinfo

plugins/tangoinfo/tangoinfo.py
Sophist-UK: I suppose you could simply store the request against an album and check if the request has already been queued or cached and then apply the response to all tracks in the album - but this might actually be more effort than it is worth.

ix5: And you'd still have to use a track processor and not an album one(individual genres, dates for each track etc.), so no real benefit there I guess. Also, it just occurred to me that there could be edge cases where MB and tango.info disagree, e.g. multi-volume compilations or something similar.
So your approach to queueing is the most sensible after all.
What irritates me is that if MB and tango.info both have new data(e.g date was 1950, MB has 2004 and tango.info 1950-01-01) the MB one takes preference.

ix5:
Strings from tango.info are now converted to unicode, so no more unfinished loading and hanging because of names like *Félix* (that was a rookie mistake).

I have thoroughly tested the plugin with much of my collection. **It works very reliably**. 
A lot of times it seems MB and tango.info disagree on something. Nothing I can do about that, but I have added a notice in the debug log so users can rectify inaccuracies.

Feel free to test this on a Windows machine; if you need to, I can provide some sample songs(or download some Carlos Gardel tracks, copyright has expired on his works afaik).

PR #57, opened 2016-08-08
Add new Playlist plugin

chinf:
My first attempt at a plugin for Picard - I started off basing this on the Cuesheet plugin but it's now quite different.

I tested this under Python 2.7.10 on several albums with a few permutations of non-ASCII data and also incomplete albums, but I expect more testing is worthwhile, and in particular I don't know if any changes are needed to ensure compatibility with Python 3.

I've run the test.py script OK.


chinf:
Just seen notes on other commits about plugins lacking licences - I'll add one to mine with my next commit (coming soon...).

PR #20, opened 2015-01-30
replaygain: Added license details

phw:
Contributors to this plugin are @phw (myself), @garyvdm, @lalinsky, @mineo, @zas and @DanMan

See https://github.com/musicbrainz/picard/commits/master/contrib/plugins/replaygain

Would be great if you could leave a short comment if you are fine with this licensing.

PR #60, opened 2016-10-05
Add Musixmatch plugin

baumea:
With this pull request i would like to add a musixmatch plugin. This plugin downloads 30% of the lyrics from Musixmatch via the Musixmatch API and tags the files accordingly. Due to licensing restrictions we cannot get more than 30% of the lyrics. To use the Musixmatch API one has first to obtain a API Key and copy the key into **init**.py. The plugin depends on the modules musixmatch (https://github.com/utstikkar/pyMusiXmatch), urrlib2, bisect, and Queue. These modules come bundled with that plugin.


Sophist-UK:
Funny enough a lyrics plugin was on my to-do list, but you have beaten me to it.

I was planning to use the Genius API to get the lyrics - have you looked at this?

Sophist-UK:
Having now looked more closely at the code, I can see why you have used MusixMatch - because this is integrating an existing MusixMatch python module into a Picard Plugin. And since MusixMatch already uses the musicbrainz track_id as an index, this is a good fit.

A few comments on the code (from a read through without having tried running it):

1. I know that you have included Queue because the MusixMatch code requires it - however this is only referenced in util.py and there it is only imported and as far as I can see is never actually used. This means that the MusixMatch code is synchronous - so Picard may become somewhat unresponsive when it is making these calls, though this is mitigated by it using a cache. However, I think it would be possible to create your own mechanism to queue the requests and process the response in a call-back - see my own plugin for artists website to see how to do this.

2. It would be great if you could include a Options / Options page to allow the user to put in their API key and store it in Picard's settings.

baumea:
no, i havent looked into the genius api, but would be nice to have :).

thanks for the comments!
1. should be easy to do, but for the options page i have no idea. i did not find a documentation. i looked into the lastfm plugin to get an idea, however the main problem seems to be that the API key has to be set as an environmental variable _before_ the musixmatch module is loaded. this means, when the plugin is loaded into picard it has to access the settings, and then load the modules.

baumea:
ok, thanks for your comments Sophist-UK. I managed to get both done.

zas:
@baumea : can you run autopep8 or the like on your files ? See https://www.python.org/dev/peps/pep-0008/

baumea:
ok, done


Am 12.01.2017 um 22:21 schrieb Laurent Monin <notifications@github.com<mailto:notifications@github.com>>:


@baumea<https://github.com/baumea> : can you run autopep8 or the like on your files ? See https://www.python.org/dev/peps/pep-0008/

zas:
> ok, done

Thanks, much better, there were a lot of issues with indentation & spaces.

plugins/musixmatch/__init__.py
```diff
@@ -0,0 +1,59 @@
+PLUGIN_NAME = 'Musixmatch Lyrics'
+PLUGIN_AUTHOR = 'm-yn'
+PLUGIN_DESCRIPTION = 'Fetch first 30% of lyrics from Musixmatch'
+PLUGIN_VERSION = '0.2'
+PLUGIN_API_VERSIONS = ["0.9.0", "0.10", "0.15"]
```
zas: @baumea : can you please add license details (for an example, see https://github.com/metabrainz/picard-plugins/pull/20/files) ?

baumea:
license details added

Sophist-UK:
@baumea Did you look into Genius API for lyrics?

baumea:
not really, no

Sophist-UK:
Pity. If you can get it to work it would provide full lyrics.

PR #63, opened 2016-10-29
Plugin to query wikidata.org and retrieve the genre

dns53:
This plugin first queries musicbrainz to see if there is a wikidata link.
This then queries wikidata and looks for property P136 that represents a genre.


plugins/wikidata/wikiata.py
```diff
@@ -0,0 +1,84 @@
...
+                        for relation in response.metadata[0].release_group[0].relation_list[0].relation:
+                            if relation.type == 'wikidata' and 'target' in relation.children:
+                                wikidata_url=relation.target[0].text
+                                self.process_wikidata(wikidata_url,metadata)
+    def process_wikidata(self,wikidata_url,metadata):
+        item=wikidata_url.split('/')[4]
+        path="/wiki/Special:EntityData/"+item+".rdf"
+        self.xmlws.get('www.wikidata.org', 443, path,
+                       partial(self.parse_wikidata_response, item,metadata),
+                                xml=True, priority=True, important=False)
```
Sophist-UK: Priority and Important relate to the queue for the particular domain (in this case Wikidata). So I recommend Priority=False here.

plugins/wikidata/wikiata.py
```diff
@@ -0,0 +1,84 @@
...
+                    if 'about' in node1.attribs:
+                        if node1.attribs.get('about') == 'http://www.wikidata.org/entity/%s' % item:
+                            for key,val in node1.children.items():
+                                if key=='P136':
+                                    for i in val:
+                                        if 'resource' in i.attribs:
+                                            tmp=i.attribs.get('resource')
+                                            if 'entity' ==tmp.split('/')[3] and len(tmp.split('/'))== 5:
+                                                genre_id=tmp.split('/')[4]
+                                                print 'Found the wikidata id for the genre: %s' % genre_id
```
Sophist-UK: Use log.debug instead of Print.

plugins/wikidata/wikiata.py
```diff
@@ -0,0 +1,84 @@
...
+                                                genre_entries.append(tmp)
+                        else:
+                            for tmp in genre_entries:
+                                if tmp == node1.attribs.get('about'):
+                                    list1=node1.children.get('name')
+                                    for node2 in list1:
+                                        if node2.attribs.get('lang')=='en':
+                                            genre=node2.text
+                                            genre_list.append(genre)
+                                            print 'Our genre is: %s' % genre
```
Sophist-UK: Use log.debug instead of print and start it with "Wikigenre: "

Sophist-UK:
Needs to be made Python 3 compliant to pass Travis automated tests.

Also, since genre is added to the ALBUM metadata asynchronously, are we sure that this makes it through to the track metadata.

I think it may also need code to delay album finalisation until the lookup has completed.

(Probably does not make sense to cache the responses as it would seem unlikely that the same call will need to be reused.)

dns53:
It is finally passing the unit tests, please take another look and let me know if there is anything missing.

I am planning on having this as a track metadata plugin and looking up wikidata url's linked to artists and works but this is a start.

Sophist-UK:
If you want it as a track metadata plugin, but looking for album or artist metadata, then you don;t want to be making 20+ calls for the same album or artist - so you need to consolidate calls and add results to all tracks that have been consolidated into a single call and cache the results.

See my Artist Official Website plugin for an example of how to do this.

dns53:
I think this version is working good enough to be merged.
What needs to be done before this can be merged and become an official plugin?

mwiencek:
If it's working for you, I'd say it's good enough to merge. 👍 

Can you just add `PLUGIN_LICENSE` and `PLUGIN_LICENSE_URL`? For example, https://github.com/metabrainz/picard-plugins/blob/173f48e956318f813aa92fce8b8f61cfef8404b0/plugins/padded/padded.py#L11-L12 (Picard itself is GPL-2+, but you can choose whatever open-source license you like.)

PR #69, opened 2016-12-24
Update plugins to Qt5

samj1912:


mwiencek:
We probably ought to create a v1 branch with the old copies for people running Picard 1.x, but that can be done later.

https://github.com/metabrainz/picard-website still needs to be modified so that it doesn't serve incompatible plugins. Currently, it always serves the latest version of the plugin. I think we can do this by getting the Picard version from the user-agent string.

And assuming we have a separate v1 branch, we'll need to handle that too, probably as a separate json file.

Freso:
I get the below output using no_release.zip built from this PR:

```python-traceback
  File "/usr/lib/python2.7/site-packages/picard/ui/mainwindow.py", line 758, in show_options
    dialog = OptionsDialog(page, self)
  File "/usr/lib/python2.7/site-packages/picard/ui/options/dialog.py", line 104, in __init__
    page = Page(self.ui.pages_stack)
  File "/home/freso/.config/MusicBrainz/Picard/plugins/no_release.zip/no_release.py", line 89, in __init__
  File "/home/freso/.config/MusicBrainz/Picard/plugins/no_release.zip/no_release.py", line 45, in setupUi
  File "/home/freso/.config/MusicBrainz/Picard/plugins/no_release.zip/no_release.py", line 49, in retranslateUi
AttributeError: type object 'QApplication' has no attribute 'UnicodeUTF8'
```

PR #70, opened 2016-12-27
Fix indentation

samj1912:


plugins/wikidata/wikidata.py
```diff
@@ -19,6 +19,7 @@
 from functools import partial
 import threading
 
+
```
Freso: I guess it should be "Fix whitespace" rather than indentation… 😏

samj1912: Ah no, indentation fixes in L179,180

Issue #25, opened 2015-03-21
Add cluster as release sends negative track numbers


l-g:
I sometimes see

 this error

```
Internal Server Error
Oops, something went wrong!

Error:

Caught exception in MusicBrainz::Server::Controller::ReleaseEditor->add "CGI param clash for mediums.0.track.-1.name=-1 at lib/MusicBrainz/Server/Controller/ReleaseEditor.pm line 111"
Stack trace:
line 111 MusicBrainz::Server::Controller::ReleaseEditor
line 44 MusicBrainz::Server::Controller::ReleaseEditor
line 104 MusicBrainz::Server::Controller::ReleaseEditor
line 330 MusicBrainz::Server
line 304 MusicBrainz::Server
line 331 MusicBrainz::Server
line 345 MusicBrainz::Server
line 391 MusicBrainz::Server
Request data:

$VAR1 = {
          'query_parameters' => {},
          'body_parameters' => {
                                 'mediums.0.track.3.length' => '78994',
                                 'mediums.0.track.8.name' => 'Marianne, My Pen Pal',
                                 'mediums.0.track.5.name' => 'Sewing Lax Marionettes',
                                 'mediums.0.track.11.name' => 'Who Are You?',
                                 'mediums.0.track.2.name' => 'Goose-skin',
                                 'mediums.0.track.7.length' => '103235',
                                 'mediums.0.track.6.name' => 'You Are My Favourite',
                                 'mediums.0.track.6.length' => '86282',
                                 'mediums.0.track.11.length' => '165773',
                                 'mediums.0.track.7.name' => 'King Carl Wants To Stay Home',
                                 'mediums.0.track.-1.name' => 'Skogspromenad',
                                 'mediums.0.track.1.length' => '170370',
                                 'mediums.0.track.1.name' => 'Tango in the Night',
                                 'mediums.0.track.10.length' => '144248',
                                 'mediums.0.track.4.name' => 'You Fell In Love With My Best Friend',
                                 'mediums.0.track.5.length' => '127451',
                                 'mediums.0.track.10.name' => 'Give Me A Break!',
                                 'mediums.0.track.12.name' => 'Crazy Love',
                                 'mediums.0.track.12.length' => '116349',
                                 'mediums.0.track.4.length' => '104463',
                                 'mediums.0.track.9.name' => 'Art Is Heart',
                                 'mediums.0.track.2.length' => '80195',
                                 'mediums.0.track.-1.length' => '71948',
[...]

PR #71, opened 2016-12-28
fixed negative track problem (Issue #25)

vibhorgupta96:
Did the same thing as:- 
https://github.com/metabrainz/picard/commit/d38104821654198f5b2a2f512b2d37a4b7b6ff63
for track numbers.


mineo:
I'm not 100% sure about this, because I think it needs to take into account
that the track numbers on different mediums might have different
oddities/might not all be in need of repair.

PR #73, opened 2017-01-13
 Plugin tangoinfo: fix breaking typo, correct sanitization

ix5:
I looked over this once again since @zas merged the plugin recently and noticed some grave oversight on my part:
- `reponse` -> `response` (was breaking `barcode_process_metadata()`)
- Correctly sanitize vocalist(s) html
- Retry with a zero prepended to the barcode twice instead of once
- Will not fail on python3 any more because of `unicode`

Sorry for causing this!

### DIFF
--- a/plugins/tangoinfo/tangoinfo.py
+++ b/plugins/tangoinfo/tangoinfo.py
@@ -1,12 +1,13 @@
 # -*- coding: utf-8 -*-
 PLUGIN_NAME = "Tango.info Adapter"
 PLUGIN_AUTHOR = "Felix Elsner"
-PLUGIN_DESCRIPTION = "\
-<p>Load genre, date and vocalist tags from the online database tango.info.</p>\
-<p>This plugin uses web scraping, but only once per album. In so doing \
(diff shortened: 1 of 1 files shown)


plugins/tangoinfo/tangoinfo.py
```diff
@@ -216,12 +223,12 @@ def album_remove_request(self, album):
     def barcode_process_metadata(self, barcode, response):
 
         # Check whether we have a concealed 404 and get the homepage
-        if "<title>Contents - tango.info</title>" in reponse:
+        if "<title>Contents - tango.info</title>" in response:
```
ix5: Can’t have that

plugins/tangoinfo/tangoinfo.py
```diff
@@ -216,12 +223,12 @@ def album_remove_request(self, album):
     def barcode_process_metadata(self, barcode, response):
 
         # Check whether we have a concealed 404 and get the homepage
-        if "<title>Contents - tango.info</title>" in reponse:
```
ix5: Can’t have that breaking everything

Issue #77, opened 2017-01-29
Add Cluster As Release: HTTPS redirection loses form data on MusizBrainz.org


AltarBeastiful:
When posting the form data to MusizBrainz.org, the plugin gets redirected from HTTP to HTTPS, losing the form data in the process.

The plugin should post to MusizBrainz on HTTPS as it is the preferred method to interact with the website.

This can be manually fixed by setting the server port to 443, but the default to 80 actually breaks this plugin.

I don't see a good way to guess if we can use https other than port without breaking compatibility for local server not using https. Should we ask for a change in the default settings of picard?

Sophist-UK:
https://github.com/metabrainz/picard/pull/600 has already been proposed for this.

mineo:
This has already been reported as [PICARD-693](https://tickets.metabrainz.org/browse/PICARD-963) and is not fixed by the pull request mentioned above.

Freso:
How will PICARD-693 not be fixed by #600?

AltarBeastiful:
https://github.com/metabrainz/picard/pull/600 fixes `build_qurl` but the plugin doesn't use it. 
This question has just been answered [in the official ticket](https://tickets.metabrainz.org/browse/PICARD-963?focusedCommentId=41989&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-41989). 

Closing as a duplicate of [PICARD-963](https://tickets.metabrainz.org/browse/PICARD-963).

PR #78, opened 2017-01-29
PICARD-963: addrelease: Always use HTTPS for MusicBrainz

Freso:
https://tickets.metabrainz.org/browse/PICARD-963 / https://github.com/metabrainz/picard-plugins/issues/77

https://tickets.metabrainz.org/browse/MBH-466 broke addrelease.py's functionality for users submitting to musicbrainz.org on port 80 (the default settings!), as HTTP 302 redirects don't work properly for POST requests. This PR makes `addrelease.py` always use HTTPS for MusicBrainz.org servers, which alleviates the issue.

### DIFF
--- a/plugins/addrelease/addrelease.py
+++ b/plugins/addrelease/addrelease.py
@@ -11,6 +11,7 @@ PLUGIN_API_VERSIONS = ["1.0.0"]
 
 from picard import config
 from picard.cluster import Cluster
+from picard.const import MUSICBRAINZ_SERVERS
 from picard.file import File
 from picard.util import webbrowser2
 from picard.ui.itemviews import BaseAction, register_cluster_action, register_file_action
@@ -39,7 +40,7 @@ HTML_ATTR_ESCAPE = {
 def mbserver_url(path):
     host = config.setting["server_host"]
     port = config.setting["server_port"]
-    if port == 443:
+    if host in MUSICBRAINZ_SERVERS or port == 443:
         urlstring = "https://%s%s" % (host, path)
     elif port is None or port == 80:
         urlstring = "http://%s%s" % (host, path)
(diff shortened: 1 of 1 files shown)


Sophist-UK:
I haven't looked at the code...

But would it be worth changing this to use `build_qurl` and avoid future issues like this?

Freso:
Well, `build_qurl` might be an option, and I considered it too, but until https://github.com/metabrainz/picard/pull/600 is merged, that will not fix the issue for users. The way I did in this patch will assure that the plugin continues to work for users of Picard 1.3.2 and earlier.

Sophist-UK:
https://github.com/metabrainz/picard/pull/600 looks like it will be merged for 1.4.

And since people can only upgrade plugins in 1.3.2 manually, the easiest  thing might be to make it use qurl and just mark the new version of this plugin as requiring 1.4. After all there are plenty of good reasons to move to 1.4 in addition to this plugin.

Freso:
But people on 1.3.2 can still manually update the plugin. And some people will not want to use a Picard not shipped with their distro and Picard 1.4 will never make it into some distro releases (e.g., any LTS Ubuntu already shipped).

Note that I am not against using `build_qurl()`, but this PR fixes the issue and fixes it *now*. [«"Good enough" is good enough»](https://www.youtube.com/watch?v=_Ek3A2b-nHU).

There are several changes I want to make to addrelease.py as it is (using `set_…()` and `get_…()` methods? really? (and yes, I wrote that code!)), so replacing the URL building with `build_qurl()` will likely come along with those. But that's a more invasive change than what's required as of this moment.

zas:
Imho, this fix is enough, `build_qurl()` would break 1.3.2 compatibilty.

Sophist-UK:
I am convinced.

PR #79, opened 2017-02-09
Build plugin files for specific api versions

samj1912:


.gitignore
```diff
@@ -1,11 +1,12 @@
 # Generated by the script
 plugins.json
 plugins/*.zip
+v?.?/
```
zas: Hmmm, i don't like this much. It matches `va.b/` but not `v2.4.5/`.
What about adding a top level directory, for example:
```
packaged_plugins/
├── v10.2.3
├── v1.2
└── v2.0
```

And ignore `/packages_plugins/**`

Hmmm, not even sure why this is needed.

zas: @samj1912 : ^^

generate.py
```diff
@@ -96,14 +98,21 @@ def zip_files():
...
-            build_json()
-        elif sys.argv[1] == "zip":
-            zip_files()
+    parser = argparse.ArgumentParser(description='Generate plugin files for Picard website.')
+    parser.add_argument('version', nargs='?')
+    parser.add_argument('--pull', action='store_true', dest='pull')
+    parser.add_argument('--no-zip', action='store_false', dest='zip')
+    parser.add_argument('--no-json', action='store_false', dest='json')
+    args = parser.parse_args()
+    if args.version == 'v1' or args.version is None:
```
zas: `version` argument should take a version in the format 'a.b.c' not 'va.b.c' imho.

.gitignore
zas: In fact, i think we should have:
```
packaged_plugins/
├── 10.2.3
├── 1.2
└── 2.0
```

generate.py
```diff
@@ -96,14 +98,22 @@ def zip_files():
...
+    parser.add_argument('--no-json', action='store_false', dest='json')
+    args = parser.parse_args()
+    if args.version is None:
+        call(["git", "checkout", "-q", 'master', '--', 'plugins'])
     else:
-        # call(["git", "pull", "-q"])
-        build_json()
-        zip_files()
+        call(["git", "checkout", "-q", args.version, '--', 'plugins'])
+    if args.version and not os.path.exists(os.path.join(args.build_dir, args.version)):
```
zas: Since we'll change the web API, i think we can make `version` a mandatory argument, and get rid of the case of no version.

Do we make any other use of `version` than for building paths ? If no, we could build the absolute destination path here, and pass it: `build_json(destdir)` & `zip_files(destdir)`

I say absolute, because we may want to write `build/` outside the git dir... it will remove the need to ignore `build` directory, and is a better practice anyway.

generate.py
```diff
@@ -96,14 +98,22 @@ def zip_files():
...
-        elif sys.argv[1] == "zip":
-            zip_files()
+    parser = argparse.ArgumentParser(description='Generate plugin files for Picard website.')
+    parser.add_argument('version', nargs='?')
+    parser.add_argument('--build_dir', default="build")
+    parser.add_argument('--pull', action='store_true', dest='pull')
+    parser.add_argument('--no-zip', action='store_false', dest='zip')
+    parser.add_argument('--no-json', action='store_false', dest='json')
+    args = parser.parse_args()
+    if args.version is None:
```
zas: What about something like:

```python
version2branch = {
  None: 'master',
  '2.0': 'my_branch_2'
}

try:
  call(["git", "checkout", "-q", version2branch[args.version], '--', 'plugins'])
except KeyError:

....
````

File: generate.py [hunk 5]
```diff
@@ -96,14 +98,22 @@ def zip_files():
...
-        elif sys.argv[1] == "json":
-            build_json()
-        elif sys.argv[1] == "zip":
-            zip_files()
+    parser = argparse.ArgumentParser(description='Generate plugin files for Picard website.')
+    parser.add_argument('version', nargs='?')
+    parser.add_argument('--build_dir', default="build")
+    parser.add_argument('--pull', action='store_true', dest='pull')
+    parser.add_argument('--no-zip', action='store_false', dest='zip')
+    parser.add_argument('--no-json', action='store_false', dest='json')
```
zas: Perhaps worth to add a `--help` ;)

generate.py
```diff
@@ -49,18 +52,18 @@ def build_json():
             print("Added: " + dirname)
             data['files'] = files
             plugins[dirname] = data
-
-    with open(plugin_file, "w") as out_file:
+    out_path = os.path.join(dest_dir, plugin_file)
```
zas: s/out_path/plugin_file_path/ ?

test.py
```diff
@@ -1,26 +1,17 @@
-from __future__ import print_function
```
zas: Not needed for python 2 ?

samj1912: Since we are only using print(".."), nope.

PR #80, opened 2017-02-09
PICARD-983, PICARD-984: Fix wikidata plugin genre merges and ui problems

antlarr:
This fixes the plugin detection of duplicate genres so they're merged correctly
instead of resulting on a genre list like ["Rock; pop", "pop"]
Also fixes the freeze of the message "loading album information" which never finished when an album was refreshed.
It uses title() for wikidata genres so they can be compared with folksonomy tags to find duplicates.
Finally, it also adds a encoding python header since the file contains non-ASCII characters.

https://tickets.metabrainz.org/browse/PICARD-983
https://tickets.metabrainz.org/browse/PICARD-984

### DIFF
--- a/plugins/wikidata/wikidata.py
+++ b/plugins/wikidata/wikidata.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Copyright © 2016 Daniel sobey <dns@dns.id.au >
 
 # This work is free. You can redistribute it and/or modify it under the
--- a/plugins/wikidata/wikidata.py
+++ b/plugins/wikidata/wikidata.py
@@ -55,7 +55,13 @@ class wikidata:
         if item_id in self.cache.keys():
             log.info('WIKIDATA: found in cache')
             genre_list=self.cache.get(item_id);
-            metadata["genre"] = genre_list
+            new_genre = metadata.getall("genre")
+
+            for str in genre_list:
+                if str not in new_genre:
+                    new_genre.append(str)
+                    log.debug('WIKIDATA: appending genre %s' % str)
+            metadata["genre"] = new_genre
             
             if tagger._requests==0:
                 tagger._finalize_loading(None)
--- a/plugins/wikidata/wikidata.py
+++ b/plugins/wikidata/wikidata.py
@@ -8,7 +8,7 @@
 PLUGIN_NAME = 'wikidata-genre'
 PLUGIN_AUTHOR = 'Daniel Sobey'
 PLUGIN_DESCRIPTION = 'query wikidata to get genre tags'
-PLUGIN_VERSION = '0.1'
+PLUGIN_VERSION = '0.2'
 PLUGIN_API_VERSIONS = ["0.9.0", "0.10", "0.15"]
 PLUGIN_LICENSE = 'WTFPL'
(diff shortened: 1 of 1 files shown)


samj1912:
You might want to bump the plugin version since it changes the plugin behaviour.

plugins/wikidata/wikidata.py
```diff
@@ -54,7 +55,13 @@ def process_request(self,metadata,tagger,item_id,type):
         if item_id in self.cache.keys():
             log.info('WIKIDATA: found in cache')
             genre_list=self.cache.get(item_id);
-            metadata["genre"] = genre_list
+            new_genre = metadata.getall("genre")
+
+            for str in genre_list:
```
samj1912: Can you rename this to a better looping variable? str is a python keyword.

samj1912:
On an unrelated side note, just took a look at the plugin code, the amount of nesting is abysmal. Some of it could be refactored into functions. Apart from that I can see some private variables being accessed directly(also not good). There are also a lot of PEP-8 errors.

plugins/wikidata/wikidata.py
antlarr: Sure!, I also thought about that but I usually try not to change other people's code when it's not absolutely needed.

plugins/wikidata/wikidata.py
```diff
@@ -54,7 +55,13 @@ def process_request(self,metadata,tagger,item_id,type):
         if item_id in self.cache.keys():
             log.info('WIKIDATA: found in cache')
             genre_list=self.cache.get(item_id);
-            metadata["genre"] = genre_list
+            new_genre = metadata.getall("genre")
+
+            for genre in genre_list:
+                if genre not in new_genre:
+                    new_genre.append(genre)
```
zas: If order isn't important, you may use a `set()` here instead of a list.

plugins/wikidata/wikidata.py
```diff
@@ -173,12 +181,11 @@ def parse_wikidata_response(self,item,item_id, response, reply, error):
...
-                new_genre.append(metadata["genre"])
-                for str in genre_list:
-                    if str not in new_genre:
-                        new_genre.append(str)
-                        log.debug('WIKIDATA: appending genre %s' % str)
+                new_genre = metadata.getall("genre")
+                for genre in genre_list:
+                    if genre not in new_genre:
+                        new_genre.append(genre)
+                        log.debug('WIKIDATA: appending genre %s' % genre)
```
zas: Aren't those few lines a duplication of the lines around 58-63 ?
If yes, better create a small function for that.

samj1912: Yeah, the duplication was present in the previous code too. There is a lot of duplication elsewhere too. This plugin can improve a lot code quality wise IMO, but maybe that is out of the scope of this PR?

plugins/wikidata/wikidata.py
antlarr: Yes, I've changed it to use a set and also the duplicated lines below, since the code is much shorter now, I don't think it's worth it to create a small function for that.

plugins/wikidata/wikidata.py
antlarr: I agree. Improving the code quality of the plugin in general is out of the scope of this PR :)

PR #87, opened 2017-03-21
Support generating multiple cue files for releases with multiple discs

antlarr:
If a release had multiple discs, a single cue file was written with all tracks. With this PR, a separate cue file is generated for each disc, each having its own tracks.


zas (changes requested):
Please bump the plugin version

PR #94, opened 2017-04-10
New plugin - workparts

MetaTunes:
Adds tags for multiple part levels for (particularly) classical music.
Version 0.1 for testing.
See readme for details.


Sophist-UK:
How does this relate to https://tickets.metabrainz.org/browse/PICARD-1043 and [Picard#676](https://github.com/metabrainz/picard/pull/676) / [Picard#695](https://github.com/metabrainz/picard/pull/695)?

plugins/workparts/Readme.md
```diff
@@ -0,0 +1,11 @@
+This is the first version of "workparts.py".
+It populates the metadata tags with information from the MusicBrainz database about the work(s) of which the track is a recording, and the work(s) of which they are a part, passing up through mutiple work-part levels until the top is reached.
+This is particularly designed to assist with tagging of classical music so that players or library managers which can display multiple work levels can have access to them.
```
Sophist-UK: If you want players to recognise this information, the names of tags need to be agreed by all parties. These names are, for example, different to those proposed by Paul Taylor, the author of Jaikoz / Song Kong. See linked Picard ticket and comments in the first PR for more links and discussion around this.

plugins/workparts/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+This first version adds custom fields for work_0 (& workId_0) being the bottom level and work_parent_n (& work_parentId_n) for each successive level n above that.
+If there is more than one work at the bottom level, then it should create work_0.1 etc, but I haven't found a suitable album to test that (and arguably it is bad style to have more than one...). If there is more than one "parent" work of a lower level work, it uses the one with the longest name, on the grounds that the longest-named is likely to be the lowest level; this might possibly be improved.
+Issues were encountered with the Picard API in that there is not a documented way to let Picard know that it is still doing asynchronous tasks in the background and has not finished processing metadata. Many thanks to Daniel Sobey for assistance in dealing with this and to Sophist for the albumartist_website code which I have used extensively. I have tried to add some more comments to help any others trying the same techniques. There are also extensive log.info and log.debug statements which could be slimmed down once testing is complete.
```
Sophist-UK: There are ways to do tell Picard that you have other calls - my albumartist_website does that - but it is not nicely architected.

As you say the log.debug messages are extremely verbose and likely to be a big distraction if debugging something else.

plugins/workparts/Readme.md
```diff
@@ -0,0 +1,11 @@
+This is the first version of "workparts.py".
+It populates the metadata tags with information from the MusicBrainz database about the work(s) of which the track is a recording, and the work(s) of which they are a part, passing up through mutiple work-part levels until the top is reached.
+This is particularly designed to assist with tagging of classical music so that players or library managers which can display multiple work levels can have access to them.
+This first version adds custom fields for work_0 (& workId_0) being the bottom level and work_parent_n (& work_parentId_n) for each successive level n above that.
+If there is more than one work at the bottom level, then it should create work_0.1 etc, but I haven't found a suitable album to test that (and arguably it is bad style to have more than one...). If there is more than one "parent" work of a lower level work, it uses the one with the longest name, on the grounds that the longest-named is likely to be the lowest level; this might possibly be improved.
```
Sophist-UK: See discussion in musicbrainz/picard#676 for examples of recordings of multiple works or works with multiple parents. This is NOT bad style BTW.

You should be careful about assuming that work lengths are longer for lower level works. There will be bound to be counter-examples, and sometimes it is better to look up and record both parents rather than make a guess or assumption.

Alternatively, you could look for parents common to more than one track and use those.

plugins/workparts/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+Issues were encountered with the Picard API in that there is not a documented way to let Picard know that it is still doing asynchronous tasks in the background and has not finished processing metadata. Many thanks to Daniel Sobey for assistance in dealing with this and to Sophist for the albumartist_website code which I have used extensively. I have tried to add some more comments to help any others trying the same techniques. There are also extensive log.info and log.debug statements which could be slimmed down once testing is complete. 
+Planned enhancements (among others) are 
+(a) to include discrimination as to type of parts relationship (i.e. exclude irrelevant parents)
+(b) to add arranger and any other relevant metadata relating to a work which is not in the standard Picard tags
```
Sophist-UK: Not sure what you are getting at here. I am going to work on making Picard be more comprehensive with its tags - please add details of tags you think might be missing as a comment to https://tickets.metabrainz.org/browse/PICARD-1048.

Sophist-UK:
I suggest you rename this Classical_Workparts to help make it clear it is really for classical music where works can be parts of bigger works.

Sophist-UK:
I have not made any comments on my read through of the code because, leaving aside the debug code and without testing it, it looks reasonable. Well done.

plugins/workparts/Readme.md
MetaTunes: I deliberately called the tags something different for now, so that it would not be assumed that the plugin had been endorsed by Paul and in order to avoid confusion. The plan was to try and align them after discussion as there are a few wrinkles to be addressed. e.g. :
1. The Picard Work tag is called Musicbrainz_work_composition in SongKong. SongKong uses the Work tag for a name derived from the title.
2. SongKong only permits 6 intermediate levels (which frankly ought to be enough!). My plugin theoretically has no limit.
3. The Musicbrainz_work.... tags generated by SongKong are the exact names in MB. My plan was to derive some names which avoid unnecessary repitition when displayed in the library manager/player. At present, I achieve this with an Mp3tag action and planned to incorporate a script in SongKong when Paul provides a full scripting capability (discussed with him). Thus there are potentially two different types of tags - full or part names.
4. I also planned to construct an "alternative" title, particularly for software that uses embedded colons (or other symbols) to differentiated the parts.
See the discussion at http://musoware.com/forums/viewtopic.php?f=2&t=387  to see what I am trying to achieve.
I agree that a consistent naming convention would be helpful (although renaming is always possible via a script).

plugins/workparts/Readme.md
MetaTunes: I agree that the use of longer names is not ideal. The plan was to find a better way of creating the "longest path" to the top once I (or someone else) had a better idea. Thanks for the comments/suggestions.

plugins/workparts/Readme.md
MetaTunes: As you realise I followed your code, in the absence of anything else I could find. Re the log.debug messages - I agree they should go. In the meantime they are clearly flagged with comments, so they are easy to delete or comment out.

plugins/workparts/Readme.md
MetaTunes: Arranger is the main one, I think. It is not entirely straightforward and I'll add comments as suggested.

MetaTunes:
Thanks for all the comments so far. Unfortunately it looks like I started on this just as something similar was happening elsewhere - I didn't spot that and no-one mentioned it in the community discussion (see https://community.metabrainz.org/t/classical-music-accessing-work-and-parts-structure/234786) either. No matter, it was an interesting exercise and has helped to flush out a few issues (at least in my mind). I'll add my comments on the other tickets and pause development for now - if the plugin is superseded by other developments then that is fine.

plugins/workparts/Readme.md
MetaTunes: Correction - it looks like arranger is added (at least in some circumstances).

Sophist-UK:
There is nothing wrong with starting with the community forums, but for future reference also worth checking tickets.metabrainz.org and asking in IRC freenode/#metabrainz before embarking on any code development.

Sophist-UK:
I haven't had time to read the latest code, however if it doesn't already replicate the code that got taken out of the Picard change that processes the original release XML, you might want to include it.

Before I took it out the MBXML code that populated iTunes Classical tags was in https://github.com/Sophist-UK/Sophist_picard/blob/bd5312b8b01485610767d75191892bc15d5d687d/picard/mbxml.py , so you might want to extract that and include it in your code as a starting point before you populate the tags with data.

Note: iTunes stupidly use the ID3 TIT1 tag for the overall work, and put Grouping into GRP1. If you want to mimic iTunes, then I guess you need to do the same (by copying Grouping into ~itunesgrouping and using Picard Grouping for the overall work as well as Work).

MetaTunes:
I'll look into that. I'd almost finished work on the new "classical work parts", so I may do a PR before looking into this. With the new version of the plugin, I will attempt to explain the philosophy and would appreciate further input to that, particularly in the light of the discussion re PICARD-1043. Just a few points in advance of that:
My approach is to provide the tags in as clear and unambiguous way as possble, and well documented in the readme, so that the user can do with them what they will (e.g via a tagger script). PT would like me to be consistent with the SongKong tags; my current preference is to use unique tag names but to also supply a tagger script that maps them to SongKong tags. 
As regards some of the plugin becoming superseded by Picard enhancements, that is fine, but if Picard restricts works to the base "recording of" plus one "parent" work, the plugin will still be needed. There are plenty of works where displaying multi-levels is helpful, particularly opera. Not many players can take advantage of the additional levels, but my favourite library manager, Muso, can and it works very well. 
I am not greatly enamoured of "movement number " or "movement total". I do not consider these to be well-defined. It is not clear whether the context is the original work or the recorded tracks (which may be a subset of movements) and I prefer to use the published numbering system (which may be Roman numerals or none a all) rather than add a system-generated Movement number. Sorry iTunes! In any case, iTunes shows the work and movement just fine without being given them explicitly.

Sophist-UK:
The whole overall work and part work thing is now on the agenda for the weekly MetaBrainz Developers chat which is on IRC Freenode #metabrainz at 8:00pm UK on Monday. Likely that this will only get discussed significantly past 8:30pm if you want to join us - your input would be welcome.

MetaTunes:
After the latest commits, the new "Classical Work Parts" plugin should be stable and well-documented, so I'd appreciate review.

MetaTunes:
This plugin will be replaced by a more comprehensive "classical extras" plugin

PR #96, opened 2017-04-11
New simpler version of Abbreviate Artist Sort

Sophist-UK:
Different output so different version rather than an upgrade.

Muh simpler algorithm based on regex.

### DIFF
--- a/plugins/abbreviate_artistsort2/abbreviate_artistsort2.py
+++ b/plugins/abbreviate_artistsort2/abbreviate_artistsort2.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# This is the Abbreviate Artist plugin for MusicBrainz Picard.
+# Copyright (C) 2013-2017 Sophist
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


samj1912:
@Sophist-UK please rebase this against the correct branch of PP that you want to merge this into.

mineo:
What's the difference to the existing abbreviate artist sort plugin? The pull request mentions different output, the plugins description just mentions that the output is written to different variables. If it's only the latter, I would prefer not having a completely separate plugin whose only difference can be achieved by a few `set` calls in the tagger script.

Sophist-UK:
@mineo Instead of `Bach, J.S.` it outputs `Bach JS` because:

a. Removing commas seems important when artists are generally separated by commas.
b. Removing periods seems important when classical artist strings are generally very long.

And instead of putting the output in new variables, it replaces existing variables so that it can coexist with e.g. Title Case, Standardise feat. etc.

I didn't want to replace the existing plugin because that would impact users tagging.

(Incidentally, the code utilises regex replacements rather than a convoluted string processing loop.)

Sophist-UK:
@samj1912 Rebase done.

sambhav:
LGTM :+1:

zas (approved):

sambhav:
You might want to consider renaming the plugin? or release it as a newer versio of the abbriviate plugin?

Sophist-UK:
I did both. But:

1. The way it needs to be used has changed - so it is not compatible with the existing version and can't be a version upgrade.
2. Yet it still does essentially the same thing just with better functionality and better performance. So it shouldn't be called something different.

Sophist-UK:
This is ready to merge - the codacy issues are for the in-built unit testing using assert statements that are not a genuine issue.

PR #97, opened 2017-04-13
Update plugins to 2.0 API

samj1912:


phw:
Why so many complete removals? Shouldn't we aim to update those, too?

samj1912:
I am going to eventually port all of them. Currently these are the ones ported. I want them to be added so people can start using plugins ASAP.

The rest will be ported over this week. You are welcome to help :)

mwiencek:
Shouldn't we just be able to make `PLUGIN_API_VERSIONS` an empty list instead of removing the files?

samj1912:
@mwiencek the generation script will still pick it up and it will be displayed in Picard?

mwiencek:
Is there a reason Picard displays incompatible plugins, or the generation script needs to pick them up? That was one of the problems I mentioned a while ago on IRC. Ideally I just think we shouldn't create these huge commits if we don't need to, since they inflate the size of the repo and make `git blame` harder to use.

samj1912:
Fair point, Ill update the generation script to reject the plugins not within compatible apis.

samj1912:
Updated this PR. Not deleting any files now, instead changed the generation script @ PW.

PR #99, opened 2017-04-17
abbreviate_artistsort: Get rid of print statements

mineo:
Picards logging module is already imported, there's no need to reinvent debug
levels etc. here

I've loaded a classical album with this plugin active, the plugin showed some warnings, but the release loaded, so I suppose everything still works.

### DIFF
--- a/plugins/abbreviate_artistsort/abbreviate_artistsort.py
+++ b/plugins/abbreviate_artistsort/abbreviate_artistsort.py
@@ -13,7 +13,6 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
 
-from __future__ import print_function
 PLUGIN_NAME = u"Abbreviate artist-sort"
 PLUGIN_AUTHOR = u"Sophist"
 PLUGIN_DESCRIPTION = u'''Abbreviate Artist-Sort and Album-Artist-Sort Tags.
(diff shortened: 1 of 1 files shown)


Sophist-UK:
I have no idea why I used print rather than log when I wrote this!!!

But don't forget to increment the version number.

zas (approved):

zas:
Those changes should be ported to 2.0 version too

plugins/abbreviate_artistsort/abbreviate_artistsort.py
```diff
@@ -228,13 +217,11 @@ def abbreviate_artistsort(tagger, metadata, track, release):
                         unsorts[i],
                         new_unsort,
                     )
-                    if _debug_level > 0:
-                        print()
-                        print("Error: Unsorted text for %s has changed from '%s' to '%s'!" % (unsortTag, unsorts[i], new_unsort))
-                        print()
+                    log.warning()
+                    log.warning("Error: Unsorted text for %s has changed from '%s' to '%s'!" % (unsortTag, unsorts[i], new_unsort))
+                    log.warning()
```
Freso, line 222: These empty `log.warning()`s are probably unneeded now, no?

zas: > These empty log.warning()s are probably unneeded now, no?

Yes, i missed those, totally unneeded ;)

Issue #101, opened 2017-04-19
addrelease's file creation is broken on Python 3


Freso:
When trying to create the temporary file for the HTML form, Python throws an exception:
```python-traceback
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/picard/ui/itemviews.py", line 46, in __callback
    self.callback(objs)
  File "/home/freso/.config/MusicBrainz/Picard/plugins/addrelease.py", line 109, in callback
    html_file = self.generate_html_file(self.form_values)
  File "/home/freso/.config/MusicBrainz/Picard/plugins/addrelease.py", line 92, in generate_html_file
    f.write(HTML_HEAD % (self.NAME, mbserver_url(self.submit_path)))
  File "/usr/lib/python3.6/codecs.py", line 377, in write
    self.stream.write(data)
TypeError: write() argument must be str, not bytes
```

Freso:
Merged into 2.0 branch.

PR #102, opened 2017-04-19
Fix "Add Cluster As Release" plugin's tempfile writing

Freso:
See https://github.com/metabrainz/picard-plugins/issues/101

Also replaces `f = (file opening); …; f.close()` with a `with (file opening) as f` and drops compatibility with Picard 1.x.

### DIFF
--- a/plugins/addrelease/addrelease.py
+++ b/plugins/addrelease/addrelease.py
@@ -79,22 +79,22 @@ class AddObjectAsEntity(BaseAction):
 
     def generate_html_file(self, form_values):
         (fd, fp) = tempfile.mkstemp(suffix=".html")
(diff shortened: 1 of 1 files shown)


zas (approved):

plugins/addrelease/addrelease.py
```diff
@@ -7,7 +7,7 @@
  the MusicBrainz database via the website by pre-populating artists,\
  track names and times."
 PLUGIN_VERSION = "0.7.1"
```
zas: Bump version ?

zas (changes requested):

zas (approved):

PR #105, opened 2017-06-01
Standardized Album Artist Name(s)

rdswift:
Plugin to add two new tags with standardized album artist name(s) for
use when both standardized and credited names are required.

StdAlbumArtist = Standardized name of first artist in the AlbumArtist.

StdAlbumArtists = Standardized names of all artists in the AlbumArtist,
separated by semi-colons.

### DIFF
--- a/plugins/stdartistname/LICENSE
+++ b/plugins/stdartistname/LICENSE
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.  This
+General Public License applies to most of the Free Software
+Foundation's software and to any other program whose authors commit to
+using it.  (Some other Free Software Foundation software is covered by
+the GNU Lesser General Public License instead.)  You can apply it to
+your programs, too.
+
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+this service if you wish), that you receive source code or can get it
+if you want it, that you can change the software or use pieces of it
+in new free programs; and that you know you can do these things.
+
+  To protect your rights, we need to make restrictions that forbid
+anyone to deny you these rights or to ask you to surrender the rights.
+These restrictions translate to certain responsibilities for you if you
+distribute copies of the software, or if you modify it.
+
+  For example, if you distribute copies of such a program, whether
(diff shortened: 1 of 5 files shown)


Sophist-UK:
A few comments:

1. Have you checked that the standardized artist names are not already in the XML provided for the album rather than making new requests for them? (You should not make extra WS calls if you do not need to as these will slow down album loading quite a lot.)

2. Because you are needing standardized names for (potentially) several artists, you will need to make several ws requests for artists and therefore probably need more coordinating code in your plugin to ensure you have all the artists before you finalise the album.

3. You should think about using AlbumArtists and StdAlbumArtists as a way to replace all the artists in AlbumArtist with standardised names rather than just the first one. See other plugins (like Smart Title Text or Abbreviated Artists v2) for code to use as a starting point.

rdswift:
I haven't been able to find any documentation with respect to writing plugins for Picard, so I've been going through the Picard source code to try to follow how the plugins are called and what information is passed.  I'll admit that I'm still a bit fuzzy on some aspects.

As to your questions:

1.  I hadn't, but after a quick look it appears that the information might be there as part of the included artist information.  That will definitely simplify things and eliminate the extra ws queries.  Thanks for bringing this to my attention.

2.  This took a lot of time and testing, but I think the coordinating code that I used as a basis (with a couple of small tweaks) is looking after that.  This is likely a moot point now, considering your point 1.

3.  If I understand your comment, I think that's already being done (sort of).  This plugin creates two tags:  (1) StdAlbumArtist which contains the standardized name for the first artist in musicbrainz_albumartistid; and (2) StdAlbumArtists which contains the standardized names for all artists in musicbrainz_albumartistid, separated by a semi-colon.  Perhaps it would be a bit more intuitive if the tags were changed to: (1) FirstStdAlbumArtist which contains the standardized name for the first artist in musicbrainz_albumartistid; and (2) StdAlbumArtist which contains the standardized names for all artists in musicbrainz_albumartistid, separated by the specified join phrase.

I have a need for the standardized name of the first artist as part of my file naming script, but others may not.  For that reason, I'm wondering if it would make more sense to do that in a separate plugin.  In any event, I'm going to close this pull request and re-work the code a bit before resubmitting.

Thanks for your comments.  I appreciate the insight into the way Picard works with plugins.

rdswift:
Will incorporate Sophist-UK's comments and resubmit.

Sophist-UK:
Bob:

1. If you want to see the XML returned to Picard do the following:
a. Turn on debugging - open the Help / Error & Debug log, and tick the checkbox
b. Load an album
c. Look in the debug log for the URL that Picard uses to get XML, and copy and paste this URL into your browser.

2. The documentation is definitely lacking - so using existing plugins as examples is your best bet. And navigating the XML to find what you want is a bit hot and miss to get right.

3. Next time, leave the PR open and just add commits to your own repo (and sync if you have cloned your repo onto your PC). Your changes will automatically be reflected in the PR.

rdswift:
1.  Good idea, but I already figured it out by going through the code for Picard.  Debug is still very useful.

2.  Sorting through the XML took a bit of time, but not too bad.  Checking for the joinphrase attribute tripped me up for a while because I forgot to check for its existence before trying to access it.

3.  ARGH!  It never crossed my mind that the PR would be automatically updated.  Something to remember for the future.  I'll try reopening it and see what happens.

Thanks again for your help to point me in the right direction.

rdswift:
Completely rewritten.  Will create a new pull request.

PR #116, opened 2017-09-14
Added AlbumArtist Extension plug-in

rdswift:
Modified Picard 1.4.2 version of the plug-in to work with Picard 2.0

### DIFF
--- a/plugins/albumartistextension/LICENSE
+++ b/plugins/albumartistextension/LICENSE
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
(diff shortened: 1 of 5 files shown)


rdswift:
Will update and resubmit.

plugins/albumartistextension/LICENSE
```diff
@@ -0,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
```
Sophist-UK, line 1: Do we really need both LICENSE and LICENSE.md?

rdswift: I cancelled this pull request and submitted a new one with just the plug-in python file.  The other files were only supposed to show up locally, but I PR'd the wrong branch.

PR #117, opened 2017-09-14
Added AlbumArtist Extension plugin

rdswift:
Updated Picard 1.4.2 version plugin to work with Picard 2.0

### DIFF
--- a/plugins/albumartistextension/albumartistextension.py
+++ b/plugins/albumartistextension/albumartistextension.py
@@ -1,0 +1,105 @@
+PLUGIN_NAME = 'AlbumArtist Extension'
+PLUGIN_AUTHOR = 'Bob Swift (rdswift)'
+PLUGIN_DESCRIPTION = '''
+This plugin provides standardized, credited and sorted artist information
+for the album artist.  This is useful when your tagging or renaming scripts
+require both the standardized artist name and the credited artist name, or
+more detailed information about the album artists.
+<br /><br />
+The information is provided in the following variables:
+<ul>
+<li>_aaeStdAlbumArtists = The standardized version of the album artists.
+<li>_aaeCredAlbumArtists = The credited version of the album artists.
+<li>_aaeSortAlbumArtists = The sorted version of the album artists.
+<li>_aaeStdPrimaryAlbumArtist = The standardized version of the first
+    (primary) album artist.
+<li>_aaeCredPrimaryAlbumArtist = The credited version of the first (primary)
+    album artist.
+<li>_aaeSortPrimaryAlbumArtist = The sorted version of the first (primary)
+    album artist.
+<li>_aaeAlbumArtistCount = The number of artists comprising the album artist.
+</ul>
+PLEASE NOTE: Tagger scripts are required to make use of these hidden
+variables.
+'''
+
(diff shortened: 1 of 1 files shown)


rdswift:
Sorry for the mess on the previous pull request.  This time I started with a fresh fork so it should be a lot cleaner.

plugins/albumartistextension/albumartistextension.py
```diff
@@ -0,0 +1,105 @@
...
+            sortArtist += tempSortName + tempPhrase
+            if aCount < 1:
+                album_metadata["~aaeStdPrimaryAlbumArtist"] = tempStdName
+                album_metadata["~aaeCredPrimaryAlbumArtist"] = tempCredName
+                album_metadata["~aaeSortPrimaryAlbumArtist"] = tempSortName
+            aCount += 1
+    else:
+        log.error("%s: %r: Missing 'artist-credit' in release metadata: %s",
+                PLUGIN_NAME, albumid, release_metadata)
+    if len(stdArtist) > 0:
```
zas: You can simply write : `if stdArtist:` to test for non-emptyness of this string.
Same goes for other tests below.

rdswift: Good point.  Looks like my perl background showing a little.  Updated as you suggested.

plugins/albumartistextension/albumartistextension.py
```diff
@@ -0,0 +1,104 @@
...
+from picard import config, log
+from picard.metadata import register_album_metadata_processor
+from picard.plugin import PluginPriority
+
+def add_artist_std_name(album, album_metadata, release_metadata):
+    albumid = release_metadata['id']
+    # Test for valid metadata node for the release
+    if 'artist-credit' in release_metadata:
+        # Initialize variables to default values
+        credArtist = ""
```
sambhav: Just nit picking but I would really like it if you could make all these variables snake_case instead of camelCase

sambhav (changes requested):

plugins/albumartistextension/albumartistextension.py
rdswift: Will do.  I agree it's not consistent as it is now, and looks ugly.

rdswift: Did I go too far in snake_casing the function name for the error logging?  Is there a set of code standards (or preferences) listed someplace?  That way I'll know what to try and follow in the future.  Thanks.

sambhav: Yes, you can find it here https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md

sambhav: There's also https://github.com/metabrainz/guidelines

sambhav (approved):

plugins/albumartistextension/albumartistextension.py
rdswift: Thanks for the information.  In reviewing it, it seems that I have over commented my code.  I'll pay closer attention to that in the future.  Thanks again.

zas (approved):

PR #118, opened 2017-09-21
Add Fix Track Numbers plugin

esotericpig:
I use a lot of Language-learning CDs, and the filenames are always something like "Unit 1 - Lesson 1." This caused Picard to set all of the track numbers using the Unit number, so I wrote this plugin to fix the issue. I also added a secondary sequential order fix. I use Picard to tag all of my audio files, so wanted to fix it using Picard. If it doesn't go into 2.0, no worries, just offering it up.


esotericpig:
Merge is called patch-1, oops. Will re-open with new name.

PR #125, opened 2017-11-21
Moodbar in 2.0

madmouser1:
Changed for Qt5 and python3 
tested on Ubuntu 17.10 with new Moodbar utility https://github.com/madmouser1/moodbar (Using Gstreamer 1.x)

### DIFF
--- a/plugins/moodbars/__init__.py
+++ b/plugins/moodbars/__init__.py
@@ -3,7 +3,7 @@
 # Changelog:
 #   [2015-09-24] Initial version with support for Ogg Vorbis, FLAC, WAV and MP3, tested MP3 and FLAC
 #   [2017-11-21] Amended to Python3 & Qt5
-#   [2017-11-21] removed unicode, 
+#   [2017-11-21] removed unicode, replaced str with string_ and untrusted input on check_call addressed 
 
 PLUGIN_NAME = "Moodbars"
 PLUGIN_AUTHOR = "Len Joubert, Sambhav Kothari"
@@ -71,7 +71,7 @@ def generate_moodbar_for_files(files, format, tagger):
         command_to_execute = command_to_execute + file_list_music
         tagger.log.debug('My debug command to execute >>>  %s' %
                          (command_to_execute))
-        check_call(command_to_execute)
+        check_call(command_to_execute, shell=False)
         # check_call([command] + options + file_list_mood + file_list)
     else:
         raise Exception('Moodbar: Unsupported format %s' % (format))
--- a/plugins/moodbars/__init__.py
+++ b/plugins/moodbars/__init__.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 
 # Changelog:
-#   [2015-09-24] Initial version with support for Ogg Vorbis, FLAC, WAV and MP3, tested MP3 and FLAC
-#   [2017-11-21] Amended to Python3 & Qt5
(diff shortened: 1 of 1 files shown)


plugins/moodbars/__init__.py
```diff
@@ -128,13 +145,13 @@ def load(self):
...
+        self.config.setting["moodbar_vorbis_command"] = str(
             self.ui.vorbis_command.text())
-        self.config.setting["moodbar_mp3_command"] = string_(
+        self.config.setting["moodbar_mp3_command"] = str(
             self.ui.mp3_command.text())
-        self.config.setting["moodbar_flac_command"] = string_(
+        self.config.setting["moodbar_flac_command"] = str(
             self.ui.flac_command.text())
-        self.config.setting["moodbar_wav_command"] = string_(
+        self.config.setting["moodbar_wav_command"] = str(
```
sambhav: Please use the string_ type to type convert to string. Its a picard internal type conversion system to handle a bunch of things.

plugins/moodbars/__init__.py
```diff
@@ -2,24 +2,25 @@
...
 #    'win32', 'cygwyn', 'darwin', 'os2', 'os2emx', 'riscos', 'atheos']
 
 import os.path
-from functools import partial
 from collections import defaultdict
 from subprocess import check_call
 from picard.album import Album, NatAlbum
 from picard.track import Track
 from picard.file import File
+from functools import partial
```
sambhav: Any reason why this was shifted from L18 to L23? I would prefer this to be with the python standard library imports.

plugins/moodbars/__init__.py
```diff
@@ -2,24 +2,25 @@
 
 # Changelog:
 #   [2015-09-24] Initial version with support for Ogg Vorbis, FLAC, WAV and MP3, tested MP3 and FLAC
+#   [2017-11-21] Amended to Python3 & Qt5
 
-PLUGIN_NAME = "Moodbars"
-PLUGIN_AUTHOR = "Len Joubert, Sambhav Kothari"
+PLUGIN_NAME = u"Moodbars"
+PLUGIN_AUTHOR = u"Len Joubert, Sambhav Kothari"
```
sambhav: Unicode isn't present in py3. We can drop the prefix here.

sambhav (changes requested):

plugins/moodbars/__init__.py
madmouser1: ok

plugins/moodbars/__init__.py
madmouser1: will do, only as I removed partial from picard.util and added straight above

madmouser1:
many thanks @samj1912 I have amended the pull request and tested.

madmouser1:
@samj1912 I have made the requested changes if you would like to review, many thanks

plugins/moodbars/__init__.py
```diff
@@ -2,15 +2,17 @@
 
 # Changelog:
 #   [2015-09-24] Initial version with support for Ogg Vorbis, FLAC, WAV and MP3, tested MP3 and FLAC
+#   [2017-11-21] Amended to Python3 & Qt5
+#   [2017-11-21] removed unicode, replaced str with string_ and untrusted input on check_call addressed
```
sambhav: As pointed out by codacy, can you please remove the trailing whitespace from here?

sambhav (approved):

PR #139, opened 2018-01-15
v0.9 of Classical Extras

MetaTunes:
Major additions to functionality - see full version history in readme for details

Issue #134, opened 2017-12-17
Classical Extras plugin makes Picard hang after a couple of albums.


hiccup2:
As briefly discussed on the forum, after processing a few albums with this plugin active, Picard will hang.
Restarting Picard will fix it briefly, but will freeze again after a few albums.
[Picardini.zip](https://github.com/metabrainz/picard-plugins/files/1565770/Picardini.zip)

W10x64
No scripts or other plugins running.

Sophist-UK:
I have just had a look at the code and it could definitely benefit from some performance improvements.

Obvious things like making constant arrays and regular expressions class constants rather than redefining them inside a method would be beneficial.

But it also looks like it does an awful lot of loop processing which (knowing Python) can almost certainly be rewritten significantly more efficiently.

All that said, I haven't spotted any memory leaks which might cause it to hang, though I note that it does create class properties which persist and which could grow over time.

MetaTunes:
I'd welcome any code improvements from a Python expert! I am well aware that it has got a bit spaghetti-like.
Re the reported issue, I have run my test database against the hiccup2's picard.ini and it causes a persistent failure to load the same 8 albums (out of 122 in the current test set). These albums load fine using my picard.ini. When I have had a problem like this before it was because of cross-contamination of class variables. I'll do a bit more digging.
Re the plans for the code. I'd like to stabilise 0.8.7 before inviting any attempts to re-engineer it. I have some minor un-released improvements that also need to be incorporated. This should form the basis for a v0.9 to be a pull request. The final (planned) version 1.0 is intended to incorporate additional functionality for instruments/ genres/ periods etc. After that it should be pretty complete and it will be time to look at migrating to Picard 2.0.

MetaTunes:
OK, I think I know what the problem is. The "Works and Parts" option "Use cache" has been turned off. When there are two albums loaded that have a work in common, the second one will try and look up the works again. Not only is this inefficient, it also confuses the queue manager into thinking it hasn't got all the works, so it waits.... So the simple work-round solution is to enable "Use cache" (the default) and it should all run OK. I'll run my big test to double check.
The question remains as to how to prevent this. I'll look at a code fix, but I really don't want people turning cache off and forgetting as it does make things run slower. The only reason for turning off cache is to re-fetch data after a MB or structure change. So as well as a code fix, I may make it so that cache will automatically be turned on for each new session, regardless of how it was left at the end of the previous session.

MetaTunes:
[...]

Issue #133, opened 2017-12-11
Classical Extras: Error if release has a partial recording


aereaux:
I am using the 0.8.6 release of classical_extras from here: https://github.com/MetaTunes/picard-plugins/ , but this was the only place I could find to report bugs.  I think this bug also happens with the version of classical_extras in this repo.

Anyways, when classical_extras is enabled, on some of the albums in my music collection, picard gets stuck on "[loading album information]", and it outputs this to stderr:
```Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/picard/webservice.py", line 315, in _process_reply
    self._handle_reply(reply, request, handler, xml, refresh)
  File "/usr/lib/python2.7/site-packages/picard/webservice.py", line 304, in _handle_reply
    handler(document, reply, error)
  File "/home/aidane/.config/MusicBrainz/Picard/plugins/classical_extras.zip/classical_extras/__init__.py", line 3206, in work_process
  File "/home/aidane/.config/MusicBrainz/Picard/plugins/classical_extras.zip/classical_extras/__init__.py", line 3485, in process_album
  File "/home/aidane/.config/MusicBrainz/Picard/plugins/classical_extras.zip/classical_extras/__init__.py", line 4315, in extend_metadata
  File "/home/aidane/.config/MusicBrainz/Picard/plugins/classical_extras.zip/classical_extras/__init__.py", line 5074, in diff_pair
IndexError: list index out of range
```
and: 
```Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/picard/webservice.py", line 315, in _process_reply
    self._handle_reply(reply, request, handler, xml, refresh)
  File "/usr/lib/python2.7/site-packages/picard/webservice.py", line 304, in _handle_reply
    handler(document, reply, error)
  File "/home/aidane/.config/MusicBrainz/Picard/plugins/classical_extras.zip/classical_extras/__init__.py", line 3206, in work_process
  File "/home/aidane/.config/MusicBrainz/Picard/plugins/classical_extras.zip/classical_extras/__init__.py", line 3485, in process_album
  File "/home/aidane/.config/MusicBrainz/Picard/plugins/classical_extras.zip/classical_extras/__init__.py", line 4116, in extend_metadata
  File "<string>", line 0
    
    ^
SyntaxError: unexpected EOF while parsing
```

I think this happens because the release include a partial recording.  If any other information is needed to debug the problem, or if this is not the right place to report this bug, let me know.

[...]

PR #140, opened 2018-01-16
Classical Extras v0.9 (2nd attempt)

MetaTunes:
See readme for full version history


MetaTunes:
Resubmitted PR after resolving conflicts in #139

MetaTunes:
This PR is now ready for review. I don't know why the Travis CI build could not complete - it was OK before I made a very minor change which I don't think could have affected it.
EDIT - spotted a problem - some code altered in error - although I don't know why it would cause the Travis problem. Will push and try again.

MetaTunes:
Travis build cancelled again - I've no idea why. Help please!

Sophist-UK:
Travis log is empty - it either didn't run at all, or they failed to save the log - so either way most likely to be a problem at Travis' end.

MetaTunes:
A bit of palaver getting through the checks by making 'non-changes' but passed, so I won't attempt any more tweaking!
Just needs review now.

MetaTunes:
Not sure if anyone is planning to review this. @zas ?
I appreciate it's quite a big plugin, but it has had a fair bit of testing (see thread at https://community.metabrainz.org/t/classical-extras-plugin/300217 and issues #133, #134, &#135).

plugins/classical_extras/__init__.py
```diff
@@ -2,39 +2,42 @@
 
 PLUGIN_NAME = u'Classical Extras'
 PLUGIN_AUTHOR = u'Mark Evens'
-PLUGIN_DESCRIPTION = u'''This plugin contains 3 classes:
+PLUGIN_DESCRIPTION = u"""This plugin contains 3 classes:
```
Sophist-UK: The PLUGIN_DESCRIPTION is intended to be used to provide a short user description of what this plugin can do for them rather than an explanation of how it works internally.

Sophist-UK: That is not to say that the current description is not useful to anyone reading the code, though I am finding it very difficult to understand what this code is trying to do given the current level of commenting.

plugins/classical_extras/Readme.md
```diff
@@ -1,74 +1,147 @@
...
-3.  Keep a backup of your picard.ini file (AppData->Roaming->MusicBrainz) in case you erase your settings or Picard crashes and loses them for you.
+1.  The plugin **will not work fully unless** you have enabled "Use release relationships" and "Use track relationships" in Picard->Options->Metadata. However, it may be that the MusicBrainz database has conflicting data between track and release relationships, in which case fix the incorrect data using "Edit relationships" in MusicBrainz.
```
Sophist-UK: If this is the case then IMO the plugin should either:

a. Turn these options on, warning the user that it is about to do so;
b. Or give a message saying they need to be turned on before quitting.

plugins/classical_extras/Readme.md
```diff
@@ -1,74 +1,147 @@
...
+2.  It is recommended only to use the plugin on one or a few release(s) at a time, particularly for initial tagging and if the "Works and parts" function is being used. The plugin is not designed to do "bulk tagging" of untagged files - it may be better to use a tool such as SongKong for that and then use the plugin to enhance the results as required. However, once you have tagged files (either in Picard or, say, SongKong) such that they all at least have MusicBrainz IDs, you should be able to re-tag multiple releases by dragging the containing folder into Picard; this is useful to pick up changed MusicBrainz data or if you change the Classical Extras version or options (but bear in mind that the "Works and parts" function will still take at least 1 second per track and **make sure you have "debug" and "info" logging turned OFF** - in the "Advanced" tab). If you do want to try uising Picard / Classical Extras to tag a large number of files, then **make sure you also turn off "warning" logging**.
```
Sophist-UK: Why can't Picard be used for bulk tagging without this plugin. Why are you recommending SongKong instead?

plugins/classical_extras/Readme.md
```diff
@@ -1,74 +1,147 @@
...
+  This section deals primarily with the application of aliases and credited_as names to replace the MusicBrainz standard names. The first box allows you to choose whether to replace MusicBrainz standard  names by aliases - either for all work-artists/performers or only work-artists (writers, composers, arrangers, lyricists etc.). The second box sets the usage of "as-credited" names: the first part of this lists all the places where as-credited names can occur (really!) and the second part allows you to apply these to performing artists and/or work-artists. Please note that, in the current version of this plugin, only aliases and credited_as names which are in the "release XML node" are available (i.e. roughly those relating to the metadata shown in the release overeview page in MusicBrainz). So, for example, if a recording is an arrangement of another work and that other work (but not the arrangement) has a composer linked to it, then the composer's alias will not be available (nor is the composer shown on the MB release overview page). In some cases (if appropriate) this can be remedied by adding the relevant link to the lowest-level work.
```
Sophist-UK: This line is over 1000 characters long. MD files will concatenate shorter lines, so this is unnecessary.

Sophist-UK: Any chance that I can have a translation of this paragraph into English so I can understand what it is describing? Or perhaps some examples are needed.

plugins/classical_extras/Readme.md
```diff
@@ -1,74 +1,147 @@
...
+
+  Classical Extras puts the recording artists into 'hidden variables' (as a minimum) using the chosen naming convention.
+  There is also option to allow you to replace the track artist by the recording artist (or to merge them). The chosen action will be applied to the 'artist', 'artists', 'artistsort' and 'artists_sort' tags. Note that 'artist' is a single-valued string whereas 'artists' is a list and may be multi-valued. Lists are properly merged, but because the 'artist' string may have different join-phrases etc, a merged tag may have the recording artist(s) in brackets after the track artist(s). Obviously, for classical music, if you use "merge" then the artist tag will have both the composer and the recording artists: this may be desirable for simple players (with no composer recognition) but otherwise may look odd.
+
+   Note that, if the original track artist is required in tag mapping (i.e. as it was before replacement/merge with recording artist), it is available through the hidden variable _cea_MB_artists.
+
```
Sophist-UK: We seem to have gone all italic from this point forward. Probably an unclosed MD markup.

plugins/classical_extras/Readme.md
```diff
@@ -1,74 +1,147 @@
...
+   Note also that, if @loujin's browser script has been used to fill the recording artist data, this will be the same as the performing artists in the Recording-Artist relationship - i.e. it may be a lengthy list rather than the principal artist for the track.
+
+4. "Other artist options":
+
+  "Modify host tags and include annotations" (Previously called "Include arrangers from all work levels"). This will gather together, for example, any arranger-type information from the recording, work or parent works and place it in the "arranger" tag ('host' tag), with the annotation (see below) in brackets. All arranger types will also be put in a hidden variable, e.g. _cwp_orchestrators. The table below shows the artist types, host tag and hidden variable for each artist type.
+
+   | Artist type | Host tag | Hidden variable |
```
Sophist-UK: Does this need to go into an MD code block to preserve formatting?

plugins/classical_extras/__init__.py
```diff
@@ -44,24 +47,1444 @@
...
+    # set_options[track]['standardize_artists'] = config.setting['standardize_artists']
+
+    options = set_options[track]
+    tm = track.metadata
+    orig_metadata = None
+    # Only look up files if needed
+    file_options = {}
+    music_file_found = None
+    found_file = False
+    for music_file in album.tagger.files:
```
Sophist-UK, line 166: You do say that this plugin isn't scalable, but this loop is almost certainly part of the reason why, not just the no. of web services calls it makes.

You need to try to process information once as the track is loaded, and not have an n^2 relationship that takes longer and longer to add a file the more files you already have loaded.

plugins/classical_extras/Readme.md
```diff
@@ -282,10 +427,62 @@ Leave the "title" tag unchanged or make it a combination of work and movement.
...
+
+A large variety of releases were used to test this plugin, but there may still be bugs, so further testing is welcome. The following release was particularly complex and useful for testing: https://musicbrainz.org/release/ec519fde-94ee-4812-9717-659d91be11d4. Also this release was a bit tricky - a large box set with some works appearing as originals and in arrangements: https://musicbrainz.org/release/5288f266-bab8-45bd-83e4-555730f02fa0.
+
```
Sophist-UK, line 443: Despite having read this, I am still in the dark about what this plugin really does. I have an idea that it gets more metadata but I have no idea what, or where it is put or how much? Some examples would be very beneficial.

Also I think a warning is needed in the Plugin Description that this plugin is for the dedicated classical music enthusiast and not for the casual or faint-hearted users.

plugins/classical_extras/__init__.py
```diff
@@ -44,24 +47,1444 @@
...
+    :param self:
+    :param album:
+    :param options:
+    :param releaseXmlNode: all the metadata for the release
+    :return: Data is returned via self.artist_aliases and self.artist_credits[album]
+
+    Note regarding aliases and credited-as names:
+    In a MB release, an artist can appear in one of seven contexts. Each of these is accessible in releaseXmlNode
+    and the track and recording contexts are also accessible in trackXmlNode.
+    The seven contexts are:
```
Sophist-UK, line 958: I make this 11 contexts if you count credited-as and alias as separate contexts.

Sophist-UK (commented):
I have not tried to review the code itself as having read the MD file and the code I have no idea of what it is trying to do or why.

The screen shots in the MD file do give a sense of the breadth of specialised functionality this provides, but it feels like a sink or swim experience if you try to use this.

My overall impression of code quality is, however, that it is good, and given the specialist nature of this plugin, I am less concerned than normal about verifying the functionality.

I am also impressed by the amount of functionality this apparently has - probably more options than Picard native.

My sole concern is that we do not get swamped with support requests by users who have jumped in and find that this is too complicated for them.

I therefore recommend two things:

1. That the MD file is used as the basis for a comprehensive set of Github Wiki pages for this plugin which break it down into smaller more digestible chunks and which provide a more gentle top-down step-by-step explanation of what the plugin is for and what it can do.

2. That this includes many more examples of how this can help users with their classical music.

MetaTunes:
Thanks for all the comments, which I will consider more deeply shortly. In the meantime, perhaps some context is achieved by looking at my website: http://music.highmossergate.co.uk/enhanced.php

Sophist-UK:
@MetaTunes Mark - the overall gist of my comments are about a lack of documentation here - asking me to go to another site to find out is exactly the problem that needs to be solved.

MetaTunes:
Some might say that there is too much documentation. Certainly a lot more than can be said of some other software I won't mention. I think the issue is that the plugin is quite complex and the documentation has evolved as it has grown. BTW I have also added extensive 'tooltips' and 'what's this' in the UI. It really is necessary to use it for a bit in order to understand it (just like Picard). I will look at the comments in a bit more depth when I get a mo.
From a practical point of view, I would like to draw a line at this point (hence the PR), get some more feedback and reflect before making any more changes. After all, I called it v0.9 deliberately. The final version for Picard 1.4 is intended to be v1.0 which will include a bit more functionality but will also address usability issues that arise.

plugins/classical_extras/__init__.py
```diff
@@ -44,24 +47,1444 @@
...
+        ('release', 'release_name', False),
+        ('work_type', 'genre', False),
+        ('ensemble_names', 'band', False),
+        ('composers', 'artist', True),
+        ('MB_artists', 'composer', True),
+        ('arranger', 'composer', True)
+    ]
+    tag_options = []
+    for i in range(0, 16):
+        if i < len(default_list):
```
zas, line 597: Isn't this equivalent to following code ?

```python
for i in range(0, len(default_list):
```

plugins/classical_extras/__init__.py
```diff
@@ -44,24 +47,1444 @@
...
+            default_source = default_list[i][0]
+            default_tag = default_list[i][1]
+            default_cond = default_list[i][2]
```
zas:
```python
default_source, default_tag, default_cond = default_list[i]
```

plugins/classical_extras/__init__.py
```diff
@@ -44,24 +47,1444 @@
...
+        tag_options.append({'option': 'cea_source_' + unicode(i + 1),
```
zas: You should store the expression `unicode(i +1)` in a variable, and replace all those 6 occurences

plugins/classical_extras/__init__.py
```diff
@@ -44,24 +47,1444 @@
...
+        lang,
+        a_list,
+        name_style,
+        name_tag,
+        sort_tag,
+        names_tag,
+        names_sort_tag):
+    # Get  artist and apply style
+    for acs in a_list:
+        for ncs in acs:
```
zas, line 1227: `ncs` and `acs`, hard to know what this means, you should use more explicit variable names

plugins/classical_extras/__init__.py
```diff
@@ -44,24 +47,1444 @@
...
+        cwp_tag = tag_strings[writer_type][1]
+        cwp_sort_tag = tag_strings[writer_type][3]
+        cwp_names_tag = cwp_tag[:-1] + '_names'
+        if writer_type in sub_strings:
+            if sub_strings[writer_type]:
+                tag += sub_strings[writer_type]
+        if tag:
+            if '~ce_tag_cleared_' + \
+                    tag not in tm or not tm['~ce_tag_cleared_' + tag] == "Y":
+                if tag in tm:
```
zas, line 1424: It would be better to use a more explicit variable name than `tm`

plugins/classical_extras/__init__.py
```diff
@@ -150,54 +1573,137 @@ def get_roman(string):
...
+
+
+def _reverse_sortname(sortname):
+    """
+    Reverse sortnames.
+    Code is from picard/util/__init__.py
+    """
+
+    chunks = [a.strip() for a in sortname.split(",")]
+    if len(chunks) == 2:
```
zas: Cache `len(chunks)`

Sophist-UK:
If you change the PLUGIN_DESCRIPTION to be a more friendly description with warnings to the innocent that would be sufficient IMO for this to be merged so that you can get feedback.

zas (commented):
This is almost impossible to review:
- code should be splitted in smaller methods (or smaller files)
- comments in the code are lacking, very hard to know what is doing what without digging
- short variable names should only be used for loop indexes
- the patch is huge, the commit history is almost useless because of generic comments

Overall the code is good, and i'm confident it works well enough, but imho the code lacks of clarity, making hard for others to provide improvements and/or fixes.

Also, think about the fact we are moving to python3/qt5 in Picard 2 (with plugins v2), some of the code here will need changes, better to anticipate and make code py2/py3 and qt4/qt5 compatible when it is simple enough. It will ease the transition to plugins v2.

MetaTunes:
>If you change the PLUGIN_DESCRIPTION to be a more friendly description

No problems with that - I quite agree!

MetaTunes:
I hadn't realised at the time I first did it that the description was for user display rather than developer use.

MetaTunes:
For reviewers / approvers: I will add another commit later this evening (which I hope Travis doesn't hiccup over) dealing with documentation issues. The functionality and code will be unchanged.

MetaTunes:
>If this is the case then IMO the plugin should either:
>a. Turn these options on, warning the user that it is about to do so;
>b. Or give a message saying they need to be turned on before quitting.

That's a good idea, but there is a slight problem with it, as is implied by the (amended) text: sometimes MB has conflicting metadata between the release and track relationships. Ideally this should be fixed by editing it, but not everyone feels comfortable with that, so I left it for the user to decide which exactly to use. I also deliberately decided not to mess with the 'native' Picard settings. On reflection, it might be best to set the defaults as suggested and to document that (and how to handle exceptions). Will consider for 1.0.

MetaTunes:
>Why can't Picard be used for bulk tagging without this plugin. Why are you recommending SongKong instead?

Native Picard doesn't tag the work hierarchy, whereas SongKong does. Classical Extras will read the SongKong tags. Another approach would be to get all the MBIDs with native Picard first and then use the plugin to refine files which have the MBID tags. With the latest version this latter approach is now possible without needing two runs (see https://community.metabrainz.org/t/classical-extras-plugin/300217/129) I will test out alternative strategies to see what works best.

MetaTunes:
>This line is over 1000 characters long. MD files will concatenate shorter lines, so this is unnecessary.
>Any chance that I can have a translation of this paragraph into English so I can understand what it is >describing? Or perhaps some examples are needed.

Unfortunately MusicBrainz is pretty complex - I have attempted to describe how MB works in the box that follows the quoted text (7 different ways an artist can be related to a track!) and even now am not 100% sure it is right. If anyone can point me to a clear documentation of this in MB then I would be most grateful. In addition, the way Picard handles aliases etc. in the context of the "translate" and "standardize" options is inconsistent with the Picard documentation.
Agreed that some examples might help. I'll look into this, but the complexity would require many examples to illustrate the different possibilities.

MetaTunes:
>We seem to have gone all italic from this point forward. Probably an unclosed MD markup.

Caused by the underscore. But it displays OK in the browser.

plugins/classical_extras/Readme.md
MetaTunes: That's a good idea, but there is a slight problem with it, as is implied by the (amended) text: sometimes MB has conflicting metadata between the release and track relationships. Ideally this should be fixed by editing it, but not everyone feels comfortable with that, so I left it for the user to decide which exactly to use. I also deliberately decided not to mess with the 'native' Picard settings. On reflection, it might be best to set the defaults as suggested and to document that (and how to handle exceptions). Will consider for 1.0.

plugins/classical_extras/Readme.md
MetaTunes: Native Picard doesn't tag the work hierarchy, whereas SongKong does. Classical Extras will read the SongKong tags. Another approach would be to get all the MBIDs with native Picard first and then use the plugin to refine files which have the MBID tags. With the latest version this latter approach is now possible without needing two runs (see https://community.metabrainz.org/t/classical-extras-plugin/300217/129) I will test out alternative strategies to see what works best.

plugins/classical_extras/Readme.md
MetaTunes: Unfortunately MusicBrainz is pretty complex - I have attempted to describe how MB works in the box that follows the quoted text (7 different ways an artist can be related to a track!) and even now am not 100% sure it is right. If anyone can point me to a clear documentation of this in MB then I would be most grateful. In addition, the way Picard handles aliases etc. in the context of the "translate" and "standardize" options is inconsistent with the Picard documentation.
Agreed that some examples might help. I'll look into this, but the complexity would require many examples to illustrate the different possibilities.
BTW I've split the para in the next commit.

plugins/classical_extras/Readme.md
MetaTunes: Caused by the underscore. But it displays OK in the browser.

Sophist-UK:
> Native Picard doesn't tag the work hierarchy, whereas SongKong does. Classical Extras will read the SongKong tags. Another approach would be to get all the MBIDs with native Picard first and then use the plugin to refine files which have the MBID tags. With the latest version this latter approach is now possible without needing two runs (see https://community.metabrainz.org/t/classical-extras-plugin/300217/129) I will test out alternative strategies to see what works best.

No - your plugin should do the SongKong-style tagging - it makes no sense to expect the user to keep switching tagging software.

MetaTunes:
Think I may have erased a comment as I moved all my comments into reply boxes (brain fade - a bit late). But I got the point about avoiding multiple tagging software if possible. I need to be comfortable that Picard will handle it now I've got a possible solution - will resolve in 1.0

plugins/classical_extras/Readme.md
MetaTunes: In the next commit

plugins/classical_extras/Readme.md
MetaTunes: I have done one example screen showing a result. I agree more could be done. 
> this plugin is for the dedicated classical music enthusiast and not for the casual or faint-hearted users

Classical music enthusiasts are neither casual nor faint-hearted ;)

plugins/classical_extras/__init__.py
MetaTunes: PLUGIN_DESCRIPTION to be modified in the next commit

plugins/classical_extras/__init__.py
MetaTunes: If you can find a solution for this I would be most grateful, but at present I can find no way of directly accessing a filename from within a "track_metadata_processor" - so the only solution was to get the tags of all the files and find a matching track id.
I was hoping to be able to use album.tagger.get_files_from_objects, but that only seems to work at the album level. I guess I may be able to speed it up by just getting the album filenames. 
I agree this could use some improvement. Nevertheless the main time issue is the 1 second constraint on web xml lookups. Even with 2000 tracks, the impact of this loop is small in relation to that inbuilt delay.
BTW, this loop does not increase the number of web service calls as the data is cached - the number of calls is dependent on the total number of works.

MetaTunes:
Thanks for the comments @zas 
As per my earlier responses, I agree that some tidying up could be done (and will be done). My plan was to release v0.9, as it works well enough, to get some more usage and feedback. This would be incorporated, along with a lot of code tidying (comments welcome - I'm a fairly newbie Python programmer and learnt on COBOL!) into v1.0 which would then form the basis of moving to Picard 2.0.
Given that context, I hope, with the documentation enhancements suggested by @Sophist-UK , that it is OK to go.

MetaTunes:
No idea why I'm getting the Codacy error
Unused variable '__class__'
3272 class PartLevels:
This happened before then disappeared for no reason. Just making a non-change for now.

plugins/classical_extras/__init__.py
Sophist-UK: No - the loop will cause Picard to appear to hang and create a poor user experience.

And I think you may be right about not being able to get to the file from the track metadata - the sequence of events AFAICR is:

* Get or guess the WS Release release MBID from the first file or the cluster.
* Create the album node with this MBID and move the first file to the Unmatched file node.
* Get the WS Release data for the MBID from MB
* Move the rest of the files with that MBID or in the same cluster to the Unmatched files node of the Album
* Process the WS data including calling album and track metadata plugins
* Associate files with tracks (because this uses the metadata to make the track MBID association or guess it.

Unless the WS Get request processes extremely fast, by the time the album / track metadata plugins are run, all the files should be in the Unmatched files node of the Album node and so you should (somehow) be able to access them.

Sophist-UK:
> No idea why I'm getting the Codacy error
> Unused variable 'class'
> 3272 class PartLevels:
> This happened before then disappeared for no reason. Just making a non-change for now.

If it doesn't disappear, it might be an unmatched curly bracket.

MetaTunes:
>If it doesn't disappear, it might be an unmatched curly bracket.

Surely that would show up as a syntax error? As I said, it has happened before and then disappeared after making a "non-change", so it seems to be a Codacy problem. I'll try something else...

MetaTunes:
There was one non-matching curly but only in a comment. I fixed it anyway, but the problem persists. Since PartLevels is the main class and it runs fine, the Codacy error is incorrect (or it wouldn't run!). I'll poke it once more - if that fails I will need some (even more) expert help.

MetaTunes:
Is this the issue: https://github.com/metabrainz/critiquebrainz/pull/172?
If so, it says it was fixed a few hours ago, but I'm still getting these errors.
I guess that fix was just in critiquebrainz - not sure if the same issue is affecting this PR

MetaTunes:
Can anyone (@zas , @samj1912 ?) help with this?  Codacy is repeatedly failing because it says that the classes are unused variables, which is plainly wrong and seems to be a pylint problem. (Note it also doesn't like the the eval(), but that is not fatal and Picard 1.4 doesn't seem to have ast.)
Thank.

plugins/classical_extras/__init__.py
MetaTunes: No. It does need to run through the whole 16, but there was a theoretical issue regarding undeclared variables, which shouln't occur in practice but which I have nonetheless fixed.

plugins/classical_extras/__init__.py
MetaTunes: Thanks. Still learning Python - far too clever.

plugins/classical_extras/__init__.py
MetaTunes: Agreed. I ran out of ideas. These are only used in a very small context.

plugins/classical_extras/__init__.py
MetaTunes: I followed the practice of other plugins by using tm as shorthand for track.metadata. I thought it was accepted terminology.

plugins/classical_extras/__init__.py
MetaTunes: Will do

plugins/classical_extras/Readme.md
MetaTunes: Latest push sets the options on (no warning, but it is still emphasised in the readme - frankly there's not much point having them off if you want to use the plugin).

MetaTunes:
Still failing Codacy. I'm afraid I have no idea how to resolve this and suspect it is a pylint problem rather than anything wrong with my code. I'd really appreciate help from the experts on this (or at least some comments?). I will have no access to my development machine for the next 2 weeks so can't make any more changes for a while.

MetaTunes:
Can a reviewer @zas or @samj1912 help please? I keep getting what appear to be false pylint "unused variable __class__" errors from Codacy. Other than that, this PR should be ready for final review and release. I've fixed most of the previous review issues raised.

mineo:
If it's obviously not a bug in your code, just ignore the warnings/errors.

MetaTunes:
It is obviously not a bug - if it were a real error, the code would not run at all. I had assumed that the reviewer(s) would not approve it with a Codacy fail, but maybe that assumption is wrong. Whatever, it does need a reviewer to approve it or to tell me why not.

sambhav (approved):

zas:
@MetaTunes : ignore the Codacy error. Is the PR ready to merge ?

MetaTunes:
@zas : I've just run a test vs a 2,000 track test set and it seems to be error-free. So please merge this and I'll make a start on v1.0. - that is intended to be the final version before migrating to Picard 2.0.

MetaTunes:
Have since found a slight bug :(
New PR issued for the fix - v0.9.1

PR #144, opened 2018-02-11
Classical Extras v0.9.1

MetaTunes:
Bug fix. For files without a MB track id, the files were not getting matched and it was necessary to run Picard without the plugin and then refresh with the plugin. v0.9.1 fixes this (but a refresh is still needed to carry out processing on pre-existing file tags).


zas (approved):

aereaux:
Will this update (and the previous v0.9 one) appear at https://picard.musicbrainz.org/plugins/ at some point?  I assume this is what it takes for it to be updatable from within picard.

MetaTunes:
It was merged a few days ago, but I'm not sure (@zas ?) how frequently the plugins download section is updated. Meanwhile, you can download the classical_extras.zip file at https://github.com/MetaTunes/picard-plugins/releases/tag/v0.9.1
Then in Picard, choose options->plugins->install and select the zip file.
(At least, that works for Windows - not sure for other OS)

MetaTunes:
@zas I see that https://picard.musicbrainz.org/plugins/ is still showing v0.9, not v0.9.1

PR #149, opened 2018-06-06
Fix papercdcase non-ASCII encoding

### LINKED ISSUE #148: Handling for Umlaute/vowels in 'papercdcase.py' from Philipp Wolfer
If you use the plugin 'Paper CD case' from Philipp Wolfer, submitted Umlaute/vowels doesn't get converted correctly in the advanced paper cd online form.

It would be great if the Umlaute/vowels would be sent as HTML entitites. 
**The special thing is this:**
Don't send the word "Träne" as `Tr&auml;ne` with `&auml;` as ä -> it will be truncated at the &-character.
Please send it as `Tr%26auml;ne` (with `%26` as `&` replacement) and it will work.

phw:
This fixes #148

Please note that papercdcase.com does not handle Unicode and uses the iso-8859-1 encoding. The patch works for properly encoding in the range of unicode code points that can be represented by a single byte, everything above will fail.

I have not yet managed to get this to work with CJK characters, even though papercdcase claims to support those. It probably uses some special encoding.

### DIFF
--- a/plugins/papercdcase/papercdcase.py
+++ b/plugins/papercdcase/papercdcase.py
@@ -20,7 +20,7 @@
 PLUGIN_NAME = 'Paper CD case'
 PLUGIN_AUTHOR = 'Philipp Wolfer, Sambhav Kothari'
 PLUGIN_DESCRIPTION = 'Create a paper CD case from an album or cluster using http://papercdcase.com'
-PLUGIN_VERSION = "1.0"
+PLUGIN_VERSION = "1.1"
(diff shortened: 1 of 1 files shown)


Sophist-UK:
PW: Are you planning a similar PR for Picard v1 version of the plugin?

phw:
@Sophist-UK #150

phw:
Looks like travis CI for the 2.0 branch is somewhat broken :(

phw:
Tested in https://github.com/metabrainz/picard-plugins/issues/148

PR #150, opened 2018-06-06
papercdcase: URL encode non-ASCII characters

phw:
Same as #149 but for the 1.0 branch.


phw:
Merging, see comments on  #149 and https://github.com/metabrainz/picard-plugins/issues/148

PR #155, opened 2018-08-15
standardise_performer: Fixed RuntimeError

phw:
In Python 3 it is no longer valid to modify the size of a dictionary that is iterated via .rawitems(), since .rawitems() returns an iterator instead of a list.

See https://community.metabrainz.org/t/standardise-performers-plugin-doesnt-work-with-python3/391982


Sophist-UK:
@phw Thanks for looking at this for me - I just don't have the time at present.

zas (approved):

PR #164, opened 2018-09-10
Last.fm Python and cache fixes

phw:
This updates the Last.fm plugin to fix two issues:

- The last changes introduced access to `re.Pattern`, which is invalid on Python < 3.7. Replace this with duck typing, which probably is also more pythonic.
- Changing the Last.fm setting still gave you cached results not using your changed settings. This was already the case for oder versions of this plugin. This PR clears the cache if `lastfm_min_tag_usage` or `lastfm_ignore_tags` get changed


zas (approved):

zas:
@phw : deployed

PR #168, opened 2018-10-06
Show 'guest, 'solo' and 'additional' on all related performer tags

rdswift:
In response to the discussion at [https://community.metabrainz.org/t/retrieving-entering-guest-and-additional-attributes-for-performers-playing-more-than-one-instrument/400201](https://community.metabrainz.org/t/retrieving-entering-guest-and-additional-attributes-for-performers-playing-more-than-one-instrument/400201), the plugin has been updated as suggested by @Sophist-UK.

### DIFF
--- a/plugins/standardise_performers/standardise_performers.py
+++ b/plugins/standardise_performers/standardise_performers.py
@@ -35,10 +35,7 @@ standardise_performers_split = re.compile(r", | and ").split
 
 
 def standardise_performers(album, metadata, *args):
-    for key, values in list(metadata.rawitems()):
-        if not key.startswith('performer:') \
-                and not key.startswith('~performersort:'):
-            continue
+    for key, values in list(filter(lambda filter_tuple: filter_tuple[0].startswith('performer:') or filter_tuple[0].startswith('~performersort:'), metadata.rawitems())):
         mainkey, subkey = key.split(':', 1)
         if not subkey:
             continue
--- a/plugins/standardise_performers/standardise_performers.py
+++ b/plugins/standardise_performers/standardise_performers.py
@@ -22,7 +22,7 @@ Performer [synthesizer]: Lol Creme
 Performer [tambourine]: Graham Gouldman
 </pre>
 '''
-PLUGIN_VERSION = '0.3'
+PLUGIN_VERSION = '0.4'
 PLUGIN_API_VERSIONS = ["0.15.0", "0.15.1", "0.16.0", "1.0.0", "1.1.0", "1.2.0", "1.3.0", "2.0"]
 PLUGIN_LICENSE = "GPL-2.0"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
@@ -49,8 +49,16 @@ def standardise_performers(album, metadata, *args):
                   PLUGIN_NAME,
                   subkey,
                   )
+        extra = ''
+        temp = ''
+        for part in instruments[0].split():
+            if part in ['guest', 'solo', 'additional']:
+                extra = "{0}{1} ".format(extra, part)
+            else:
+                temp = "{0} {1}".format(temp, part).strip()
+        instruments[0] = temp
         for instrument in instruments:
-            newkey = '%s:%s' % (mainkey, instrument)
+            newkey = '%s:%s%s' % (mainkey, extra, instrument)
             for value in values:
                 metadata.add_unique(newkey, value)
         del metadata[key]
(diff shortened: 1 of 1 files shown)


plugins/standardise_performers/standardise_performers.py
```diff
@@ -49,8 +49,16 @@ def standardise_performers(album, metadata, *args):
                   PLUGIN_NAME,
                   subkey,
                   )
+        extra = ''
+        temp = ''
+        for part in instruments[0].split():
+            if part in ['guest', 'solo', 'additional']:
```
phw: Inside Picard also the `minor` prefix is handled

plugins/standardise_performers/standardise_performers.py
```diff
@@ -49,8 +49,16 @@ def standardise_performers(album, metadata, *args):
                   PLUGIN_NAME,
                   subkey,
                   )
+        extra = ''
```
phw: I usually prefer using lists for this and join them later, e.g. have `extra = []`, then just do `extra.append(part)` and later call `' '.join(extra)`. Also I would probably call it `prefixes` instead of `extra`, I think that's clearer.

phw (changes requested):
Functionality wise looking good and for me worked as expected. I would just suggest some minor stylistic changes and addition of the `minor` prefix

Sophist-UK:
I agree with Philipp's suggestions, otherwise looks good from a desk read through.

plugins/standardise_performers/standardise_performers.py
```diff
@@ -49,8 +49,16 @@ def standardise_performers(album, metadata, *args):
...
+        temp = ''
+        for part in instruments[0].split():
```
Sophist-UK: If I were programming this I would do:
```
prefixes = []
words = instruments[0].split()
for word in words[:]:
    if not word in ['guest', 'solo', 'additional']:
        break
    prefixes.append(word)
    words.remove(word)
instruments[0] = " ".join(words)
prefix = prefixes ? " ".join(prefixes) + " " : ""
```

plugins/standardise_performers/standardise_performers.py
rdswift: Thanks.  I've added that to the checks.

plugins/standardise_performers/standardise_performers.py
rdswift: Changed to use lists and variable names changed to `prefixes` and `prefix` as written in the code provided by @Sophist-UK

plugins/standardise_performers/standardise_performers.py
rdswift: Code changed as suggested except that I had to use an `if-else` statement for the `prefix` assignment rather than the C style `?` statement shown.

Sophist-UK:
I know this is really a separate issue but lines 38-41 could be improved by using a Python List filter to eliminate the looping and need for an if statement i.e. something like `for key, values in list(filter(lambda key: key.startswith('performer:') or key.startswith('~performersort:'), metadata.rawitems())):`.

plugins/standardise_performers/standardise_performers.py
Sophist-UK: D'oh!! Too much PHP programming and not enough Python.

plugins/standardise_performers/standardise_performers.py
```diff
@@ -35,10 +35,7 @@
 
 
 def standardise_performers(album, metadata, *args):
-    for key, values in list(metadata.rawitems()):
-        if not key.startswith('performer:') \
-                and not key.startswith('~performersort:'):
-            continue
+    for key, values in list(filter(lambda filter_tuple: filter_tuple[0].startswith('performer:') or filter_tuple[0].startswith('~performersort:'), metadata.rawitems())):
```
phw: I know this was suggested by @Sophist-UK , but I am really not a fan of this. IMHO this makes readability much worse, and I honestly don't see the benefit.

Sophist-UK: Benefit is performance as `filter` is going to be much faster than manual checks - however the overall performance gain is small and I agree that readability is not as good.

rdswift: From my perspective, I learned how to use `filter()`.  [grin]

Seriously though, I can see both points so it really doesn't matter to me.  If you want me to revert that commit, just let me know.

phw: Yep, I would suggest to revert this part, then we can merge it

rdswift: Done.  Code switched back to separate test rather than incorporating a filter into the `for ... in ...` line.

For some reason Codacy is taking forever (24 minutes and counting).  I'll keep checking and if it shows any problems, I'll address them.

phw (approved):

phw:
Thanks. Yes, Codacy sometimes just hangs forever. We use it for work also, and have the same issue there. It is really nice when it works, but quite often it doesn't

PR #169, opened 2018-10-07
Hopefully this should fix the problems with the wikidata-genre plugin.

dns53:
The problems come from not tracking how many outstanding requests there are currently and only when there are no outstanding requests.
When there are more than one wikidata link for example this would get out of sync and cause problems.
This should hopefully fix this.

### DIFF
--- a/plugins/wikidata/wikidata.py
+++ b/plugins/wikidata/wikidata.py
@@ -78,9 +78,9 @@ class wikidata:
                     album._requests += 1
                     self.albums[item_id].append(album)
                 else:
-                    self.requests[item_id]=[metadata]
+                    self.requests[item_id] = [metadata]
                     album._requests += 1
-                    self.albums[item_id]=[album]
+                    self.albums[item_id] = [album]
 
                     log.debug('WIKIDATA: first request for this item')
 
@@ -131,10 +131,9 @@ class wikidata:
                 album._requests -= 1
                 log.debug('WIKIDATA:  TOTAL REMAINING REQUESTS %s' %
                           album._requests)
-                if album._requests == 0 :
+                if not album._requests:
                     self.albums[item_id].remove(album)
                     album._finalize_loading(None)
-            #del self.requests[item_id]
 
     def process_wikidata(self, wikidata_url, item_id):
         with self.lock:
@@ -202,7 +201,7 @@ class wikidata:
 
             for album in self.albums[item_id]:
                 album._requests -= 1
-                if album._requests == 0 :
+                if not album._requests:
                     self.albums[item_id].remove(album)
                     album._finalize_loading(None)
                 log.info('WIKIDATA:  TOTAL REMAINING REQUESTS %s' % album._requests)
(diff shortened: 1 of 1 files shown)


plugins/wikidata/wikidata.py
```diff
@@ -54,33 +64,34 @@ def process_request(self, metadata, album, item_id, type):
...
                     self.albums[item_id].append(album)
-                    return
-                self.requests[item_id] = [metadata]
-                album._requests += 1
-                self.albums[item_id] = [album]
-                log.debug('WIKIDATA: first request for this item')
+                else:
+                    self.requests[item_id]=[metadata]
+                    album._requests += 1
+                    self.albums[item_id]=[album]
```
phw: Please add some spacing around `=`

plugins/wikidata/wikidata.py
```diff
@@ -187,10 +202,10 @@ def parse_wikidata_response(self, item, item_id, response, reply, error):
 
             for album in self.albums[item_id]:
                 album._requests -= 1
-                album._finalize_loading(None)
-                log.info('WIKIDATA:  TOTAL REMAINING REQUESTS %s' %
-                         album._requests)
-            del self.requests[item_id]
+                if album._requests == 0 :
```
phw: You could write just `if not album._requests`

plugins/wikidata/wikidata.py
```diff
@@ -115,16 +125,21 @@ def musicbrainz_release_lookup(self, item_id, metadata, response, reply, error):
...
+                album._requests -= 1
+                log.debug('WIKIDATA:  TOTAL REMAINING REQUESTS %s' %
+                          album._requests)
+                if album._requests == 0 :
+                    self.albums[item_id].remove(album)
                     album._finalize_loading(None)
-                    log.debug('WIKIDATA:  TOTAL REMAINING REQUESTS %s' %
-                              album._requests)
-                del self.requests[item_id]
+            #del self.requests[item_id]
```
phw: If this is not needed anymore I would prefer the line to be removed completely

plugins/wikidata/wikidata.py
```diff
@@ -115,16 +125,21 @@ def musicbrainz_release_lookup(self, item_id, metadata, response, reply, error):
...
-            with self.lock:
-                for album in self.albums[item_id]:
-                    album._requests -= 1
+            log.info('WIKIDATA: no wikidata url found for item_id: %s ', item_id)
+        with self.lock:
+            for album in self.albums[item_id]:
+                album._requests -= 1
+                log.debug('WIKIDATA:  TOTAL REMAINING REQUESTS %s' %
+                          album._requests)
+                if album._requests == 0 :
```
phw: You could write just `if not album._requests`

phw (changes requested):
Thanks a lot for resubmitting this. Apart from a few minor stylistic comments this looks fine to me, and we definitely should get this merged ASAP

phw (approved):

PR #170, opened 2018-10-11
Updated lastfmplus plugin for Picard/Lastfm 2.0 APIs

Jormangeud:
I updated this to work with the newer APIs some time ago.

### DIFF
--- a/plugins/lastfmplus/__init__.py
+++ b/plugins/lastfmplus/__init__.py
@@ -112,11 +112,11 @@ def apply_translations_and_sally(tag_to_count, sally, factor):
     return list(ret.items())
 
 
-def _tags_finalize(album, metadata, tags, next):
+def _tags_finalize(album, metadata, tags, get_next):
     """Processes the tag metadata to decide which tags to use and sets metadata"""
 
-    if next:
-        next(tags)
+    if get_next:
+        get_next(tags)
     else:
         cfg = album.tagger.config.setting
 
@@ -334,7 +334,7 @@ def _tags_finalize(album, metadata, tags, next):
                     metadata["comment:Songs-DB_Custom1"] = "18%s0s" % str(metadata["originalyear"])[2]
 
 
-def _tags_downloaded(album, metadata, sally, factor, next, current, data, reply, error):
+def _tags_downloaded(album, metadata, sally, factor, get_next, current, data, reply, error):
     q = QtCore.QUrlQuery(reply.url())
     cachetag = q.queryItemValue('artist')
     if q.queryItemValue('method') == "Track.getTopTags":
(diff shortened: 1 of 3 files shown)


phw:
Thanks for this. Can you tackle the complaints Codacy had at https://app.codacy.com/app/MetaBrainz/picard-plugins/pullRequest?prid=2340007 ?

zas:
What's the status of this PR ?
@phw

phw:
@zas It's likely fine, I haven't really reviewed the code in detail. It's probably no secret that I'm not the biggest fan of the lastfmplus plugin and that I am more in favor of updating the simpler last.fm plugin with some of the more important features. But I appreciate that @Jormangeud took the time to update the lastfmplus plugin and I also realize there are users really wanting this plugin.

It's still a pitty it does not come with a .ui file, but that's how it is. If you are fine with merging this I am as well.

phw:
So I tried reviewing this. Considering only the scope of this PR, which is basically porting an existing plugin to updated APIs, this seems to be fine and clean. Good work on that, @Jormangeud 

But I still have general issues with this plugin. It has some big code complexity, and I don't personally want to maintain this or fix bugs in it. Also it has some very opinionated and strange handling of tags. Seriously, WTF is `comment:Songs-DB_Custom3`? And why does it need to store the year in 3 different tags, especially that strange `year_tag` tag? And why does this plugin make heavy use of `~id3:`, causing changes to ID3 tags the normal Picard user does not even see and cannot easily change?

I am fine with merging this if others are willing to fix issues should they arrive. But if there should be open issues that make this plugin the source of problems and nobody is willing to fix this I am personally much more likely to submit a PR to remove the plugin (again) then to sit down and fix it myself.

This plugin has some nice ideas, though, like proper white listing, extraction of data other then genre (especially something like "mood"), distinguishing between major and minor tags. If somebody does a clean implementation of the nice things for the standard last.fm plugin I would be more than happy to see this merged.

Just my 2.5 cents ;)

Sophist-UK:
Based on community comments, I would like to propose that this plugin be split into 2:

a. A plugin to get genres from lastFM and to and them to the genre tag.
b. A plugin to bring order and structure to the more general genre tag mess, as the lastFM plugin does for the lasFM genre mess.

phw:
@zas What shall we do with this? Really given the lack of responses here I doubt anyone will maintain this. Me certainly not, as I had stated previously. So I'm actually in favor or closing this PR. If someone is serious about keeping this plugin alive and is willing to do necessary changes to improve the code quality and keeping it compatible with future Picard releases they can use this PR as a starting point.

@Jormangeud Sorry for that, I really appreciate the effort you put into updating this. But a one time code drop is just not going to work in the long run. This plugin has been dragged along previously already.

zas:
Yup, I vote for closing it too. Not enough interest in this for it to be worth over the time.

PR #83, opened 2017-02-26
New plugin: Compatible TXXX frames

tungol:
This plugin improves the compatibility of ID3 tags by using only
a single value for TXXX frames. Multiple value TXXX frames
technically don't comply with the ID3 specification.

Related to https://github.com/metabrainz/picard/pull/635

PR #171, opened 2018-10-19
Compatible TXXX

phw:
This plugin improves the compatibility of ID3 tags by using only a single value for TXXX frames. Multiple value TXXX frames technically don't comply with the ID3 specification (but are wildly supported, hence this is the default in Picard).

See [PICARD-990](https://tickets.metabrainz.org/browse/PICARD-990), https://github.com/metabrainz/picard/pull/635, and https://github.com/metabrainz/picard-plugins/pull/83


phw:
Merged, I addressed the changes suggested in https://tickets.metabrainz.org/browse/PICARD-990

PR #172, opened 2018-10-19
Fix tangoinfo crashes

phw:
Problems:

1. Tango info plugin is crashing due to insufficient error handling 
2. Tango info is not returning any data.

I have fixed (1), and this version of the plugin can be used without Picard crashing. However, it still does not return any data. It looks like tango.info is doing something with the requests here, all requests made by this plugin gives status 200 responses, but with Content-Length: 0 and no body.

For any address on tango.info I can do successful requests with the browser or curl, but the exactly same request with the Picard webservice gives an empty body. I even tried to copy the request headers as I saw them used in Firefox to the Picard code to make sure there was no special treatment of our request, but same result:

```
GET https://tango.info:443/
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:62.0) Gecko/20100101 Firefox/62.0
Accept-Encoding: gzip, deflate, br
Accept-Language: de-DE,de;q=0.8,en-US;q=0.5,en;q=0.3
Cache-Control: no-cache
Pragma: no-cache
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8

200 OK
Date: Fri, 19 Oct 2018 08:57:34 GMT
Server: Apache
Vary: User-Agent
Upgrade: h2,h2c
Connection: Upgrade, Keep-Alive
X-Content-Type-Options: nosniff
X-Frame-Options: sameorigin
Referrer-Policy: strict-origin-when-cross-origin
X-XSS-Protection: 1; mode=block
Content-Security-Policy: default-src 'none'; frame-ancestors 'self'; frame-src 'self' https://www.google.com https://calendar.google.com  https://www.youtube.com; img-src 'self' https://*.openstreetmap.org; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; media-src 'self';
Content-Length: 0
Keep-Alive: timeout=5, max=100
Content-Type: text/html; charset=UTF-8
```

Also worth noting that the requests made by the plugin work on any other host but "tango.info". Any help debugging this is welcomed.

But even if we don't get this fixed soon I would be in favor of merging the changes here and make a new release. This will at least give a tango.info plugin that is not crashing Picard

### DIFF
--- a/plugins/tangoinfo/tangoinfo.py
+++ b/plugins/tangoinfo/tangoinfo.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 PLUGIN_NAME = "Tango.info Adapter"
-PLUGIN_AUTHOR = "Felix Elsner, Sambhav Kothari"
+PLUGIN_AUTHOR = "Felix Elsner, Sambhav Kothari, Philipp Wolfer"
 PLUGIN_DESCRIPTION = """
 <p>Load genre, date and vocalist tags from the online database
 <a href"http://tango.info">tango.info</a>.</p>
@@ -8,7 +8,7 @@ PLUGIN_DESCRIPTION = """
 it does not cause unnecessary server load for either MusicBrainz.org
 or tango.info</p>
 """
-PLUGIN_VERSION = "1.0"
+PLUGIN_VERSION = "1.1"
 PLUGIN_API_VERSIONS = ["2.0"]
 PLUGIN_LICENSE = "GPL-2.0"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
@@ -98,18 +98,16 @@ class TangoInfoTagger:
             #         )
             return
 
-        tint = str("0%s-%s-%s" % (
-                        barcode,
-                        str(track_metadata["discnumber"])\
-                                if track_metadata.get("discnumber") else "1",
-                        str(track_metadata["tracknumber"])
-                        ))
+        tint = "0%s-%s-%s" % (
+            barcode,
+            str(track_metadata.get("discnumber", "1")),
+            str(track_metadata["tracknumber"]))
 
         if barcode in self.albumpage_cache:
             if self.albumpage_cache[barcode]:
                 if not self.albumpage_cache[barcode].get(tint):
-                    log.debug("%s: No information on tango.info for barcode \
-                            %s" % (PLUGIN_NAME, barcode))
+                    log.debug("%s: No information on tango.info for barcode %s",
+                              PLUGIN_NAME, barcode)
                 else:
                     for field in ['genre', 'date', 'vocal']:
                         # Checks, as not to overwrite with empty data
@@ -117,8 +115,8 @@ class TangoInfoTagger:
                             track_metadata[field] = self\
                                     .albumpage_cache[barcode][tint][field]
             else:
-                log.debug("%s: No information on tango.info for barcode %s" \
-                        % (PLUGIN_NAME, barcode))
+                log.debug("%s: No information on tango.info for barcode %s",
+                          PLUGIN_NAME, barcode)
         else:
             #log.debug("%s: Adding to queue: %s, new track: %s" \
             #       % (PLUGIN_NAME, barcode, album._new_tracks[-1]))
@@ -133,9 +131,7 @@ class TangoInfoTagger:
 
         if self.albumpage_queue.append(barcode, (track, album, tint)):
             log.debug("%s: Downloading from tango.info: track %s, album %s, \
-                  with TINT %s" % (\
-                    PLUGIN_NAME, str(track), str(album), tint)
-            )
+                  with TINT %s", PLUGIN_NAME, str(track), str(album), tint)
 
             host = 'tango.info'
             port = 443
@@ -149,7 +145,8 @@ class TangoInfoTagger:
     def website_process(self, barcode, zeros, response, reply, error):
 
         if error:
-            log.error("%s: Error retrieving info for barcode %s", PLUGIN_NAME, barcode)
+            log.error("%s: Error retrieving info for barcode %s",
+                      PLUGIN_NAME, barcode)
             tuples = self.albumpage_queue.remove(barcode)
             for track, album, tint in tuples:
                 self.album_remove_request(album)
@@ -164,13 +161,12 @@ class TangoInfoTagger:
         if tangoinfo_album_data:
             if zeros > 0:
                 log.debug("%s: "
-            "tango.info does not seem to have data for barcode %s.  However, "
-            "retrying with barcode %s (i.e. the same with 0 prepended) was "
-            "successful. This most likely means either MusicBrainz or "
-            "tango.info has stored a wrong barcode for this release. You might "
-            "want to investigate this discrepancy and report it." \
-                          % (PLUGIN_NAME, barcode[zeros:], barcode)
-                )
+                    "tango.info does not seem to have data for barcode %s.  However, "
+                    "retrying with barcode %s (i.e. the same with 0 prepended) was "
+                    "successful. This most likely means either MusicBrainz or "
+                    "tango.info has stored a wrong barcode for this release. You might "
+                    "want to investigate this discrepancy and report it.",
+                    PLUGIN_NAME, barcode[zeros:], barcode)
 
             for track, album, tint in tuples:
                 tm = track.metadata
@@ -193,22 +189,21 @@ class TangoInfoTagger:
                     "Could not load album with barcode %s even with zero "
                     "prepended(%s).  This most likely means tango.info does "
                     "not have a release for this barcode (or MusicBrainz has a "
-                    "wrong barcode)" \
-                          % (PLUGIN_NAME, barcode[1:], barcode)
-                )
+                    "wrong barcode)",
+                    PLUGIN_NAME, barcode[1:], barcode)
                 for track, album, tint in tuples:
                     self.album_remove_request(album)
                 return
 
-            log.debug("%s: Retrying with 0-padded barcode for barcode %s" % \
-                    (PLUGIN_NAME, barcode))
+            log.debug("%s: Retrying with 0-padded barcode for barcode %s",
+                      PLUGIN_NAME, barcode)
 
             for track, album, tint in tuples:
                 retry_barcode = "0" + str(barcode)
                 retry_tint = "0" + tint
                 # Try again with new barcode, but at most two times(param zero)
                 self.website_add_track(
-                        album, track, retry_barcode, retry_tint, zeros=(zeros+1)
+                    album, track, retry_barcode, retry_tint, zeros=(zeros+1)
                 )
                 self.album_remove_request(album)
 
@@ -224,14 +219,14 @@ class TangoInfoTagger:
 
         # Check whether we have a concealed 404 and get the homepage
         if "<title>Contents - tango.info</title>" in response:
-            log.debug("%s: No album with barcode %s on tango.info" % \
-                      (PLUGIN_NAME, barcode))
+            log.debug("%s: No album with barcode %s on tango.info",
+                      PLUGIN_NAME, barcode)
             return
 
         table = table_regex.search(response)
         if not table:
-            log.warning("%s: No table found on page for barcode %s on tango.info" % \
-                        (PLUGIN_NAME, barcode))
+            log.warning("%s: No table found on page for barcode %s on tango.info",
+                        PLUGIN_NAME, barcode)
             return
 
         albuminfo = {}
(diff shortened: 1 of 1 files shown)


Sophist-UK:
When I view https://tango.info/ in a browser and examine the response headers I get these in both Chrome and Firefox:
```
HTTP/1.1 200 OK
Date: Fri, 19 Oct 2018 13:34:58 GMT
Server: Apache
Strict-Transport-Security: max-age=31536000;
Vary: User-Agent,Accept-Encoding
Upgrade: h2,h2c
Connection: Upgrade, Keep-Alive
X-Content-Type-Options: nosniff
X-Frame-Options: sameorigin
Referrer-Policy: strict-origin-when-cross-origin
X-XSS-Protection: 1; mode=block
Content-Security-Policy: default-src 'none'; frame-ancestors 'self'; frame-src 'self' https://www.google.com https://calendar.google.com  https://www.youtube.com; img-src 'self' https://*.openstreetmap.org; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; media-src 'self'; 
Keep-Alive: timeout=5, max=100
Content-Type: text/html; charset=UTF-8
Transfer-Encoding: chunked
```
I am not sure where the `Content-Length: 0` response header is coming from. It is possible that the site is looking at the user-agent request header from Picard and sending a different response, but that seems less likely. So I wonder whether it is the `Transfer-Encoding: chunked` that is the problem and that Qt cannot handle it.

Or is the home URL incorrect?

phw:
> I am not sure where the `Content-Length: 0` response header is coming from.

Good catch. I can confirm Curl is not showing a Content-Length header as well. So theory: tango.info does not send Content-Length at all, Qt gets confused by this and sets it to 0, which causes it also not to read the body (or read only 0 bytes). Plausible?

> Or is the home URL incorrect?

No, doesn't matter which URL you use. They all give an empty body in Picard :(

Sophist-UK:
The purpose of chunked transfers is to enable transfers of data (in chunks obviously) when the server may not have all the data to determine the length but wants to start to send it, and the client wants to start processing the data before the server has sent (or is even ready to send) the end of it.

Assuming that Picard is prepared to wait until all the data has been received before calling the reply call-back, it would still need code explicitly to collect all the chunked data and reconstitute it - and if you want have a call-back which processes each chunk as it is received, that would be a whole new API. So, I think we will need code in Picard's webservice object to handle chunked encoding - and even if we decide not to support it we should at least check the response headers and warn if we received a `Transfer-Encoding: Chunked` header.

phw:
But shouldn't the handling of such data be part of the underlying HTTP library (Qt's in our case)?

Sophist-UK:
My point is that Qt might be able to help a bit, but Chunked data is fundamentally different in concept because it is supporting real-time or response-time sensitive applications or streaming.

So, **if** you had a way of telling Qt explicitly to dechunk it, then Qt could do that for you and fake the Content-Length header. But unless Qt has that, then Picard Webservice code needs to be enhanced to handle web pages like this that use `Transfer-Encoding: Chunked` - by:

1. Spotting that it is chunked;
2. Collecting the chunks as they arrive, taking out the chunking special characters and consolidating the data;
3. Calling the response call-back only after the last chunk has been received and de-chunked.

Sophist-UK:
P.S. A quick google search for how to handled chunked data in Qt didn't come up with anything. So looks like we may be leading edge here. 

Edit: Did another search. One item says to "ignore QNetworkReply::RemoteHostClosedError error" - but I have no idea how you might do this. Also found some c++ code at https://github.com/syntelos/http-stream-qt that you might be able to use as a basis.

phw:
Unfortunately your chunked theory seems to be not the issue :( As I expected Qt just handles that for you normally. E.g. our webservice code just handles the demo page at https://jigsaw.w3.org/HTTP/ChunkedScript nicely, the response headers are:

```
200 OK
Cache-Control: max-age=0
Date: Sat, 20 Oct 2018 15:13:07 GMT
Transfer-Encoding: chunked
Content-Type: text/plain
Etag: "1j3k6u8:tikt981g"
Expires: Sat, 20 Oct 2018 15:13:06 GMT
Last-Modified: Mon, 18 Mar 2002 14:28:02 GMT
Server: Jigsaw/2.3.0-beta4
Strict-Transport-Security: max-age=15552015; includeSubDomains; preload
Public-Key-Pins: pin-sha256="cN0QSpPIkuwpT6iP2YjEo1bEwGpH/yiUn6yhdy+HNto="; pin-sha256="WGJkyYjx1QMdMe0UqlyOKXtydPDVrk7sl2fV+nNm1r4="; pin-sha256="LrKdTxZLRTvyHM4/atX2nquX9BeHRZMCxg3cf4rhc2I="; max-age=864000
X-Frame-Options: deny
X-XSS-Protection: 1; mode=block
```

The body contains the full 1000 lines of test data. Note how this shows `Transfer-Encoding: chunked`, but requests to tango.info don't.

EDIT: To clarify, requests to tango.info with curl or browser seem to have the Transfer-Encoding: chunked header, but my response header logging in Picard lists no such tag.

Sophist-UK:
Well I can't understand why Firefox / Chrome have `Transfer-encoding:Chunked` but Picard doesn't.

But funnily enough when I view https://jigsaw.w3.org/HTTP/ChunkedScript in Chrome / Firefox they  **don't** show a chunked header:
```
HTTP/2.0 200 OK
cache-control: max-age=0
date: Sat, 20 Oct 2018 15:45:04 GMT
content-type: text/plain
etag: "1j3k6u8:tikt981g"
expires: Sat, 20 Oct 2018 15:45:03 GMT
last-modified: Mon, 18 Mar 2002 14:28:02 GMT
server: Jigsaw/2.3.0-beta4
strict-transport-security: max-age=15552015; includeSubDomains; preload
public-key-pins: pin-sha256="cN0QSpPIkuwpT6iP2YjEo1bEwGpH/yiUn6yhdy+HNto="; pin-sha256="WGJkyYjx1QMdMe0UqlyOKXtydPDVrk7sl2fV+nNm1r4="; pin-sha256="LrKdTxZLRTvyHM4/atX2nquX9BeHRZMCxg3cf4rhc2I="; max-age=864000
x-frame-options: deny
x-xss-protection: 1; mode=block
X-Firefox-Spdy: h2
```

phw:
> But funnily enough when I view https://jigsaw.w3.org/HTTP/ChunkedScript in Chrome / Firefox they don't show a chunked header:

I think this is because of HTTP 2.0. If you use curl you also have to force HTTP 1.1 with `--http1.1` to see the header.

phw:
Which reminds me: I also tried using HTTP 2.0 in Picard, which worked but gave the same empty result for tango.info

Sophist-UK:
Well - for at least one of my web sites I have a .htaccess file which rejects a whole bunch of user agents (typically used for hackers probing your site) - though I give a 40x return code for them rather than 200. But it is entirely possible that tango,info does something similar. Have you tried spoofing the Picard user agent from a browser to see what happens when you issue the exact same request headers?

Sophist-UK:
What is also interesting is that the response headers include `Vary:User-agent` which suggests that they are coding to return different results based on the user agent. I guess it is entirely possible that they are just returning an empty page on the Picard user agent.

Edit: Sorry - you said that you spoofed that already. I have no further suggestions at this stage.

phw:
I copied the headers Firefox send one by one, including things like DNT :(

phw:
As I want to have at least the crashes fixed ASAP I merged this. I added an issue about this plugin not working, see https://github.com/metabrainz/picard-plugins/issues/179

ix5:
@Sophist-UK it seems the combination of HTTP/2.0, TLSv1.3 and upgrading to Qt 5 has introduced quite subtle issues that are very hard to track down.

I am now quite certain that the issue is in tango.info's Apache server sending some kind of unexpected/malformed headers, and Qt not catching them but instead bailing. Every browser agent and wget/curl seem to catch this and somehow silently recover.

## Exhibit A: mitmproxy
mitmproxy logs:
```
127.0.0.1:33560: server connect tango.info:443 (202.61.230.140:443)
127.0.0.1:33560: 202.61.230.140:443: HTTP/2 protocol error: \
Received header value surrounded by whitespace \
b"default-src 'none'; frame-ancestors 'self'; frame-src 'self' \
https://www.google.com https://calendar.google.com https://accounts.google.com \
https://www.youtube.com; img-src 'self' https://anna.info/x/ https://*.openstreetmap.org; \
script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; media-src 'self'; "
127.0.0.1:33560: server disconnect tango.info:443 (202.61.230.140:443)
```
(note trailing whitespace after `media-src: self; `)

This is due to the `h2` package (for HTTP/2 I assume) throwing an error.
Monkey-patching `/usr/lib/python3.10/site-packages/h2/utilities.py` with the following:
```python
def _reject_surrounding_whitespace(headers, hdr_validation_flags):
    for header in headers:
        header_tmp = [hdr for hdr in header]
        if header[0][0] in _WHITESPACE:
            header_tmp[0] = header[0][1:]
        if header[0][-1] in _WHITESPACE:
            header_tmp[0] = header[0][:-1]
        if header[1] and header[1][0] in _WHITESPACE:
            header_tmp[1] = header[1][1:]
        if header[1] and header[1][-1] in _WHITESPACE:
            header_tmp[1] = header[1][:-1]
        yield header_tmp
```
... yields these headers (in mitmproxy):
```
HTTP/2.0 200
date :	Thu, 27 Jan 2022 17:47:25 GMT
server :	Apache/2.4.41 (Ubuntu)
vary :	User-Agent
referrer-policy :	strict-origin-when-cross-origin
x-xss-protection :	1; mode=block
content-security-policy :	default-src 'none'; frame-ancestors 'self'; frame-src 'self' https://www.google.com https://calendar.google.com https://accounts.google.com https://www.youtube.com; img-src 'self' https://anna.info/x/ https://*.openstreetmap.org; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; media-src 'self';
content-length :	0
content-type :	text/html; charset=UTF-8
```

Note: `Content-length` is 0 there again.

## Exhibit B: Telegram
This is just something that occurred to me, other Qt-based apps might have issues as well. And it turns out, Telegram Desktop (Qt-based) will not render a link preview for tango.info links!

## Exhibit C: Proxying
Then there's this interesting bit: Proxying the connection via an intermediary nginx server makes picard receive correct data (correct content-length, actual content served) and the plugin works fine.
However, mitmproxy complains about `HTTP/2 protocol error: Connection-specific header field present: b'upgrade'.`

nginx config:
```
server {
    listen 443 ssl http2;
    listen [::]:443 ssl http2;
    server_name foo.org;
    ssl_certificate_key /etc/ssl-certs/privkey.pem;
    root /srv/www/;
    index index.html;
    access_log  /var/log/nginx/access.log timed_combined;
    error_log   /var/log/nginx/error.log;
    proxy_ssl_session_reuse off;
    location / {
        proxy_pass https://tango.info;
        proxy_ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3;
        client_max_body_size    0;
        proxy_connect_timeout   36000s;
        proxy_read_timeout      36000s;
        proxy_send_timeout      36000s;
        send_timeout            36000s;
        proxy_request_buffering off;
    }
}
```

ix5:
Might be interesting to either of you, here are two Wireshark `.pcapng` files
(using SSL stripping via `SSLKEYLOGFILE="$HOME/.ssl-key.log" firefox`)

`tango_info.pcapng` is the regular connection to tango.info's Apache server
`tango_info_proxy.pcapng` is the proxied connection via my own test server.

[tango_info_pcapng.zip](https://ix5.org/files/tango_info_pcapng.zip)

ix5:
Boy, that was a "fun" one to track down. The issue is that tango.info chokes on the `:443` part of the `Host` header.

This succeeds:
```
curl -v -H 'Host: tango.info' https://tango.info/08427328130110
```
This fails with a zero-length response body:
```
curl -v -H 'Host: tango.info:443' https://tango.info/08427328130110
```

Currently, Picard calls Qt's networking stack in such a way that a `QUrl` is constructed that always sets the `port` part, which will be sent with the `Host` header:
https://github.com/metabrainz/picard/blob/89bc0cbecb0e7466fc1f17938f3e76aefbcb3eac/picard/util/__init__.py#L573-L598

I'm not sure setting that header is even necessary - perhaps for SNI stuff?

I've monkey-patched around this, but I'm hoping I'll get a response from the tango.info maintainer on fixing this on their end.
https://git.ix5.org/felix/picard-tangoinfo/src/commit/ae0956c1d17044b5d4be091be377d5c5714c6ac2/tangoinfo.py#L33-L42
https://git.ix5.org/felix/picard-tangoinfo/src/commit/ae0956c1d17044b5d4be091be377d5c5714c6ac2/tangoinfo.py#L151-L166

PR #173, opened 2018-10-29
Add better error handling for AcousticBrainz plugin.

m-szu:
Problem: 
-Poor error handling in AcousticBrainz plugin

Solution:
- added _error_ parameter usage in _result()_
- checking does json key "highlevel" exist in _result()_
- changed _webservice.download()_ to _webservice.get()_ to not use _load_json()_

### DIFF
--- a/plugins/acousticbrainz/acousticbrainz.py
+++ b/plugins/acousticbrainz/acousticbrainz.py
@@ -23,7 +23,7 @@ PLUGIN_DESCRIPTION = '''Uses AcousticBrainz for mood and genre.
 WARNING: Experimental plugin. All guarantees voided by use.'''
 PLUGIN_LICENSE = "GPL-2.0"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
-PLUGIN_VERSION = "1.1"
+PLUGIN_VERSION = "1.1.1"
 PLUGIN_API_VERSIONS = ["2.0"]
 
 from functools import partial
--- a/plugins/acousticbrainz/acousticbrainz.py
+++ b/plugins/acousticbrainz/acousticbrainz.py
@@ -39,6 +39,11 @@ ratecontrol.set_minimum_delay((ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT), 50)
 
 
 def result(album, metadata, data, reply, error):
+    if error or not "highlevel" in data:
+        album._requests -= 1
+        album._finalize_loading(None)
+        return
+
     moods = []
     genres = []
     try:
@@ -60,12 +65,13 @@ def result(album, metadata, data, reply, error):
 
 
 def process_track(album, metadata, release, track):
-    album.tagger.webservice.download(
+    album.tagger.webservice.get(
         ACOUSTICBRAINZ_HOST,
         ACOUSTICBRAINZ_PORT,
         "/%s/high-level" % (metadata["musicbrainz_recordingid"]),
         partial(result, album, metadata),
-        priority=True
+        priority=True,
+        parse_response_type=None
     )
     album._requests += 1
 


phw:
First off: Your commit looks fine to me, that is how I would expect the plugin to look like.

BUT I tested this by changing ACOUSTICBRAINZ_HOST and the path to something that will not return JSON, and then we get an actual crash because Picard does not treat errors when parsing the response as errors in the request. That means if the response is not JSON, the response parser is failing, but the web service module only logs this as an error but still returns an result with `error` set to `None`.

But now instead of a parsed JSON object `data` is a `QByteArray`, and your check of `"highlevel" in data` causes an exception. I see three options to deal with this:

1. Revert the `parse_response_type` and use `load_json` in the plugin again.
2. Check if `data` is a `QByteArray`. This is a bit meh, because the idea of using `parse_response_type` was to simplify the code, not to complicate it :D
3. Make sure Picard sets the `error` when a parser response is encountered.

I think we should got with 1) for now, because it is cleaner than 2) and ensures the plugin works without crashes in current Picard. We should also do 3) (there is already a related ticket at [PICARD-1338](https://tickets.metabrainz.org/browse/PICARD-1338)), but this will only fix it for future Picard releases.

phw (approved):

plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -60,14 +65,13 @@ def result(album, metadata, data, reply, error):
 
 
 def process_track(album, metadata, release, track):
-    album.tagger.webservice.download(
+    album.tagger.webservice.get(
```
phw, line 68: Have you actually run this? I think this will crash right now, you either have to switch back this to the `download` method or set `parse_response_type=None`. Otherwise the built-in JSON parsing in the webservice still happens, leading to errors when doing `load_json` again!

phw (changes requested):
Sorry, I just realized there is a bug with the reversal of the load_json change. That needs to get fixed.

plugins/acousticbrainz/acousticbrainz.py
m-szu: Yeah... sorry for that.
Also rebased all commits.

plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -39,6 +39,11 @@
 
 
 def result(album, metadata, data, reply, error):
+    if error or not "highlevel" in data:
+        album._requests -=1
```
zas: A whitespace after the `=` is missing (PEP8)

zas (changes requested):

plugins/acousticbrainz/acousticbrainz.py
```diff
@@ -60,14 +65,14 @@ def result(album, metadata, data, reply, error):
...
         ACOUSTICBRAINZ_PORT,
         "/%s/high-level" % (metadata["musicbrainz_recordingid"]),
         partial(result, album, metadata),
-        priority=True
+        priority=True,
+        parse_response_type=None
     )
     album._requests += 1
 
-
```
zas: This whiteline should stay.

zas (changes requested):

zas (approved):

phw:
@zas, @samj1912  We don't have a policy for this, but should we have the plugin version number bumped for each PR merged? IMHO it makes sense, so this would mean to probably bump the version number to `1.1.1` or so

zas:
> @zas, @samj1912 We don't have a policy for this, but should we have the plugin version number bumped for each PR merged? IMHO it makes sense, so this would mean to probably bump the version number to `1.1.1` or so

I'd say, in general, it has to be increased on every significant change. That's up to the plugins authors to decide though.

m-szu:
> > @zas, @samj1912 We don't have a policy for this, but should we have the plugin version number bumped for each PR merged? IMHO it makes sense, so this would mean to probably bump the version number to `1.1.1` or so
> 
> I'd say, in general, it has to be increased on every significant change. That's up to the plugins authors to decide though.

So my job is done, isn't it?

phw (approved):

phw:
@virusMac Yes, all done, thanks a lot. Sorry, I had not realized I never updated my review to approve. I merged the PR and bumped the version number.

PR #174, opened 2018-10-30
Handle exceptions with in Tonal Rhythm Plugin

abhinavohri:
When AcousticBrainz was down during a recent migration and was returning HTML instead of the expected JSON data it caused Picard to crash if the AB plugin was enabled. I have update the AcousticBrainz Tonal-Rhythm plugin for better error handling to avoid crashes.

### DIFF
--- a/plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
+++ b/plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
@@ -27,7 +27,7 @@ from the AcousticBrainz database.<br/><br/>
 '''
 PLUGIN_LICENSE = "GPL-2.0"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
-PLUGIN_VERSION = '1.1'
+PLUGIN_VERSION = '1.1.1'
 PLUGIN_API_VERSIONS = ["2.0"]  # Requires support for TKEY which is in 1.4
 
 from picard import log
@@ -45,7 +45,7 @@ ratecontrol.set_minimum_delay((ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT), 50)
 class AcousticBrainz_Key:
 
     def get_data(self, album, track_metadata, trackXmlNode, releaseXmlNode):
-        if not musicbrainz_recordingid in track_metadata:            
+        if not musicbrainz_recordingid in track_metadata:
             log.error("%s: Error parsing response. No MusicBrainz recording id found.",
                       PLUGIN_NAME)
             return
--- a/plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
+++ b/plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
@@ -34,6 +34,7 @@ from picard import log
 from picard.metadata import register_track_metadata_processor
 from functools import partial
 from picard.webservice import ratecontrol
+from picard.util import parse_json
 
 ACOUSTICBRAINZ_HOST = "acousticbrainz.org"
 ACOUSTICBRAINZ_PORT = 80
@@ -44,9 +45,14 @@ ratecontrol.set_minimum_delay((ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT), 50)
 class AcousticBrainz_Key:
 
     def get_data(self, album, track_metadata, trackXmlNode, releaseXmlNode):
+        if not musicbrainz_recordingid in track_metadata:            
+            log.error("%s: Error parsing response. No MusicBrainz recording id found.",
+                      PLUGIN_NAME)
+            return
         recordingId = track_metadata['musicbrainz_recordingid']
         if recordingId:
-            log.debug("%s: Add AcusticBrainz request for %s (%s)", PLUGIN_NAME, track_metadata['title'], recordingId)
+            log.debug("%s: Add AcousticBrainz request for %s (%s)",
+                      PLUGIN_NAME, track_metadata['title'], recordingId)
             self.album_add_request(album)
             path = "/%s/low-level" % recordingId
             return album.tagger.webservice.get(
@@ -54,16 +60,23 @@ class AcousticBrainz_Key:
                         ACOUSTICBRAINZ_PORT,
                         path,
                         partial(self.process_data, album, track_metadata),
-                        priority=True, important=False)
-        return
+                        priority=True,
+                        important=False,
+                        parse_response_type=None)
 
     def process_data(self, album, track_metadata, response, reply, error):
         if error:
             log.error("%s: Network error retrieving acousticBrainz data for recordingId %s",
-                PLUGIN_NAME, track_metadata['musicbrainz_recordingid'])
+                      PLUGIN_NAME, track_metadata['musicbrainz_recordingid'])
             self.album_remove_request(album)
             return
-        data = response
+        try:
+            data = parse_json(response)
+        except JSONDecodeError:
+            log.error("%s: Network error retrieving AcousticBrainz data for recordingId %s",
+                      PLUGIN_NAME, track_metadata['musicbrainz_recordingid'])
+            self.album_remove_request(album)
+            return
         if "tonal" in data:
             if "key_key" in data["tonal"]:
                 key = data["tonal"]["key_key"]


plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,17 +44,23 @@
...
-            return album.tagger.webservice.get(
-                        ACOUSTICBRAINZ_HOST,
-                        ACOUSTICBRAINZ_PORT,
-                        path,
-                        partial(self.process_data, album, track_metadata),
-                        priority=True, important=False)
+        try:
+            recordingId = track_metadata['musicbrainz_recordingid']
+            if recordingId:
+                log.debug("%s: Add AcusticBrainz request for %s (%s)", PLUGIN_NAME, track_metadata['title'], recordingId)
```
zas: Unrelated to this PR, but you may want to fix the typo in this debug text, AcusticBrainz -> AcousticBrainz

abhinavohri:
```suggestion
                log.debug("%s: Add AcousticBrainz request for %s (%s)", PLUGIN_NAME, track_metadata['title'], recordingId)
```

phw:
The main issue that actually lead to crashes in the past is in the `process_data` function, which just tests `if "tonal" in data:`. This fails if the service did not return valid JSON and parsing fails, in this case data will be of type `QByteArray`. See also my comment at https://github.com/metabrainz/picard-plugins/pull/173#issuecomment-434222461

We probably should check for exceptions in process_data, even if this arguably is a Picard bug (Picard should I think set `error` in these cases).

@zas, @samj1912, what are your thoughts on this issue?

abhinavohri:
@phw When is the data of type `QByteArray`, if the service fails or works properly. What is the data type in other case ? I am a bit confused. 

I have a plan. Please tell if its appropriate.
If we are aware in each case of the data type we can check it, before performing rest of code logic and skip it in case of wrong data. Another option I think can be to enclose the function in try/catch block and throw an exception in case of error.

phw:
@abhi-ohri It is confusing :D I try to explain:

Usually the response of a request would be `QByteArray`, because this works for any response of the server, even binary data. But the `WebService.get` function provides a parameter `parse_response_type`, which is set to `'json'` by default (currently it can bei either `None`, `'json'` or `'xml'`). If this is set the web service module tries to parse the response with the given parser. In the case of `'json'` a dict with the JSON data is returned instead of a `QByteArray`.

So far everything is well. In our case `process_data` expects such a dict, and it also makes sure to test if the keys it expects in the data are present in the JSON dict (with e.g. `if "tonal" in data:`). All looks good.

Problems come up when the server sends a response that cannot be parsed as JSON (just text, or invalid JSON). In this case the web service module tries to parse the data as JSON, fails, and falls back to returning just a `QByteArray` again. This is bad, because `process_data` does not expect this and `"tonal" in data` raises an exception.

I think this is an issue with Picard and Picard should return an `error` object in this case. We also have a ticket in Picard to fix this ([PICARD-1338](https://tickets.metabrainz.org/browse/PICARD-1338)). But until we have released a fixed version of Picard the plugins itself should handle this.

We could wrap most of the code in `process_data` into a try block for now.

phw:
@abhi-ohri I think this is currently the save thing to do. But let's wait for some feedback of other developers

zas:
> @abhi-ohri I think this is currently the save thing to do. But let's wait for some feedback of other developers

I agree. A proper fix is required on Picard side, but for now let's just ensure plugins catch exceptions.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,17 +44,23 @@
...
+                self.album_add_request(album)
+                path = "/%s/low-level" % recordingId
+                return album.tagger.webservice.get(
+                            ACOUSTICBRAINZ_HOST,
+                            ACOUSTICBRAINZ_PORT,
+                            path,
+                            partial(self.process_data, album, track_metadata),
+                            priority=True, important=False)
+        except Exception as e:
+            log.error("%s: Track %s (%s) Error parsing response: %s", PLUGIN_NAME, track_metadata["musicbrainz_recordingid"], track_metadata["title"], str(e))
```
zas: Please wrap arguments, this line is far too long.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -63,22 +69,26 @@ def process_data(self, album, track_metadata, response, reply, error):
...
+                    track_metadata["key"] = key
+                    log.debug("%s: Track '%s' is in key %s", PLUGIN_NAME, track_metadata["title"], key)
+            if "rhythm" in data:
+                if "bpm" in data["rhythm"]:
+                    bpm = int(data["rhythm"]["bpm"] + 0.5)
+                    track_metadata["bpm"] = bpm
+                    log.debug("%s: Track '%s' has %s bpm", PLUGIN_NAME, track_metadata["title"], bpm)
+            self.album_remove_request(album)
+        except Exception as e:
+            log.error("%s: Network error retrieving acousticBrainz data for recordingId %s", PLUGIN_NAME,
```
zas: The format string has only 2 placeholders, but you pass 4 arguments, that's incorrect.

zas (changes requested):

abhinavohri:
@zas Made the required changes. Anything else?

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,41 +44,55 @@
...
+                    log.debug("%s: Track '%s' is in key %s",
+                              PLUGIN_NAME, track_metadata["title"], key)
+            if "rhythm" in data:
+                if "bpm" in data["rhythm"]:
+                    bpm = int(data["rhythm"]["bpm"] + 0.5)
+                    track_metadata["bpm"] = bpm
+                    log.debug("%s: Track '%s' has %s bpm",
+                              PLUGIN_NAME, track_metadata["title"], bpm)
+            self.album_remove_request(album)
+        except Exception as e:
```
sambhav: Not a good way to solve the problem by putting such a huge try except block, that too with a generic exception.

sambhav (changes requested):

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
sambhav: It would be better to limit the range of the try clause - also, only catch KeyError (and Type Errors?)

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,41 +44,55 @@
...
-            self.album_add_request(album)
-            path = "/%s/low-level" % recordingId
-            return album.tagger.webservice.get(
-                        ACOUSTICBRAINZ_HOST,
-                        ACOUSTICBRAINZ_PORT,
-                        path,
-                        partial(self.process_data, album, track_metadata),
-                        priority=True, important=False)
+        try:
+            recordingId = track_metadata['musicbrainz_recordingid']
```
sambhav: Here as well, this seems to be the only possibility of an exception, that too a specific type.

sambhav (changes requested):

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
abhinavohri: @samj1912 I don't think the range of try clause cannot be limited here because the problem is due to the line ` if "tonal" in data:` itself. As to the catch block, since we do not want specific error handling for an exception, how about just logging the exception instead of logging the fixed message ? 

 Is it OK or should I go on to use a specific catch block? Also, as I said I don't know if its possible to try block here? Please guide me if its possible.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
abhinavohri: So I should include just the first line in the try block? @samj1912

abhinavohri:
@phw Since the problem occurs only when `process_data` receives a `QByteArray`? How about checking if the `data` is an instance of `QByteArray` type and then proceeding further? We could eliminate try/catch altogether with it.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
phw: Another approach would be to go the way the other AcousticBrainz plugin is doing it and not rely on Picard's automatic JSON parsing (because that's where the problem originates):

1. Use `parse_response_type=None` in the call to `album.tagger.webservice.get`
2. Use `load_json` from `picard.util` to parse the response and only wrap that into a try catch. So instead o the current

    ```
    data = response
    ```

    There would be something like:
    ```
    try:
        data = parse_json(response)
    except JSONDecodeError as e:
        # Error handling here
    ```

This will avoid the big try block to catch everything and replace it with a very specific error handling

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
phw: See my comment below for a better approach

phw (changes requested):

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
abhinavohri: @phw I have fixed the other try block. Here, I think there isn't a possibility of error unless `track_metadata` throws a `TypeError`. So what should I do with this try block?

phw: Agreed, I think we don't need the try block here. You can just test for
```
if not musicbrainz_recordingid in track_metadata:
    return
```

At the start. Also having that finally block currently is wrong, because this is handled in `process_data` with the call to `self.album_remove_request`.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,26 +45,41 @@
...
+                      PLUGIN_NAME, track_metadata['musicbrainz_recordingid'])
             self.album_remove_request(album)
             return
-        data = response
+        try:
+            data = parse_json(response)
+        except JSONDecodeError as e:
+            log.error("%s: Network error retrieving acousticBrainz data for recordingId %s",
+                      PLUGIN_NAME, track_metadata['musicbrainz_recordingid'])
+            return
```
phw, line 79: `self.album_remove_request(album)` needs to be called here before the return.

phw (changes requested):
Thanks for the changes, this is going in the right direction. Also I agree with you assessment of the first try block. See my comments for changes

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -72,13 +88,16 @@ def process_data(self, album, track_metadata, response, reply, error):
...
+                          PLUGIN_NAME, track_metadata["title"], key)
         if "rhythm" in data:
             if "bpm" in data["rhythm"]:
                 bpm = int(data["rhythm"]["bpm"] + 0.5)
                 track_metadata["bpm"] = bpm
-                log.debug("%s: Track '%s' has %s bpm", PLUGIN_NAME, track_metadata["title"], bpm)
+                log.debug("%s: Track '%s' has %s bpm",
+                          PLUGIN_NAME, track_metadata["title"], bpm)
         self.album_remove_request(album)
+
```
phw: Please don't add this blank line. We mostly follow [PEP 8](https://www.python.org/dev/peps/pep-0008/#blank-lines), and it says to use a single blank line for method definition inside classes.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,26 +45,37 @@
 class AcousticBrainz_Key:
 
     def get_data(self, album, track_metadata, trackXmlNode, releaseXmlNode):
+        if not musicbrainz_recordingid in track_metadata:            
+            log.error("%s: Error parsing response", PLUGIN_NAME)
```
phw: The log message should better reflect the issue encountered

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -72,13 +84,14 @@ def process_data(self, album, track_metadata, response, reply, error):
...
                 track_metadata["key"] = key
-                log.debug("%s: Track '%s' is in key %s", PLUGIN_NAME, track_metadata["title"], key)
+                log.debug("%s: Track '%s' is in key %s",
+                          PLUGIN_NAME, track_metadata["title"], key)
         if "rhythm" in data:
             if "bpm" in data["rhythm"]:
                 bpm = int(data["rhythm"]["bpm"] + 0.5)
                 track_metadata["bpm"] = bpm
-                log.debug("%s: Track '%s' has %s bpm", PLUGIN_NAME, track_metadata["title"], bpm)
-        self.album_remove_request(album)
```
phw: `self.album_remove_request(album)` must stay here. It's very important otherwise Picard will still report a pending request.

phw (changes requested):

abhinavohri:
@phw Improved the log message  added call to `self.album_remove_request()`.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,26 +45,38 @@
...
         if error:
             log.error("%s: Network error retrieving acousticBrainz data for recordingId %s",
-                PLUGIN_NAME, track_metadata['musicbrainz_recordingid'])
+                      PLUGIN_NAME, track_metadata['musicbrainz_recordingid'])
+            self.album_remove_request(album)
+            return
+        try:
+            data = parse_json(response)
+        except JSONDecodeError:
+            log.error("%s: Network error retrieving acousticBrainz data for recordingId %s",
```
Freso: Extremely minor nitpick, but: "acousticBrainz" should be "AcousticBrainz". :)
```suggestion
            log.error("%s: Network error retrieving AcousticBrainz data for recordingId %s",
```

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -72,12 +85,14 @@ def process_data(self, album, track_metadata, response, reply, error):
                     if scale == "minor":
                         key += "m"
                 track_metadata["key"] = key
-                log.debug("%s: Track '%s' is in key %s", PLUGIN_NAME, track_metadata["title"], key)
+                log.debug("%s: Track '%s' is in key %s",
+                          PLUGIN_NAME, track_metadata["title"], key)
```
Freso: This seems like a change unrelated to the issue/PR?

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -72,12 +85,14 @@ def process_data(self, album, track_metadata, response, reply, error):
...
         if "rhythm" in data:
             if "bpm" in data["rhythm"]:
                 bpm = int(data["rhythm"]["bpm"] + 0.5)
                 track_metadata["bpm"] = bpm
-                log.debug("%s: Track '%s' has %s bpm", PLUGIN_NAME, track_metadata["title"], bpm)
+                log.debug("%s: Track '%s' has %s bpm",
+                          PLUGIN_NAME, track_metadata["title"], bpm)
```
Freso: This seems like a change unrelated to the issue/PR?

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,26 +45,38 @@
 class AcousticBrainz_Key:
 
     def get_data(self, album, track_metadata, trackXmlNode, releaseXmlNode):
+        if not musicbrainz_recordingid in track_metadata:            
+            log.error("%s: Error parsing response. No musicbrainz recording id found.",
```
Freso: Capitalisation matters!

```suggestion
            log.error("%s: Error parsing response. No MusicBrainz recording id found.",
```

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
```diff
@@ -44,26 +45,38 @@
...
+                      PLUGIN_NAME, track_metadata['title'], recordingId)
             self.album_add_request(album)
             path = "/%s/low-level" % recordingId
             return album.tagger.webservice.get(
                         ACOUSTICBRAINZ_HOST,
                         ACOUSTICBRAINZ_PORT,
                         path,
                         partial(self.process_data, album, track_metadata),
-                        priority=True, important=False)
-        return
```
Freso, line 58: Maybe I missed some discussion: was this `return` removed on purpose?

Freso (changes requested):
A couple of very minor nitpicks (capitalisation/naming) and a few questions (that may be non-issues). Other than that, the code overall looks good to me.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
abhinavohri: I wrapped all logging statements for uniform look of code. Should I revert it?

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
abhinavohri: Should I revert?

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
abhinavohri: Yeah, Sorry for error.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
abhinavohri: Fixed.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
abhinavohri: Yes, this `return` is unneeded.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
Freso: Let's see what the others think, but it might be nice to not clutter the git history. If someone is looking at this code later, they may think this formatting change was required to "Handle exceptions with in Tonal Rhythm Plugin" rather than merely be an inconsequential formatting change. (Which are fine, but should be their own PRs—or at the very least their own commit. :))

You may want to read e.g., https://www.freshconsulting.com/atomic-commits/ for more background on why I care. :)

abhinavohri: I understand it . I have removed the formatting change.

plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
phw: :+1:

phw:
@abhi-ohri This is looking good now, thanks. Can you squash all commits into a single commit? The changes are small enough that they don't warrant having 11 commits. I think then we are good to merge this

zas (approved):

phw (approved):

PR #175, opened 2018-10-31
Create Classical Extras Plugin for Picard v2

MetaTunes:
Version 2.0.2 of Classical Extras


Sophist-UK:
Please change the PR title to something more meaningful like "Create Classical Extras Plugin for Picard v2".

MetaTunes:
Sorry - default name was supplied by GitHub!

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+         },
+        {'option': 'cwp_instruments_tag',
+         'name': 'instruments tag',
+         'type': 'Text',
+         'default': 'instrument'
+         },
+        {'option': 'cwp_instruments_MB_names',
+         'name': 'use MB instrument names',
+         'type': 'Boolean',
+         'default': T
```
mineo: This is named incorrectly, it's not a list, but a dictionary.

mineo: Isn't this comment incorrect? The second check in this function checks if `multi_list` is a list.

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+    if DEBUG or INFO:
+        write_log(release_id, 'debug', 'Parsing data - looking for %s', match)
+    if INFO:
+        write_log(release_id, 'info', 'Looking in object: %s', obj)
+    if isinstance(obj, list):
+        objlen = len(obj)
+        for i, item in enumerate(obj):
+            if isinstance(item, XmlNode):
+                item = item.__dict__
+            if INFO:
```
mineo, line 423: To be completely honest, the one thing that makes this pull request incredible difficult to review is the ratio of logging statements to real code. Just scrolling around here a bit, there's about 35 lines logging stuff between lines 460 and 520 - that's about 50%!

Here, for example, the only real code appears to be:

```python
        for i, item in enumerate(obj):
            if isinstance(item, XmlNode):
                item = item.__dict__
            parse_data(release_id, item, response_list, *match)
```

That's 5 lines, but with the logging statements it's taking up 20 lines.

I get that logging things is important, but this is really excessive.

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+RE_SCALES = r'(major|minor)?(?:\b|$)'
+RE_KEYS = re.compile(
+    RE_NOTES + RE_ACCENTS + RE_SCALES,
+    re.UNICODE | re.IGNORECASE)
+
+# LOGGING
+
+# If logging occurs before any album is loaded, the startup log file will
+# be written
+log_files = collections.defaultdict(dict)
```
mineo, line 141: Instead of keeping a dictionary of release ids to logfiles, why not (if you really want to keep the per-release logging), keep a dictionary mapping release ids to [loggers](https://docs.python.org/3/library/logging.html?highlight=logging#logger-objects)? That would allow you to set the log level in each logger separately instead of wrapping each `write_log` call in some `if $LOG_LEVEL` structure.

Also, `collections.defaultdict(dict)` looks wrong. Shouldn't that be something like `collections.defaultdict(<function that opens a log file for a release id>)`?

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+    if log_type != 'info' and log_type != 'basic':  # i.e. non-custom log items
+        message2 = PLUGIN_NAME + ': ' + message
+    else:
+        message2 = message
+    if log_type == 'debug' and options["log_debug"]:
+        if release_id in release_status and 'debug' in release_status[release_id]:
+            add_list_uniquely(release_status[release_id]['debug'], msg)
+        else:
+            release_status[release_id]['debug'] = [msg]
+        if args:
```
mineo: These `if args` tests are unnecessary. If there are no args, `log.foo(message, *args)` will not pass any additional arguments to `log.foo`.

plugins/classical_extras/__init__.py
mineo, line 2820: I'd find this easier to read if the docstring of each variable would be on the line before the variable itself.

mineo, line 2853: If this is maps to a boolean, `collections.defaultdict(dict)` (which indicates this maps to another dictionary) is incorrect. This appears to be true for some other dicts here as well.

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+    re.UNICODE | re.IGNORECASE)
+
+# LOGGING
+
+# If logging occurs before any album is loaded, the startup log file will
+# be written
+log_files = collections.defaultdict(dict)
+# entries are release-ids: to keep track of which log files are open
+release_status = collections.defaultdict(dict)
+# release_status[release_id]['works'] = True indicates that we are still processing works for release_id
```
mineo, line 144: There appear to be a lot more valid keys for the dictionaries saved in `release_status`. I think it would be good to not use dictionaries here, but something like `ReleaseStatus` objects that are always at least correctly initialised.

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+                'Found music file: %r',
+                music_file_found)
+        for section in all_sections:
+            if options[override[section]]:
+                parent_section = parent_sections[section]
+                if parent_section in file_options and file_options[parent_section]:
+                    try:
+                        options_dict = file_options[parent_section]['Classical Extras'][sect_text[parent_section] + ' options']
+                    except TypeError:
+                        if options['log_error']:
```
mineo, line 841: Wouldn't it be useful to actually log the error instead of "something went wrong"?

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+        if options['log_info']:
+            write_log(
+                release_id,
+                'info',
+                'Get_options is returning options shown below for file: %s',
+                music_file_found)
+            write_log(release_id, 'info', options_dict)
+
+
+def plugin_options(option_type):
```
mineo, line 948: Please move this 800-line function into a file like `const.py` that just provides constants (`ARTISTS_OPTIONS`, ...) for all these lists. It looks like every user needs to know which dictionary they want anyways.

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+            path.append(current_node)
+            current_node = node
+        elif stream.isEndElement():
+            current_node = path.pop()
+        elif stream.isCharacters():
+            current_node.text += str(stream.text())
+    return document
+
+
+def parse_data(release_id, obj, response_list, *match):
```
mineo, line 382: I might be wrong here, but it looks like `parse_data` itself is the only function that sets `response_list` to a non-empty list. In that case, you could make it a keyword-only argument, so other callers can just not specify it:

```python
def parse_data(release_id, obj, *match, response_list=None):
    if response_list is None:
        response_list = []
    [...]
    parse_data([...], *match, response_list=response_list)
    [...]
```

plugins/classical_extras/__init__.py
mineo: This is never written to and it looks like a constant that probably doesn't change a lot, you can move it into `const.py`.

mineo: Constant dictionaries, move them into `const.py`.

mineo: I'm having trouble understanding this. Since `index` is at most `len(string) - 1`, how can this condition be true?

mineo: You're allowed to just import `picard.util` ;-)

mineo, line 1606: It's not clear to me why replacing hyphens and quotes happens in a function that supposedly removes honorifics from names.

mineo: If `multi_list` is empty, this will not work.

mineo: Why doesn't this need all the other replacements from https://github.com/metabrainz/picard-plugins/pull/175/files#diff-c5c533fa72e35da36e7cf8a8b60c7839R3528 as well?

mineo: The two branches here look very similar. Wouldn't it make more sense to just initialize `tm[tag]` if it's not set and then just use one code path?

mineo, line 2700: This appears to be nearly the same as `picard.ui.searchdialog.to_seconds`.

mineo, line 2715: I do not understand this sentence :(

mineo, line 2737: This can also return a list of years. Maybe it would be more useful if this always returned a list, even if it just contained one element?

mineo: Dumping and loading immediately afterwards looks very weird.

mineo: Move to `const.py`. Likewise for `insertions`.

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,9857 @@
...
+    """
+    Note that, on initial load, new_metadata == orig_metadata; but, after refresh, new_metadata will have
+    the same track metadata as tm (plus the file metadata as per orig_metadata), so a trackid match
+    is then possible for files that do not have musicbrainz_trackid in orig_metadata. That is why
+    new_metadata is used in the above test, rather than orig_metadata, but orig_metadata is then used below
+    to get the saved options.
+    """
+
+    # Find the tag with the options:-
+    if track_file:
```
mineo, line 747: It would be nice if you could convertthese

```python
if condition:
    huge chunk of code
```

or

```python
for element in somelist:
    huge chunk of code
```

like in https://github.com/metabrainz/picard-plugins/pull/175/files#diff-c5c533fa72e35da36e7cf8a8b60c7839R2054 type of code blocks into

```python
if condition:
    call-some-function
```

or

```python
for element in somelist:
    call-some-function-for-element
```

blocks. That would reading the code and keeping track of what happens where much easier.

mineo (dismissed):

plugins/classical_extras/__init__.py
MetaTunes: Not quite sure what "this" refers to in the comment. genre_options is a list of dictionaries.

MetaTunes: This comment appears to refer to different code from that quoted. Which line of code does "this comment" refer to?

plugins/classical_extras/__init__.py
MetaTunes: That's a fair comment. There is probably quite a lot of legacy logging that could be usefully pruned. The excessive lines, however were created by autopep8 putting each parameter on a separate line - it didn't look that bad when I wrote it :)

plugins/classical_extras/__init__.py
MetaTunes: Thanks - I'll look into that as part of a general review of the logging. The per-release logging is mostly just for debugging (i.e. developer use) but can also be useful to explain why a particular outcome has occurred (i.e. an intended consequence in a complex situation).

plugins/classical_extras/__init__.py
MetaTunes: will fix

plugins/classical_extras/__init__.py
MetaTunes: Does this comment refer to lines 1611-1614 or something else?

MetaTunes: What is "this" referring to in "If this is maps"? The quoted lines (1611-1614) are part of a function which builds a list of dictionaries for loading and saving options in ClassicalExtrasOptionsPage. This enables options to be simply added in a declarative fashion. "Boolean" is string which is just used to call the correct method - self.ui.__dict__[ui_name].setChecked(...) - on lines 9755-6.

plugins/classical_extras/__init__.py
MetaTunes: I guess this comment might apply to other dictionaries as well. I would like to make it all a bit more OO, but perhaps as a subsequent exercise?

plugins/classical_extras/__init__.py
MetaTunes: Not sure I follow - the quoted code does seem to be logging the error.

MetaTunes:
Hi @mineo . Thanks for all the comments - I'm working through them. However, many of them just highlight the same 4 lines of code 1611-1614 while apparently referring to something else and I'm struggling to decipher them. Any suggestions?

plugins/classical_extras/__init__.py
MetaTunes: Will do

plugins/classical_extras/__init__.py
mineo: It doesn't log the actual `TypeError` object anywhere, though. So even if you look at the logs, you'd need to manually recreate the error to get to the bottom of it.

mineo (commented):
@MetaTunes If you click on the "Files changed" tab, the comments do show up at the correct place.

plugins/classical_extras/__init__.py
MetaTunes: You are correct. I'm still learning Python (if that were not apparent). I don't think that worked in 2.x, but should do so now in 3.x, so I'll give it a go.

plugins/classical_extras/__init__.py
MetaTunes: Will do

MetaTunes: Will do

MetaTunes: I agree! I borrowed this code from elsewhere and obviously didn't think about it too deeply. I think that whole condition can come out.

plugins/classical_extras/__init__.py
MetaTunes: Having a bit of a problem with this. Refactoring the signature seems to result in a hang (infinite loop?). It would be nicer not to have to specify dummy args each time, but I'll leave it for now and fix the other issues.

plugins/classical_extras/__init__.py
MetaTunes: Of course! I think this dates from the early versions when I was not sure it was wise to use functions that weren't part of the API. Since when, I realised that there is no API per se. Hopefully no-one will change this function and break my code.

MetaTunes: Nor me now ;) This dates way back to early versions. I'll try removing it and see if anything breaks.

MetaTunes: Amended comment: "Works with a list of lists or a list of strings"

MetaTunes: Ah. Previously only trapped None - not empty list. Changed to 
if list_len < 2:
        if list_len == 1:
            return multi_list[0]  # Nothing to do!
        else:
            return []

MetaTunes: renaming to lcs_dict

MetaTunes: It would, but it's not that simple. Setting tm[tag] = [] doesn't work as the metadata object does not keep it. Instead I'll use a flag to say that a new tag is required and hopefully simplify the code quite a bit.

MetaTunes: It does - an oversight. I'll put them all in one place. In the long run, I'll look for a better way of handling these.

MetaTunes: I didn't know about that function. Mine is almost identical but has a small error check built in (which might be unnecessary). I am still doubtful about freely using any function in Picard other than the ones that are essential. It seems to me that there is no API in the normal sense and that, effectively, the whole Picard code is open to use. On the one hand, this is great, but excessive use might lead to the risk of changes to Picard affecting the plugin, because the maintainer does not realise that the function is used by plugin code.

MetaTunes: Composer last names can be (optionally) used to prefix the album name. This places the last names in a sequence which is sorted in descending order of the total track times attributable to that composer. So, for multi-composer albums, the most "important" composer gets listed first.

MetaTunes: Will do

MetaTunes: I'll add blank lines for clarity

MetaTunes: It is a collection of booleans, viz {track: boolean, track : boolean, etc.}

MetaTunes: I know, but it was the only way I could get the dictionary to be saved to a tag correctly.

MetaTunes: Will do

MetaTunes: I know, but it was the only way I could get the dictionary to be saved correctly in the tag.

MetaTunes: Will do

plugins/classical_extras/__init__.py
MetaTunes: Adding the error object to the log

plugins/classical_extras/__init__.py
MetaTunes: It does break :( This function also standardises hyphens and apostrophes in names to enable comparisons. Will update the comment.

plugins/classical_extras/__init__.py
MetaTunes: I've looked into this and the standard logging doesn't seem to handle some unicode chars (eg '\u0151') very well. Since what I have written works and isn't (I hope) too inelegant, I'd prefer to leave it as is.

MetaTunes:
@mineo - I've dealt with all the comments so far (I hope) and revised code. Two matters are not fully dealt with:
Logging: The standard logging gave me errors, but I have removed all the conditional elements as these were no longer necessary.
parse_data: Although your suggestion should work in Python 3, I couldn't find a solution to the infinite loop that seemed to be created (possibly something to do with passing mutable lists as arguments?). The additional dummy arg in the calls is not too much of a hassle.

plugins/classical_extras/__init__.py
MetaTunes: I've done that for a number of the "worst offenders". I've left a few long bits where I think the readability is OK and a new function would not improve it.

MetaTunes:
I see Codacy doesn't like it. I'll re-run autopep8 etc.

phw:
Codacy has some valid complaints at https://app.codacy.com/app/MetaBrainz/picard-plugins/pullRequest?prid=2471883 , most of which should be easy to fix.

MetaTunes:
Just looking at them. Funnily enough, most of the issues are not new, but Codacy didn't reject the first PR. I'll fix what I can.

MetaTunes:
I really don't understand Codacy. All the remaining issues plus loads more were there at the first PR when it didn't complain. Now it wants these methods turned into functions. I really don't want to do that as it helps for them to sit next to the relevant calling code.

MetaTunes:
Aha - realised that removing conditionals for logging also removed class references - hence the Codacy issues were new. Fixed by using @staticmethod decorators.
Should be OK for follow-up on review now.

MetaTunes:
Hi @mineo - did you have any further comments on this or is it ready for merge?

plugins/classical_extras/__init__.py
mineo: Then I don't understand why this is a `collections.defaultdict(dict)`. That is for a `{track: dict, track: dict}`. Wouldn't `collections.defaultdict(lambda: False)` (the lyricist isn't filled by default) be more appropriate?

MetaTunes:
OK, so where do we go from here? The previous (Picard 1.4) version was merged OK. This one is a big improvement, and has been used by a number of people who downloaded it from my GitHub. Any reported bugs have been dealt with and it passes the checks. I appreciate that it is a big plugin, and not easy to review because of the size and complexity. I was [specifically asked](https://community.metabrainz.org/t/classical-extras-plugin-for-picard-1-4-x/300217/165) to make it available as a default plugin for 2.0.x and therefore put some work into tidying it up for release. I have dealt with all the comments raised so far. Given the track record, if no-one wants to review it any further, could it just be merged?

plugins/classical_extras/__init__.py
```diff
@@ -0,0 +1,8382 @@
...
+# FILE READING AND OBJECT PARSING
+
+_node_name_re = re.compile('[^a-zA-Z0-9]')
+
+
+def _node_name(n):
+    return _node_name_re.sub('_', str(n))
+
+
+def _read_xml(stream):
```
phw, line 360: This seems to be just a copy of `picard.util.xml.parse_xml`. Any specific reason `parse_xml` was not used here?

phw (approved):
I am actually in favor of merging this, especially as this plugin is pretty popular and already in use by many people.

I wanted to review this earlier, but this plugin is quite a beast. To be honest I have done no proper review of the functionality, just skimmed over it for common code smells. 

There are still a few valid concerns by @mineo . Also for me it is not quite clear why `_read_xml` is needed over Picard's `parse_xml`. Personally my biggest concern is the custom logging, this is really taking a lot of the code here. I still don't quite get what is wrong with Picard's logging, so if there is some thing that should be fixed there we maybe we should do it. But given the huge amount of different conditions and use cases this plugin has I can kind of see why one want to have a separate log file just for this. So I'm personally fine with merging it that way.

I'm sure we can get future updates to this plugin get merged much faster.

@MetaTunes Also I just wanted to let you know that we are aware that the current central plugin repository is not ideal, especially not for bigger plugins like yours. @zas and myself had already some chatting about this, and we both agreed that it would be best to replace it with a more decentralized approach, where each plugin can live in it's own repository. But no concrete plans so far.

Sophist-UK:
> @MetaTunes Also I just wanted to let you know that we are aware that the current central plugin repository is not ideal, especially not for bigger plugins like yours. @zas and myself had already some chatting about this, and we both agreed that it would be best to replace it with a more decentralized approach, where each plugin can live in it's own repository. But no concrete plans so far.

I have been thinking about this for a sort-of-plugins approach for an entirely different non-music tool. And the conclusion I came to for that environment (which is different to Picard) was that:

a. Some plugins might just provide code to be used by other catalogue or user written plugins - to ensure stability and backwards compatibility they needed to be in a centrally controlled repository with a standard set of approvers. To avoid the complications of maintaining approver lists on large numbers of repos, it probably makes more sense to have them in a single central repo.

b. If you want to have separate approver lists for individual plugins, then they will need to be in separate repos. It may / may not be a good idea to allow plugin authors to approve their own changes which might be backwardly incompatible or of poor quality. 

c. If you want to include in the catalogue plugins entirely under the control of the author and held in the authors repo, then to prevent reputational damage to the (Picard) brand, I think it has to be abundantly clear that these are third party plugins not endorsed by Picard.

MetaTunes:
Thanks @phw - IIRC the remaining concerns from @mineo were regarding the logging and the signature of the parse_data function. I'll keep these on a todo list, but for reasons explained previously it wasn't practical to make the changes just now. Re the custom logging generally, my problem with the standard Picard logging (in 1.4) was that it kept crashing the app. Also, if you load a lot of releases it is a nightmare to find the problem if it is in just one release. The custom logging overcomes both problems. It also enables me to provide support to users - debugging any semantic issues just requires them to send me the log file for that release (if I can't recreate it).

rdswift:
@MetaTunes / @Sophist-UK / @phw / @mineo / @zas / @samj1912 

In order to facilitate the discussion and start to document the thoughts and decisions of the participants regarding the handling of Picard third-party plugins in general, I've started a Google Doc at [https://docs.google.com/document/d/18Bx5IpKfERnO3Ag529FNiRc_TnU9pLdq7-japRdV3iM/edit?usp=sharing](https://docs.google.com/document/d/18Bx5IpKfERnO3Ag529FNiRc_TnU9pLdq7-japRdV3iM/edit?usp=sharing).  Please feel free to add your comments, suggestions and changes.

Sophist-UK:
Thanks Bob. I have added two suggestions to this document.

P.S. We should take all discussion elsewhere as these comments are intended for the PR itself and for the specific plugin which this PR targets.

rdswift:
@Sophist-UK thanks for the input.  I've also created a new ticket [PICARD-1404](https://tickets.metabrainz.org/browse/PICARD-1404) to capture discussion and moved the working document to a [wiki page](https://wiki.musicbrainz.org/User:Rdswift/PicardPluginProposal) for easier editing by anyone interested.  My apologies to all for hijacking this PR.

plugins/classical_extras/__init__.py
MetaTunes: I can't remember where I got _read_xml from, but I don't think it was from picard.util.xml.parse_xml, which I wasn't aware of. I suspect it is similar code from a common source. As I think I have implied in other responses above, the Picard API is a pretty scanty (or overwhelming) thing in that only a bit is documented as an API per se but a lot is open to usage (or abusage). I haven't tried using the Picard function yet, but I'll note it for future. However, as I said earlier, there is a risk from a plugin developer's point of view in just re-using Picard code which isn't in an official documented API.

MetaTunes:
Is this going to be merged or not? Not sure who does this. @zas ?
In any case I have some very small (but important) corrections to make to the help references in the UI and the PLUGIN_DESCRIPTION. Given the pain involved in passing all the checks so far, I thought it would be better to get this PR accepted then make those minor changes as a subsequent PR. Please advise.

phw:
@MetaTunes Sorry for the delay, I wanted to get back to this earlier. I'm going to merge this, but we still need @zas's help to update the website.

I agree having this merged will make it easier to get changes in and reviewed. Also I consider this plugin important and really want to see it listed.

MetaTunes:
Thanks. Unfortunately, one of the things that needs changing is the link to the readme in the PLUGIN_DESCRIPTION that will appear on the listing. I'll do this and other small items and then do new (hopefully much easier!) PR.

PR #176, opened 2018-11-01
addrelease: Handle disc number tags with totaldisc information

mineo:
This also removes some Python 2 compatibility code in `test.py` and adds doctests for the disc number extraction.

I don't feel strongly about using doctests, but it seemed the easiest solution without always importing unittest, even if just importing the plugin in Picard.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,4 @@ python:
   - "3.5"
   - "3.6"
   - "3.7"
-before_install:
-  - pip3 install picard
 script: python test.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 dist: xenial
 language: python
+cache:
+  pip: true
 python:
   - "3.5"
   - "3.6"
   - "3.7"
+before_install:
+  - pip3 install picard
 script: python test.py
--- a/plugins/addrelease/addrelease.py
+++ b/plugins/addrelease/addrelease.py
@@ -171,11 +171,10 @@ class AddClusterAsRelease(AddObjectAsEntity):
                 # disc numbers need to be changed to accommodate that.
                 self.discnumber_shift = max(self.discnumber_shift, 0 - m)
             m = m + self.discnumber_shift
-        except Exception as e:
+        except ValueError as e:
             # The most likely reason for an exception at this point is a
-            # ValueError because the disc number in the tags was not a
-            # number. Just log the exception and assume the medium number
-            # is 0.
+            # because the disc number in the tags was not a number. Just log
+            # the exception and assume the medium number is 0.
             log.info("Trying to get the disc number of %s caused the following error: %s; assuming 0",
                      metadata["~filename"], e)
             m = 0
--- a/plugins/addrelease/addrelease.py
+++ b/plugins/addrelease/addrelease.py
@@ -172,9 +172,9 @@ class AddClusterAsRelease(AddObjectAsEntity):
                 self.discnumber_shift = max(self.discnumber_shift, 0 - m)
             m = m + self.discnumber_shift
         except ValueError as e:
-            # The most likely reason for an exception at this point is a
-            # because the disc number in the tags was not a number. Just log
-            # the exception and assume the medium number is 0.
+            # The most likely reason for an exception at this point is because
+            # the disc number in the tags was not a number. Just log the
+            # exception and assume the medium number is 0.
             log.info("Trying to get the disc number of %s caused the following error: %s; assuming 0",
                      metadata["~filename"], e)
             m = 0
--- a/plugins/addrelease/addrelease.py
+++ b/plugins/addrelease/addrelease.py
@@ -6,7 +6,7 @@ PLUGIN_DESCRIPTION = "Adds a plugin context menu option to clusters and single\
  files to help you quickly add them as releases or standalone recordings to\
  the MusicBrainz database via the website by pre-populating artists,\
  track names and times."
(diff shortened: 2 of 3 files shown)


mineo:
Well, the travis test doesn't pass (yet) because the changes from https://github.com/metabrainz/picard/pull/998 haven't been released yet. But now we even have tests for our source distributions ;-) Anyways, this can wait a while.

phw:
@mineo The build errors now are related to https://github.com/metabrainz/picard/pull/998, right?

mineo:
I think so, yes.

plugins/addrelease/addrelease.py
```diff
@@ -116,38 +116,84 @@ class AddClusterAsRelease(AddObjectAsEntity):
...
+            discnumber = metadata.get("discnumber", "1")
+            # Split off any totaldiscs information
+            discnumber = discnumber.split("/", 1)[0]
+            m = int(discnumber)
+            if m <= 0:
+                # A disc number was smaller than or equal to 0 - all other
+                # disc numbers need to be changed to accommodate that.
+                self.discnumber_shift = max(self.discnumber_shift, 0 - m)
+            m = m + self.discnumber_shift
+        except Exception as e:
```
phw: Wouldn't it be better for check for `ValueError` here?

plugins/addrelease/addrelease.py
```diff
@@ -116,38 +116,84 @@ class AddClusterAsRelease(AddObjectAsEntity):
...
+        """
+        # As per https://musicbrainz.org/doc/Development/Release_Editor_Seeding#Tracklists_data
+        # the medium numbers ("m") must be starting with 0.
+        # Maybe the existing tags don't have disc numbers in them or
+        # they're starting with something smaller than or equal to 0, so try
+        # to produce a sane disc number.
+        try:
+            discnumber = metadata.get("discnumber", "1")
+            # Split off any totaldiscs information
+            discnumber = discnumber.split("/", 1)[0]
```
phw, line 167: `discnumber` could be `None`. I think it's better to check this here instead of relying on catch-all-exception handling

.travis.yml
```diff
@@ -1,6 +1,10 @@
 language: python
+cache:
+  pip: true
 python:
   - "3.4"
   - "3.5"
   - "3.6"
+before_install:
+  - pip3 install picard
```
phw: As this currently breaks the CI can we do without this for now?

plugins/addrelease/addrelease.py
```diff
@@ -116,38 +116,84 @@ class AddClusterAsRelease(AddObjectAsEntity):
...
+        >>> a = AddClusterAsRelease()
+        >>> m["discnumber"] = "-2/2"
+        >>> a.extract_discnumber(m)
+        0
+        >>> m["discnumber"] = "-1/4"
+        >>> a.extract_discnumber(m)
+        1
+        >>> m["discnumber"] = "1/4"
+        >>> a.extract_discnumber(m)
+        3
```
phw, line 156: I don't get this case. Why should `a.extract_discnumber("1/4") == 3` be true? Wouldn't it be `0` (first disc)?

phw (changes requested):
I agree that doctests is a reasonable and easy solution for having tests for the plugins, especially for the simple cases most plugins are.

Apart from some minor details it's mostly the failing build that is a problem. Maybe we should disable this for now until `pip install picard` works.

.travis.yml
mineo: Sure.

plugins/addrelease/addrelease.py
mineo: No. The important part is that `a` has not been reset or reassigned after extracting a disc number of -2, so disc 1 is in fact disc 4 (-2, -1, 0, 1), but medium numbers start with 0, so 3. This behaviour was originally added in https://github.com/metabrainz/picard/commit/d38104821654198f5b2a2f512b2d37a4b7b6ff63 as a fix for https://tickets.metabrainz.org/browse/PICARD-407.

plugins/addrelease/addrelease.py
mineo: Yes.

plugins/addrelease/addrelease.py
mineo: I don't see how `discnumber` could be None (or even an empty string). The default value passed to metadata.get is `"1"`, which should be returned in case of any falsy or None value in the tag.

phw: Right, I missed the default value in `get`

plugins/addrelease/addrelease.py
phw: Ok, got it

plugins/addrelease/addrelease.py
```diff
@@ -171,11 +171,10 @@ def extract_discnumber(self, metadata):
                 # disc numbers need to be changed to accommodate that.
                 self.discnumber_shift = max(self.discnumber_shift, 0 - m)
             m = m + self.discnumber_shift
-        except Exception as e:
+        except ValueError as e:
             # The most likely reason for an exception at this point is a
```
phw: I think with the updated text the "a" at the end of this line is too much ("...is because the disc number in  the tags was not a number").

phw (approved):
Apart from a small grammatical correction in the comment I think this is good to merge.

We can re-enable the doctests once `pip install picard` works, which hopefully it will with the next release and https://github.com/metabrainz/picard/pull/998 included.

PR #178, opened 2018-11-05
Do test on Python 3.7

Freso:
Picard supports Python 3.7, so making sure plugins work on it is very useful.

Heavily inspired by https://github.com/metabrainz/picard-plugins/pull/177 ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: python
 python:
   - "3.5"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,5 @@ language: python
 python:
   - "3.5"
   - "3.6"
+  - "3.7"
 script: python test.py


Freso:
Ugh. Looks like the Travisbuntu is too ancient to have Py3.7? I'll try and look into this later.

phw:
Unfortunately it won't work that easily, unless they have Python 3.7 finally available on Travis CI :) See https://github.com/metabrainz/picard/commit/c0448cbbf983d7af64421843eecc78d383633f03 for a working solution if the builds will fail

EDIT: Yep, failed :(

Freso:
Instead of making an additional `matrix` entry, I just changed the distribution used to be Ubuntu 16.04 instead of 14.04.

phw (approved):

zas (approved):

PR #181, opened 2018-11-12
Update plugins/wikidata/wikidata.py

pbrackin:
Refer to 

https://tickets.metabrainz.org/browse/PICARD-1411 
https://community.metabrainz.org/t/picard-keeps-locking-up/358547/4

I was not liking some of the esoteric genres that came from last.fm, so I decided to try wikidata & while it worked for some cases, I found it was not always stable.  It seems like this is pretty well documented in the JIRA DB + community forums.

The most glaring issue was this recursion that happened when you right-click the "other versions" option to switch releases.  It would eventually call to finalize the album which would call run_track_metadata_processors which would call to process the track then to finalize again, and then...

So, I spent some time carefully stepping through & trying to understand the plugin flow & data structures that are in place.  I was able to deduce the one problematic call to finalize there was not needed & thus eliminate the recursion.  There are some locks they put in which seems to acknowledge a re-entrant issue, but perhaps they didn't see through to the recursion.  But as best I can tell, no other threads ever come in so I am not sure the locks are needed once the recursion problem is lifted.  I did some basic testing with both the recursion & the locks out & everything seemed to work ok for me.  But I left the locks in for now, in case there are some edge cases I am not seeing thus far.  Locks do represent a certain amount of overhead tho, so some consideration might be given to their removal.

Besides the recursion, I identified extra album insertions that may have been causing some albums to end up un-finalized.  Removing the extra insertions seems to have eliminated these dangling albums.

Also, I moved the server & port to instance variables to avoid hundred of unneeded config lookups.

The above items in combination seemed to also speed the overall process up a bit.  Nothing super-major, but noticeable. 

Lastly, I sorted the genre list because if left unsorted, the sets can come out in different order for every lookup, causing "false positives" for the genre field (looks like a change when, in fact,  just the order is changing).

I ran the unittests under the test folder & it said 68 tests OK.  Not sure if you had other tests or what, since this is my first PR on this project.

### DIFF
--- a/plugins/wikidata/wikidata.py
+++ b/plugins/wikidata/wikidata.py
@@ -33,23 +33,27 @@ class wikidata:
         # key: mbid, value: list of strings containing the genre's
         self.cache = {}
 
-        # find the mb url if this exists
-        self.mb_host = config.setting["server_host"]
-        self.mb_port = config.setting["server_port"]
+        # metabrainz url
+        self.mb_host = ''
+        self.mb_port = ''
+
+        # web service & logger
+        self.ws = None
+        self.log = None
 
     # not used
     def process_release(self, album, metadata, release):
         self.ws = album.tagger.webservice
         self.log = album.log
         item_id = dict.get(metadata, 'musicbrainz_releasegroupid')[0]
-        
+
         log.info('WIKIDATA: processing release group %s ' % item_id)
         self.process_request(metadata, album, item_id, type='release-group')
         for artist in dict.get(metadata, 'musicbrainz_albumartistid'):
             item_id = artist
             log.info('WIKIDATA: processing release artist %s' % item_id)
             self.process_request(metadata, album, item_id, type='artist')
-            
+
     # Main processing function
     # First see if we have already found what we need in the cache, finalize loading
     # Next see if we are already looking for the item
@@ -183,7 +187,7 @@ class wikidata:
                 for metadata in self.requests[item_id]:
                     new_genre = set(metadata.getall("genre"))
                     new_genre.update(genre_list)
-                    #sort the new genre list so that they don't appear as new entries (not a change) next time
+                    # sort the new genre list so that they don't appear as new entries (not a change) next time
                     metadata["genre"] = sorted(new_genre)
                     self.cache[item_id] = genre_list
                     log.debug('WIKIDATA: setting genre : %s ' % genre_list)
@@ -191,7 +195,7 @@ class wikidata:
             else:
                 log.info('WIKIDATA: Genre not found in wikidata')
 
-            log.info('WIKIDATA: Seeing if we can finalize tags %s  ' % len(self.albums[item_id]))
+            log.info('WIKIDATA: Seeing if we can finalize tags %d  ' % len(self.albums[item_id]))
 
             for album in self.albums[item_id]:
                 album._requests -= 1
@@ -201,6 +205,8 @@ class wikidata:
                 log.info('WIKIDATA:  TOTAL REMAINING REQUESTS %s' % album._requests)
 
     def process_track(self, album, metadata, trackXmlNode, releaseXmlNode):
+        self.mb_host = config.setting["server_host"]
+        self.mb_port = config.setting["server_port"]
         self.ws = album.tagger.webservice
         self.log = album.log
 
--- a/plugins/wikidata/wikidata.py
+++ b/plugins/wikidata/wikidata.py
@@ -8,14 +8,12 @@
 PLUGIN_NAME = 'wikidata-genre'
 PLUGIN_AUTHOR = 'Daniel Sobey, Sambhav Kothari'
 PLUGIN_DESCRIPTION = 'query wikidata to get genre tags'
-PLUGIN_VERSION = '1.1'
+PLUGIN_VERSION = '1.2'
 PLUGIN_API_VERSIONS = ["2.0"]
 PLUGIN_LICENSE = 'WTFPL'
 PLUGIN_LICENSE_URL = 'http://www.wtfpl.net/'
 
 from functools import partial
-import threading
-
 from picard import config, log
 from picard.metadata import register_track_metadata_processor
 
@@ -23,12 +21,11 @@ from picard.metadata import register_track_metadata_processor
 class Wikidata:
 
     def __init__(self):
-        self.lock = threading.Lock()
         # Key: mbid, value: List of metadata entries to be updated when we have parsed everything
         self.requests = {}
 
         # Key: mbid, value: List of items to track the number of outstanding requests
-        self.albums = {}
+        self.itemAlbums = {}
 
         # cache, items that have been found
         # key: mbid, value: list of strings containing the genre's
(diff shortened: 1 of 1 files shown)


plugins/wikidata/wikidata.py
```diff
@@ -32,7 +32,11 @@ def __init__(self):
         # cache, items that have been found
         # key: mbid, value: list of strings containing the genre's
         self.cache = {}
-        
+
+        # find the mb url if this exists
+        self.mb_host = config.setting["server_host"]
+        self.mb_port = config.setting["server_port"]
```
phw: In this case we should keep reading the values from `config.setting` in `process_request`. If this is only read in the constructor runtime changes to the settings will have no affect until Picard gets restarted.

plugins/wikidata/wikidata.py
```diff
@@ -129,8 +125,7 @@ def musicbrainz_release_lookup(self, item_id, metadata, response, reply, error):
         with self.lock:
             for album in self.albums[item_id]:
```
phw: This looks strange to me, why does it iterate over `self.albums` here? `musicbrainz_release_lookup` is only called once per album request, I think only for that album the requests should be lowered.

phw (changes requested):
Your changes make sense, but I think this does not go far enough. So the feedback below is not really about your code, but the existing code.

As I read the code there is a race condition between `process_wikidata` and `parse_wikidata_response`. In `process_wikidata` the requests for all albums are increased:

```
for album in self.albums[item_id]:
    album._requests += 1
```

In `parse_wikidata_response` the counts get removed with:

```
for album in self.albums[item_id]:
    album._requests -= 1
```

Problem here is the list of albums when `process_wikidata` gets called is not necessarily the same as in `parse_wikidata_response`. I also don't get why the counts for all albums need to get increased. Wouldn't it be sufficient to do this for the currently processed album only?

And if there is a reason to do it for all in the list a copy of this list should probably be passed to `parse_wikidata_response` so that it decreases the calls only for those albums that had been increased previously.

pbrackin:
OK.  Well, in general, I agree that the design is somewhat flawed.  I came across many of the same questions as you - only without the benefit of your in-depth understanding of Picard's guts :P  At the very least, the current design can benefit from some more changes - some of which I discussed in the PR.  At the most, possibly a complete re-do with only a recycling of the parse code could be in order.   

However, my efforts here were mainly targeted towards "just get it working".   Not sure where you're at on your release cycle, but it may be prudent to get the current changes into the next release and then allow some time for discussion on the design & subsequent associated implementation.  It may be helpful to hear from the original authors, if they are still around, on their rationale on the existing design and as well, they may want to make some changes too.  

Of course... holding back the changes in this PR on all that may not be prudent as the plugin is pretty broken in the current state & with my current PR changes, it does work much better.

Having said that.. if you feel strongly to hold off til a more robust solution is in place and we don't hear from the original authors - I'd be happy to dig in and revamp it.  Of course, I'd need some time.

I will go ahead and provide some feedback on your feedback in the mean time :O)

plugins/wikidata/wikidata.py
pbrackin: Hmmm....  How often do you change the server/port?   Has it EVER changed?  Do people really leave Picard running for very long normally?  Is a restart of Picard a big problem for the rare case when such a change occurs?   Not saying the config lookup is a huge overhead, but doing config lookups for every request is just poor design (very non-DRY) & it does add up over hundreds / thousands of requests.  You are doing a lot of work to cover a very far removed edge case.  A better design is to have a publisher/subscriber model where plugins get notified via callback of config changes.  A compromised approach of lookup on error could be effective if you really felt its a common enough scenario.

plugins/wikidata/wikidata.py
pbrackin: You're asking the wrong guy :)  I can only speculate that they are doing this because that had too many copies - possibly from the code I removed in process_release?   If it confuses you, imagine how a noob like me felt!  Again tho.. this is an area ripe for deisgn mod, IMHO.

plugins/wikidata/wikidata.py
phw: People do change this value. Not often, but they do. And the expectation is very much that the changed option takes effect immediately. The only option where we unfortunately require a restart is changing the UI language, and we inform the user about it. If any other existing option does not take effect after setting it this is considered a bug.

I agree that a pub/sub model would be better for some use cases, but since there currently is none we have to work with fetching the value when needed. This change here does not make the code more DRY, it introduces a bug. So please revert this part.

phw:
Since plugin changes can be published independently of Picard we are not bound to any release schedule here. I'm happy to merge this as a step forward if the config issue gets resolved.

But I still think this needs some more cleanup before we can consider it fixed.

plugins/wikidata/wikidata.py
```diff
@@ -180,24 +175,23 @@ def parse_wikidata_response(self, item, item_id, response, reply, error):
...
+                    metadata["genre"] = sorted(new_genre)
                     self.cache[item_id] = genre_list
                     log.debug('WIKIDATA: setting genre : %s ' % genre_list)
 
             else:
                 log.info('WIKIDATA: Genre not found in wikidata')
 
-            log.info('WIKIDATA: Seeing if we can finalize tags %s  ' %
-                     len(self.albums[item_id]))
+            log.info('WIKIDATA: Seeing if we can finalize tags %s  ' % len(self.albums[item_id]))
```
zas: `%d` format string place holder makes more sense with `len()`

plugins/wikidata/wikidata.py
pbrackin: Actually it does make the code more DRY - *MUCH* more DRY.  However, I understand your desire & the limitations of the framework you have built thus far.  What I will do is meet you in the middle on this and put the lookup in process_track, as that is called far fewer times than process_request.

plugins/wikidata/wikidata.py
pbrackin: I didn't actually change this line other than making it one-line, but I can throw this is the next commit for you :)

pbrackin:
Good call.  I will try to help you out further & rework it some more if we don't hear from the authors on any of this.

One more thing... you have a design flaw with all the calls to the protected album member _finalize_loading.  This can be dangerous.  You should consider a wrapper for this.  You could use the wrapper for additional checks and logic like checking for recursions to avert plugins sinking the whole ship, as we saw here when using the 'Other releases" context item.

phw:
> you have a design flaw with all the calls to the protected album member _finalize_loading. This can be dangerous. You should consider a wrapper for this.

See the existing ticket at https://tickets.metabrainz.org/browse/PICARD-1403 . Suggestions and code welcome. Especially how to best handle that we ensure only one request decrement for each increment.

pbrackin:
OK good!

plugins/wikidata/wikidata.py
phw: Thanks, that's excellent

phw (approved):

plugins/wikidata/wikidata.py
```diff
@@ -75,27 +82,20 @@ def process_request(self, metadata, album, item_id, type):
                         'WIKIDATA: request already pending, add it to the list of items to update once this has been found')
                     self.requests[item_id].append(metadata)
 
-                    album._requests += 1
-                    self.albums[item_id].append(album)
```
phw, line 79: Looking at the code again, I don't think this should be removed.

The idea seems to be, that the same `item_id` (e.g. work) can be processed multiple times. The plugin will not do additional requests, but will remember the album to be updated later It will also increase the request count, so once the request has completed it can be updated and in the meantime shows as still loading data.

This functionality is broken when this gets removed. What probably should be changed is having `self.albums[item_id]` not as a list but a set, because it should contain each album only once.

phw (changes requested):

phw:
This plugin is somewhat tricky. I noticed another issue, which actually might be the root problem of some of the issues users have. To check if an album can be removed from the plugins list we have something like this in the code:

```
if not album._requests:
                self.albums[item_id].remove(album)
                album._finalize_loading(None)
```

But this will only ever work reliable if wikidata is the only thing dong requests. But this is seldom the case. So if wikidata finishes the requests, but other plugins or covert loaders are still busy, the album will stay in wikidata's list. Any attempt of later reloading this album will get stuck.

Having `_finalize_loading` already getting called in  `process_request` seems to have been an attempt to fix this. But the proper way is probably for the wikidata plugin to track only it's own requests for an album (in addition to manipulating the album request counter).

I did some debugging here, I will see if I can submit a fix for this directly to this pull request tomorrow.

pbrackin:
Not sure I fully understand the dependency of one plugin on other plugins that you are describing, or even the design intent of album._requests.  Seems like you would want to have it all managed in the album object or higher & just let the plugin mark an album if a change is needed, so the album would open and finalize as needed - managing requests internally & also eliminate any plugin inter-dependencies.  But it seems like the plugins have too much power with the ability to block the main thread, leave albums open, etc...  Not being critical, just saying this approach could perhaps be improved over time.  In the mean time, some more exhaustive testing of new plugins or plugin mods should be considered.

I have continued to test the current revision of wikidata (the one in this PR) against subsets of a large media library & its been working flawlessly for me.  Its truly a massive improvement over what's currently being made available to users.  I could test the reloading problem you are describing if you can explain the test procedure more clearly.  I am running a couple other plugins - the fanart.tv one, mainly.

I was going to start on the additional improvements we discussed this next week.  I was going to create a new JIRA ticket to track that.  But if you are going to make changes here maybe its better to wait on those...

phw:
My attempt of improving this: https://github.com/pbrackin/picard-plugins/pull/1

pbrackin:
I am still testing, but so far its been working good.  One thing I found though... if there is a network problem, a track may get an incomplete genre set from what it would otherwise get.  To the user this shows up like a change and the change is a reduced genre set.

The problem seems to be that the webservice does not send an error indication via the callback so wikidata is kind of in the dark that there was a problem.

Below is a snippet from the log.   Here again, I am not sure about the design intent, so maybe you can guide me in on if this is considered "normal" behavior for the plugin to not get an error indication. 

I noticed that if I simulate a complete network outage by driving the musicbrainz.org requests to null then I get the red album discs in the UI.  This seems goods so as to indicate to the user they need to retry later.  But for the case I am describing here, its not turning red.  

Probably can address this in a separate ticket, but wanted to bring it up since I see it in my testing on occasion.

```
E: 07:45:44,103 C:\Users\paul\Documents\GitHub\picard\picard\webservice\__init__._handle_reply:383: Network request error for https://musicbrainz.org:443/ws/2/release-group/467a5697-30e7-34b8-9813-d07e6af9c653?inc=url-rels: Error transferring https://musicbrainz.org:443/ws/2/release-group/467a5697-30e7-34b8-9813-d07e6af9c653?inc=url-rels - server replied: Service Temporarily Unavailable (QT code 403, HTTP code 200)
D: 07:45:44,105 C:\Users\paul\Documents\GitHub\picard\picard\webservice\__init__._handle_reply:395: Retrying https://musicbrainz.org:443/ws/2/release-group/467a5697-30e7-34b8-9813-d07e6af9c653?inc=url-rels (#2)
```

phw (approved):
I think this is looking good now. I really appreciate you did the cleanup here. I think it is better to have this clean with the fixes, otherwise the next one working on this just wonders what this is all about.

plugins/wikidata/wikidata.py
```diff
@@ -125,23 +122,24 @@ def musicbrainz_release_lookup(self, item_id, metadata, response, reply, error):
...
-                          album._requests)
-                if not album._requests:
-                    self.albums[item_id].remove(album)
-                    album._finalize_loading(None)
+            log.debug('WIKIDATA: No wikidata url found for item_id: %s ', item_id)
+
+        album = self.itemAlbums[item_id]
+        album._requests -= 1
+        if not album._requests:
+            self.itemAlbums = {k: v for k, v in self.itemAlbums.items() if v != album}
```
phw, line 130: I think we still have an issue here, since this line only ever gets executed when `album._requests` becomes 0, but actually this will not always be true if other plugins are used in parallel (only one plugin will be the one dealing with the last request, and this might not be wikidata). So we have the risk of not updating `itemAlbums` correctly.

Anyway I think we should merge this first and deal with this issue separately.

phw:
> The problem seems to be that the webservice does not send an error indication via the callback so wikidata is kind of in the dark that there was a problem.
> 
> Below is a snippet from the log. Here again, I am not sure about the design intent, so maybe you can guide me in on if this is considered "normal" behavior for the plugin to not get an error indication.

The problem here is that this plugin never passes any error to `album._finalize_loading()`. To actually show an error the plugin is supposed to call `album._finalize_loading(error)`. Note that currently the call to `_finalize_loading` is guarded by a check against `album._requests` being 0 (see also my comment above), which strictly speaking is not necessary since this is checked in `_finalize_loading` itself. That is especially true if passing an error, which must be done regardless the status of  `album._requests`.

Regarding the design intent you are asking the wrong person, until a few weeks ago I had never looked at the wikidata code in detail. We are kind of in the same boat here :)

LGTM!

plugins/wikidata/wikidata.py
phw: The only way to solve this probably is for this plugin to keep track of its own request count.

pbrackin:
But how can the plugin pass an error if it never gets an error back from the webservice to pass?  From what I could tell, the callback just never gets called for these cases.  I suppose you could add some sort of timer and make an error on timeout, but seems like the right solution is for the webservice to notify the plugin?

phw:
> But how can the plugin pass an error if it never gets an error back from the webservice to pass?

On a request error the callback will be called with the `error` object set, see the error handling with `if error`.

plugins/wikidata/wikidata.py
pbrackin: OK yeah I can do that too in a separate PR.  If you change the album request counter thing to an API / wrapper call, this change could be made at the same time, or if that is not coming soon we could go ahead and do this.

pbrackin:
Yeah that is what I thought.  That is not happening in this case.

phw:
> Yeah that is what I thought. That is not happening in this case.

It definitely is supposed to happen, and it does happen here on my tests. If there is any network error (I tested by changing the wikidata.org host to something invalid, but could also be an error response like 500 from the server) the error object will be set and the wikidata logging happens.

If that really does not work for you, can you give some details about the exact error condition?

plugins/wikidata/wikidata.py
phw: We shouldn't wait for any Picard API changes, because there are no concrete plans yet and the earliest release that could have some changes here could be Picard 2.2.

zas (approved):
I think we have enough positive changes in this PR to merge, it looks good, further improvements should be made in new PRs.

pbrackin:
OK when I get a chance I will open a new ticket with more data on the "lack of callback error" for the one case I was describing.  Thanks for y'alls help getting this PR in to make this plugin operational :O)

pbrackin:
One more question...  I saw the merge happened.  When & how does this new plugin version become available?  I saw the generate.py script in the plugin repo has to be run.  Does that get run by your build system? Or do I need to do something with my local Picard to see it?

phw:
Unfortunately not, the website needs to be redeployed manually and I was told this is a rather painful process. We talked about this today and would like to streamline the process, but it'll require some work.

zas:
>  When & how does this new plugin version become available?

New plugin versions were deployed just now. As explained by @phw , it's a tedious process, we have plans to improve this, but for now, it just sucks ;)

PR #185, opened 2018-11-27
Update GPL2/3 or later licenses

phw:
Clarify the license by setting it explicit to "GPL-2.0-or-later" or "GPL-3.0-or-later" where this applies. Also unify the spelling of these licenses by ussing the [SPDX name](https://spdx.org/licenses/)

This is not a license change, the affected plugins clearly specify the respective license in their headers. We were just sloppy with the `PLUGIN_LICENSE` value in the past and not as explicit as we should have been. There are also a couple of plugins explicitly having only GPL-2.0 in their headers (both acousticbrainz plugins), the rest does not specify it exactly.


zas:
@phw this PR includes commits from https://github.com/metabrainz/picard-plugins/pull/184 it seems

phw:
@zas I merged #184 and rebased against 2.0, now we have only the relevant changes here.

zas (approved):

PR #189, opened 2018-12-14
Add format_performer_tags plugin

rdswift:
Here is the current version of the **Format Performer Tags** plugin for consideration.  I understand that @phw wants to improve the UI (which I greatly appreciate).  I've also included the documentation for the plugin in this PR, because I wasn't sure how to handle it.  When you guys sort out the new plugin management system, links to external user guides might be something worth considering, although I'm not sure how that would work with translations.


phw (approved):
Thanks for this. I will submit my changes against this once it got merged

PR #190, opened 2018-12-14
Add UI changes by outsidecontext / phw

rdswift:
@phw I had your changes already merged into my development branch, so I thought it would be easier for me to create the PR to pull them into the main branch.  I hope that's okay.


phw:
Perfect, thanks. Much appreciated

rdswift:
Actually, I should be thanking you for the work on the UI examples and such.  That is a lot more user friendly when choosing settings.

Any thoughts regarding my TODO comment?

```
# TODO:  Include format placeholder in UI text for User Guide URL
#
# Modify self.format_description in ui_options_format_performer_tags.py to include a placeholder
# such as {user_guide_url} so that the translated string can be formatted to include the value
# of PLUGIN_USER_GUIDE_URL to dynamically set the link while not requiring retranslation if the
# link changes.  Preliminary code something like:
#
#         temp = (self.ui.format_description.text).format(user_guide_url=PLUGIN_USER_GUIDE_URL,)
#         self.ui.format_description.setText(temp)
```

phw:
> Any thoughts regarding my TODO comment?

Yes, I think we could do it similar to how you suggested in the comment. One thing to note here is that we currently lack the functions to properly translate plugins anyway. In the early days of Picard some plugins got translated, but only because they were shipped together with Picard and had their translations included in Picard's translation files. That's probably another topic we should address in any new plugin system.

rdswift:
Thanks.  For now I just noted it in the "Considerations" section of the [wiki page](https://wiki.musicbrainz.org/User:Rdswift/PicardPluginProposal) discussiing proposed changes to the plugin management system.

rdswift:
Obviously I did something wrong, because I didn't mean to close this PR.  I'll try to re-open it, or re-create it.  I'm sure I still have the updated information here.

PR #193, opened 2018-12-17
Add artist_variables plugin

rdswift:
Add new plugin to (eventually) replace the ***album_artist_extension*** plugin, as discussed in the [community forum](https://community.metabrainz.org/t/artist-variables-plugin/409544/2).  This new plugin provides the same functionality (and more), but the new tag variable names have changed.

### DIFF
--- a/plugins/artist_variables/__init__.py
+++ b/plugins/artist_variables/__init__.py
@@ -1,0 +1,189 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018 Bob Swift (rdswift)
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.
+
+PLUGIN_NAME = 'Artist Variables'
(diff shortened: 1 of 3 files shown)


Sophist-UK:
1. Is there really a need for this? Could you explain how you think that this data might be used?

2. Is there a reason to create a new plugin rather than extending the existing albumartistextension plugin in a backwardly compatible way?

rdswift:
1.  I actually use the data in my renaming scripts to set directories based on the standardized name of the primary album artist, then add the additional artist names into the track filenames as appropriate.  I also use it to do some unusual things if the credited name is different from the standard name.

2.  I suppose that I could extend the existing albumartistextension plugin as you described, but the new functionality also includes the track artists so the name would no longer be appropriate.  An alternative would be to create a similar trackartistextension plugin, but that would be a virtual duplicate including much of the same code.  Perhaps the new "artist_variables" plugin could be modified to include backwards compatibility with the existing albumartistextension plugin, while still retaining all the additional functionality.  That way it could be a drop-in replacement for the albumartistextension plugin and not require any modifications to the users' scripts.

rdswift:
I'm withdrawing this submission until the changes to the `Format Performer Tags` are sorted out.

Sophist-UK:
I see what you mean about the plugin naming. Perhaps a new plugin is a better solution.

A few naming things;

1. Are these variables about "artist" (singular) or "artists" (plural)? My gut says "artists" - so perhaps it should be called "Additional artists variables" or "More artist variables".

2. I would personally prefer to see the variable names start with "_artist_" or "_artists_" so that it was clear which plugin they were associated with. And to have a naming convention that grouped related variables more closely and which separated the name parts with _. So we would have _artists_album_pri_id, _artists_album_pri_std, _artists_album_pri_credit, _artists_album_pri_std_sort, _artists_album_pri_credit_sort etc.

Sophist-UK:
P.S. I see you have closed the PR - in the light of my comment above, please reopen it (or request an admin to reopen it) if you want to continue.

rdswift:
I closed it for now because it was screwing up another open PR to include some UI improvements by @phw  in the ***Format Performers*** plugin.  Once that has been merged and closed, I will re-open this one with some revisions.  Thanks for the suggestions about naming the variables differently.  It does make things a bit clearer and easier for the user to understand.

PR #194, opened 2018-12-18
Wikidata-genre UI implementation

pbrackin:
Adding UI for wikidata-genre plugin:
`wikidata.py -> __init__.py, `
adding QT UI files.  
bump version to 1.3.

Plus changes to handle the options (including static and dynamic changes) under various scenarios. 

One setting default I wasn't sure what to do on was the default delimiter.  I left the default delimiter to "; ", but was wondering if you wanted to change the default to the Kodi default of " / "?  Seems there is no standard, sadly, but I brought it up because I noted the relationship between Kodi and Musicbrainz.  But...


phw:
Thanks a lot for this, these options are really useful. I currently can't give this a full review (and I hope somebody else comes along for this ;) ), but can you address the small code style issues found by Codacy (https://app.codacy.com/app/MetaBrainz/picard-plugins/pullRequest?prid=2756241)?

From just quickly skimming over it and looking at the available options this looks good to me.

pbrackin:
I put the PR in too soon... We got some help testing plus I added another config so letting it cook some here is prolly best.  I think no rush on this one anyway.  One of the codacy bugs I was able to clean out, but the other was from the QT gen'd code.  I will have to look at see why its adding an un-required import.  I could change the gen'd code but usually like to avoid that approach.

phw:
We can ignore the Qt generated issue. In Picard itself we have some codacy config to ignore the generated files, we probably should add that rule to the plugins repo as well.

pbrackin:
I think this PR is ready now.  We had several users testing it out over the last few weeks & prior to the last commit I did which was mainly just cosmetic.   If there is anything I can do to help get it reviewed, let me know.  Also, with the success of this "wikidata options UI" feature, some suggestions came in for some additional features which might further the usefulness of this plugin, but I was thinking to address those in a separate ticket.  My work here is probably done for now as my free time resources are getting sucked up by other demands ATM...

plugins/wikidata/__init__.py
```diff
@@ -0,0 +1,397 @@
...
+        self.ui.setupUi(self)
+        if not config.setting["write_id3v23"]:
+            self.ui.genre_delimiter.setEnabled(False);
+            self.ui.genre_delimiter_label.setEnabled(False);
+        else:
+            self.ui.genre_delimiter.setEnabled(True);
+            self.ui.genre_delimiter_label.setEnabled(True);
+
+    def info(self):
+        pass
```
phw: Overriding info() should actually not be necessary, since no other plugin providing an options page does this.

The method actually exists in `OptionsPage`, but it's unused and should probably be removed. I currently can't remember for what we originally had this, but it's likely unused for a long time. I'll submit a PR to remove this from Picard, see https://tickets.metabrainz.org/browse/PICARD-1459

plugins/wikidata/__init__.py
```diff
@@ -0,0 +1,397 @@
...
+
+        if len(genre_list) > 0:
+            log.debug('WIKIDATA: item_id: %s' % item_id)
+            log.debug('WIKIDATA: Final list of wikidata id found: %s' % genre_entries)
+            log.debug('WIKIDATA: Final list of genre: %s' % genre_list)
+            log.info('WIKIDATA: Total items to update: %d ' % len(self.requests[item_id]))
+
+            for metadata in self.requests[item_id]:
+                if genre_source_type == Wikidata.RELEASE_GROUP:
+                    metadata['release_group_genre_sourced'] = True
```
phw: Is this intentionally a visible variable? This will cause it to be written to tags.

I think this would be better set as a hidden variable `metadata['~release_group_genre_sourced']`. That way it does not show up in file tags, but the user can still use it in scripting. They even can set it as a proper tag again if they wish to see it in their file tags.

phw (changes requested):
I think this is mostly ready for merging, just some minor points up for discussion.

plugins/wikidata/__init__.py
```diff
@@ -0,0 +1,397 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2016 Daniel sobey <dns@dns.id.au >
+
+# This work is free. You can redistribute it and/or modify it under the
+# terms of the Do What The Fuck You Want To Public License, Version 2,
+# as published by Sam Hocevar. See http://www.wtfpl.net/ for more details.
+
+PLUGIN_NAME = 'wikidata-genre'
```
phw: Side question: What do you think about changing the name to "Wikidata Genre"? The current naming really looks odd compated to all other plugins listed.

plugins/wikidata/__init__.py
phw: This method has now been removed from Picard in https://github.com/metabrainz/picard/pull/1090

pbrackin: OK good; I think this was just brought in as a pattern that existed in another addon.  I will remove this.

plugins/wikidata/__init__.py
pbrackin: Its been a little while, but I can't think of any reason why it would need to be visible.   I didn't know about the private-prefix tilde thingy tho.  I will make it private, I think it should be fine like that.

plugins/wikidata/__init__.py
pbrackin: Yeah its goofy.  When I first saw it, I wondered why they went with a lower-case "w".  But I eventually figured out that the actual wikidata project used a lower case "w".   The "-" is still a mystery to me tho. 

My vote is def FOR congruency in the plugin names, but here's some thoughts... Would changing PLUGIN_NAME break the upgrade path?  As in, peeps with existing wikidata-genre installs wouldn't see the update (somehow they'd have to know to delete the old and start a-fresh with the new name)?  Or worse probably, could they end up with 2 versions of the plugin trying to run at the same time?  Maybe your server has a way to suss that out, or create a special case?

phw: No, wouldn't break anything. This luckily is just the display name. The internal name which is used for updates is "wikidata" (the folder name or in case of a single file plugin the filename without extension)

phw (approved):

plugins/wikidata/__init__.py
phw: I went ahead and pushed a commit to change the name to "Wikidata Genre"

phw:
Merged, thanks a lot for these changes.

PR #195, opened 2018-12-23
Created Add Column album plugin

evandrocoan:
The same for version 1.4: https://github.com/metabrainz/picard-plugins/pull/138

Tested with version 2.1.0

### DIFF
--- a/plugins/add_album_column/__init__.py
+++ b/plugins/add_album_column/__init__.py
@@ -1,0 +1,33 @@
+# -*- coding: UTF-8 -*-
+
+#
+# Licensing
+#
+# Channel Manager Main, Create and maintain channel files
+# Copyright (C) 2017 Evandro Coan <https://github.com/evandrocoan>
+#
+#  This program is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU General Public License as published by the
+#  Free Software Foundation; either version 3 of the License, or ( at
+#  your option ) any later version.
+#
+#  This program is distributed in the hope that it will be useful, but
+#  WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+#  General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+PLUGIN_NAME = u"Add Album Column"
+PLUGIN_AUTHOR = u"Evandro Coan"
+PLUGIN_DESCRIPTION = "Add the Album column to the main window panel."
+
+PLUGIN_VERSION = "1.0"
+PLUGIN_API_VERSIONS = ["2.0"]
+PLUGIN_LICENSE = "GPLv3"
+PLUGIN_LICENSE_URL = "http://www.gnu.org/licenses/"
+
+from picard.ui.itemviews import MainPanel
+MainPanel.columns.append((N_('Album'), 'album'))
--- a/plugins/add_album_column/__init__.py
+++ b/plugins/add_album_column/__init__.py
@@ -3,8 +3,22 @@
 #
 # Licensing
 #
-# Channel Manager Main, Create and maintain channel files
-# Copyright (C) 2017 Evandro Coan <https://github.com/evandrocoan>
+# Add Album Column, Add the Album column to the main window panel
+# Copyright (C) 2019 Evandro Coan <https://github.com/evandrocoan>
+#
+#  Redistributions of source code must retain the above
+#  copyright notice, this list of conditions and the
+#  following disclaimer.
+#
+#  Redistributions in binary form must reproduce the above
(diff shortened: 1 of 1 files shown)


plugins/add_album_column/__init__.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: UTF-8 -*-
+
+#
+# Licensing
+#
+# Channel Manager Main, Create and maintain channel files
```
mineo: I find "Add Album Column, Add the Album column to the main window panel" (from https://github.com/metabrainz/picard-plugins/pull/138/files) to be significantly better :-)

plugins/add_album_column/__init__.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: UTF-8 -*-
+
+#
+# Licensing
+#
+# Channel Manager Main, Create and maintain channel files
+# Copyright (C) 2017 Evandro Coan <https://github.com/evandrocoan>
```
mineo: 2018?

plugins/add_album_column/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+PLUGIN_NAME = u"Add Album Column"
+PLUGIN_AUTHOR = u"Evandro Coan"
+PLUGIN_DESCRIPTION = "Add the Album column to the main window panel."
+
+PLUGIN_VERSION = "1.0"
+PLUGIN_API_VERSIONS = ["2.0"]
+PLUGIN_LICENSE = "GPLv3"
```
mineo: Since https://github.com/metabrainz/picard-plugins/pull/185 we are using SPDX license identifiers, so this should be either " GPL-3.0-only" or " GPL-3.0-or-later".

mineo (dismissed):

evandrocoan:
I hope I fixed everything.

evandrocoan:
Oops.

mineo:
Looks good to me. I'm not sure about two things that maybe @zas and @samj1912 know more about:

- All other single-file plugins are named `<pluginname>.py` instead of `__init__.py`. Since they're packaged in zip files, this shouldn't matter, I think?
- The plugin also appears to be active if it's disabled in the UI (at least I get an Album column in the main window). Was that also the case in Picard 1.x?

evandrocoan:
> All other single-file plugins are named <pluginname>.py instead of __init__.py. Since they're packaged in zip files, this shouldn't matter, I think?

I do not know what is the difference.

> The plugin also appears to be active if it's disabled in the UI (at least I get an Album column in the main window). Was that also the case in Picard 1.x?

Yes. I just tested. I do not why it does that. It seems to be loaded, even if it is disabled. May be it is a bug?

phw:
> All other single-file plugins are named <pluginname>.py instead of __init__.py. Since they're packaged in zip files, this shouldn't matter, I think?

Yes, it doesn't matter much the way we currently distribute the plugins as a ZIP. It does make a difference if you give this to users as single files. If it is named add_album_column.py the file can just be placed in the plugin directory as is. If it is called __init__.py creating a subdirectory is mandatory.

I personally don't have a strong opinion which naming to prefer with our current ZIP structure, both ways have benefits.

phw:
>> The plugin also appears to be active if it's disabled in the UI (at least I get an Album column in the main window). Was that also the case in Picard 1.x?

> Yes. I just tested. I do not why it does that. It seems to be loaded, even if it is disabled. May be it is a bug?

Not a bug really, but a limitation of the plugin system. Picard's plugin system does not directly support a plugin like this. In some way this plugin just abuses the fact that plugins get executed on load to monkey patch Picard. It really is more a hack then a proper plugin.

There is some discussion about redesigning the plugin system, and one point definitely should be to allow for explicit activation / deactivation hooks for plugins. Together with properly defined interfaces for some UI manipulation that would allow more flexibility here.

In the meantime maybe the fact that this plugin is always active and cannot be deactivated should be mentioned in the description.

phw (approved):

mxmilkiib:
Can I request a screenshot of what this plugin looks like?

evandrocoan:
Of course! 

It looks like this with Picard version `2.1.0` on Windows 10:
image

wisdomtooth:
I get (running from pip installed version):

`/usr/local/lib/python3.7/site-packages/picard/plugin.plugin_error:261: Failed loading zipped plugin 'add_album_column'`

evandrocoan:
There is some sort of bug, where you need to go to the `Picard's` plugin directory and manually unzip the plugin file:
1. image
1. image
1. image
1. image
1. image
1. Once you do that, you can restart `Picard's` and it will be working
1. image
1. Before restart:
image
1. After restart:
image

phw:
@wisdomtooth This was supposed to be fixed with https://github.com/metabrainz/picard-plugins/commit/af3de8bb57dc23075aa53645399c48d906d03b83 . Strange, I will check tomorrow

PR #202, opened 2019-01-23
instruments: Add a new plugin to populate an ~instruments tag.

dseomn:
Based on
https://github.com/dseomn/miscellaneous/blob/b353167085f6e70d770d494bd780858fdaf50843/personal-picard-plugins/instruments.py

### DIFF
--- a/plugins/instruments/instruments.py
+++ b/plugins/instruments/instruments.py
@@ -1,0 +1,89 @@
+# MusicBrainz Picard plugin to add an ~instruments tag.
+# Copyright (C) 2019  David Mandelberg
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


phw (approved):

dseomn:
@phw It looks like https://picard.musicbrainz.org/plugins/ was regenerated after this PR was merged, but I still don't see the plugin there (or in Picard). Any ideas?

phw:
Yes, we have an issue with the Picard website which cannot parse the plugin if it uses special Python 3 syntax. We are aware of that, but I currently can't say when we will have a fix. The quick workaround we did on the release day was to just skip plugins it could not process. Another quick way to solve it for this plugin would be to remove the type hints.

dseomn:
Thanks for the explanation, I just sent #213 to get the plugin working for now, and filed #214 so that I'll get a notification when it's time to revert #213.

PR #204, opened 2019-02-16
TheAudioDB cover art plugin

phw:
Plugin using [TheAudioDB](https://www.theaudiodb.com) as source. The functionality is mostly similar to the existing Fanart.tv plugin

### DIFF
--- a/plugins/theaudiodb/__init__.py
+++ b/plugins/theaudiodb/__init__.py
@@ -1,0 +1,156 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2015-2019 Philipp Wolfer
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


plugins/theaudiodb/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+
+    """Image from The Audio DB"""
+
+    support_types = True
+    sourceprefix = "AUDIODB"
+
+    def parse_url(self, url):
+        super().parse_url(url)
+        # Workaround for Picard always returning port 80 regardless of the scheme
+        self.port = self.url.port(443 if self.url.scheme() == 'https' else 80)
```
zas, line 56: I'm seeing this code too much, i wonder if we could make things better somehow.

zas (approved):

plugins/theaudiodb/__init__.py
phw: Yes, we need to review this. The point is that `url.port()` without parameter would return `-1` if the port is not explicitly specified in the URL, which for Qt means the default port. And this actually works well with Qt networking requests as well, if you pass it -1 it will use the appropriate port depending on the scheme. So likely we could just drop these checks alltogether. We just have to make sure we properly interpret port == -1 in other places we might use the port directly (e.g. we should not generate a URL "example.com:-1")

PR #205, opened 2019-02-16
BPM: Use a separate thread pool

### LINKED ISSUE #200: BPM plugin causes metadata tab not to update
When bulk updating tracks with bpm values using https://github.com/metabrainz/picard-plugins/commit/aaf833de6fe0d4e56163a8678316f75ffe2df141 the editor area become unresponsive. Assistance is needed to resolve. This is tested against Picard Version 2.1.1.dev2

Example of behavior: 
4 - lots of files not working 1 - reworked

phw:
Use a separate thread pool for the BPM calculation.

Otherwise the metadata box would not update on selection changes, since it used the same thread pool for updating the UI.

Also call `QtCore.QCoreApplication.processEvents()` during the BPM calculation, this keeps the UI more responsive. Otherwise it gets really laggy.

Fixes #200

### DIFF
--- a/plugins/bpm/__init__.py
+++ b/plugins/bpm/__init__.py
@@ -21,7 +21,6 @@ from aubio import source, tempo
 from numpy import median, diff
 from collections import defaultdict
 from functools import partial
-from PyQt5 import QtCore
 from subprocess import check_call
 from picard.album import Album, NatAlbum
 from picard.track import Track
@@ -47,20 +46,15 @@ class FileBPM(BaseAction):
     def __init__(self):
         super().__init__()
         self._close = False
-        self.thread_pool = QtCore.QThreadPool(self)
         self.tagger.aboutToQuit.connect(self._cleanup)
 
     def _cleanup(self):
-        if self._close:
-            return
         self._close = True
-        self.thread_pool.waitForDone()
 
     def _add_file_to_queue(self, file):
         thread.run_task(
             partial(self._calculate_bpm, file),
-            partial(self._calculate_bpm_callback, file),
-            thread_pool=self.thread_pool)
+            partial(self._calculate_bpm_callback, file))
 
     def callback(self, objs):
         for obj in objs:
--- a/plugins/bpm/__init__.py
+++ b/plugins/bpm/__init__.py
@@ -8,11 +8,11 @@
 #
 
 PLUGIN_NAME = "BPM Analyzer"
-PLUGIN_AUTHOR = "Len Joubert, Sambhav Kothari"
+PLUGIN_AUTHOR = "Len Joubert, Sambhav Kothari, Philipp Wolfer"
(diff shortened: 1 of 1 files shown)


phw:
@zas: What is your opinion on this? While this change works it is actually quite some work to implement due to all the cleanup work required.

I am thinking about introducing a `priority_thread_pool` in Picard instead, which would be used by operations where the user expects more or less instant feedback, in this case updating the metadata box when an item gets selected. Implementation of this inside Picard would be trivial (2 or 3 lines) and calculation heavy plugins like this one would not need to worry about it (Update: Something like https://github.com/phw/picard/commit/02d5bea9bb1c16644af6d5c68f64b19234af8132).

In case this comes up: I also tried to increase the priority of the tasks queued up by the metadata box update. This definitely improves the situation, but it is still very laggy. It could take over a second until the view updated, with a separate thread pool it was virtually instant.

zas:
> I am thinking about introducing a `priority_thread_pool` in Picard instead,

Yes, it might be better, limiting code redundancy if different plugins need the same feature.

madmouser1 (approved):
many thanks

phw:
@zas Looking at this at more detail I actually question that we need to start a thread at https://github.com/metabrainz/picard/blob/master/picard/ui/metadatabox.py#L449 at all.

The executed call is running Python code and due to the GIL does not run truely parallel anyway. Maybe it would be better to call it without a thread and instead call `QtCore.QCoreApplication.processEvents()` inside the for loops in `_update_tags()` so UI events are executed even if comparing a lot of files.

phw:
Scratch that last idea, it behaves worse then the currently threaded solution. Just using a separate thread pool for it is a good idea.

But my testing showed also something different: For the BPM module it actually makes some sense to implement that thread pool handling itself, because it means it can also abort the threads cleanly (see the hanlding of `self._close`). Without that Picard waits for all threads to close. If you are calculating the BPM values for several hundred files this means the Picard process keeps running for quite a while after the window has already been closed.

So my current recommendation is merging this PR as is but in addition also commit the separate thread pool for UI updates to Picard.

zas:
> So my current recommendation is merging this PR as is but in addition also commit the separate thread pool for UI updates to Picard.

LGTM

zas (approved):

phw:
Actually the calls to `QtCore.QCoreApplication.processEvents()` are nonsense here, since this only triggers processing of events in the current thread, and there are none. It would be an alternative to running in threads. I'll remove this call and merge then.

phw:
@zas I updated this again and took a middle route. I did not implement a separate thread pool, but use the default one. This will work fine with the patch for Picard 2.1.3, but of course will require users to update Picard to benefit for it.

But I kept the code for early exit from the threads. This makes exiting Picard behave much nicer, since BPM analyzing can run for several minutes if you have it running for many files. Actually I think we should look at more places inside Picard where we could exit threads earlier.

What do you think?

PR #207, opened 2019-02-17
Amazon cover art provider plugin

phw:
Amazon cover art provider as plugin.

See discussion at https://github.com/metabrainz/picard/pull/1101


zas (approved):

phw:
Following our discussion at https://github.com/metabrainz/picard/pull/1101 I'll set the API version to 2.2. This way the plugin won't be available for earlier Picard versions, where it does not make sense.

PR #210, opened 2019-03-01
musixmatch: Use Picard's webservice implementation

### LINKED ISSUE #158: musixmatch plugin does synchroneous requests
The musixmatch does all requests directly on the UI thread with `urllib.request`.

Should be updated to use the Picard webservice helpers instead.

phw:
Update MusixMatch plugin to use Picard's webservice classes instead of custom implementation.

Benefits:

- Requests are not blocking UI thread (fixes #158)
- Better handling of API key (previously API key was set only on startup, that means changing the API key required Picard restart)
- Less and cleaner code


zas (approved):
Much better imho

PR #211, opened 2019-03-04
Apiseeds Lyrics plugin

avalloneandrea:


plugins/apiseeds-lyrics/apiseeds-lyrics.py
```diff
@@ -0,0 +1,116 @@
...
+ratecontrol.set_minimum_delay((APISEEDS_HOST, APISEEDS_PORT), APISEEDS_RATE_LIMIT)
+
+
+def process_result(album, metadata, response, reply, error):
+
+    try:
+        data = load_json(response)
+        lyrics = data['result']['track']['text']
+        metadata['lyrics'] = lyrics
+        log.info('{}: lyrics found for track {}'.format(PLUGIN_NAME, metadata['title']))
```
phw: I think this and most other output here should actually be `log.debug`

plugins/apiseeds-lyrics/apiseeds-lyrics.py
```diff
@@ -0,0 +1,116 @@
...
+APISEEDS_HOST = 'orion.apiseeds.com'
+APISEEDS_PORT = 443
+APISEEDS_RATE_LIMIT = 60 * 1000 / 200
+ratecontrol.set_minimum_delay((APISEEDS_HOST, APISEEDS_PORT), APISEEDS_RATE_LIMIT)
+
+
+def process_result(album, metadata, response, reply, error):
+
+    try:
+        data = load_json(response)
```
phw: Any specific reason why you use `load_json` here? Most often the better approach is to call `webservice.get()` with `parse_response_type='json'`.

If parsing fails `error` will be set. Also you should check for `error` first. If it is not empty log it and return early.

plugins/apiseeds-lyrics/apiseeds-lyrics.py
```diff
@@ -0,0 +1,116 @@
...
+
+    finally:
+        album._requests -= 1
+        album._finalize_loading(None)
+
+
+def process_track(album, metadata, release, track):
+
+    apikey = config.setting['apiseeds_api_key']
+    if (apikey is None):
```
phw: I think this is better written as `if not apikey:`. Also this check currently failed in my tests. I did not debug fully, but it likely was set to an empty string in my case and not `None`. But just check with `not apikey` and it's save and looks a bit more pythonic.

plugins/apiseeds-lyrics/apiseeds-lyrics.py
```diff
@@ -0,0 +1,116 @@
...
+        log.error('{}: API key is missing, please provide a valid value'.format(PLUGIN_NAME))
+        return
+
+    artist = metadata['artist']
+    if (artist is None):
+        log.error('{}: artist is missing, please provide a valid value'.format(PLUGIN_NAME))
+        return
+
+    title = metadata['title']
+    if (title is None):
```
phw: Same here, check for `if not artist` and `if not title`

plugins/apiseeds-lyrics/apiseeds-lyrics.py
```diff
@@ -0,0 +1,116 @@
...
+    apiseeds_params = {'apikey': apikey}
+    album._requests += 1
+    log.info('{}: GET {}?{}'.format(PLUGIN_NAME, quote(apiseeds_path), urlencode(apiseeds_params)))
+
+    album.tagger.webservice.get(
+        APISEEDS_HOST,
+        APISEEDS_PORT,
+        apiseeds_path,
+        partial(process_result, album, metadata),
+        parse_response_type=None,
```
phw: See my comments above, I think it would be better to set `parse_response_type='json'` here.

plugins/apiseeds-lyrics/apiseeds-lyrics.py
```diff
@@ -0,0 +1,116 @@
...
+def process_result(album, metadata, response, reply, error):
+
+    try:
+        data = load_json(response)
+        lyrics = data['result']['track']['text']
+        metadata['lyrics'] = lyrics
+        log.info('{}: lyrics found for track {}'.format(PLUGIN_NAME, metadata['title']))
+
+    except:
+        log.info('{}: lyrics NOT found for track {}'.format(PLUGIN_NAME, metadata['title']))
```
phw: Since this is logging in case of an exception I think you should log it as an error with `exc_info=True` (will include stack trace information) set. Something like:

```
log.error('{}: failed fetching lyrics for track {}'.format(PLUGIN_NAME, metadata['title']), exc_info=True)
```

Actually you can also do something like the following, but your use of `string.format` would be consistent with the rest of your code:

```
log.error('%s: failed fetching lyrics for track %s', PLUGIN_NAME, metadata['title'], exc_info=True)
```

phw (changes requested):
Thanks a lot for this, this is awesome. Pretty clean implementation. I have a couple of fixes and improvements, but actually nothing big. Looking forward to getting this merged.

plugins/apiseeds-lyrics/apiseeds-lyrics.py
avalloneandrea: Actually, after replacing `load_json` with `parse_response_type='json'`, the only possibile exception is a KeyError raised by `data['result']['track']['text']`. I don't think it's useful to have the stack trace in this case.

phw: Agreed, yes

avalloneandrea:
I've implemented your suggestions in version 1.0.1, and moved the processor into it's own class in version 1.0.2...but maybe this wasn't a good idea, because it raised a _Method could be a function_ issue

phw:
Thanks for the update

> because it raised a _Method could be a function_ issue

You could just use `@staticmethod` on this method.

plugins/apiseeds_lyrics/apiseeds_lyrics.py
```diff
@@ -0,0 +1,122 @@
...
+        album.tagger.webservice.get(
+            self.apiseeds_host,
+            self.apiseeds_port,
+            apiseeds_path,
+            partial(self.process_response, album, metadata),
+            parse_response_type='json',
+            priority=True,
+            queryargs=apiseeds_params)
+
+    def process_response(self, album, metadata, document, reply, error):
```
phw: You still need to check for `error` with something like:

```
if error:
    album._requests -= 1
    album._finalize_loading(error)
    return
```

plugins/apiseeds_lyrics/apiseeds_lyrics.py
```diff
@@ -0,0 +1,122 @@
...
+
+        try:
+            lyrics = document['result']['track']['text']
+            metadata['lyrics'] = lyrics
+            log.debug('{}: lyrics found for track {}'.format(PLUGIN_NAME, metadata['title']))
+
+        except:
```
phw: Better check explicitly for `KeyError` that's the only one that should be raised here

plugins/apiseeds_lyrics/apiseeds_lyrics.py
```diff
@@ -0,0 +1,122 @@
...
+from picard.metadata import register_track_metadata_processor
+from picard.ui.options import register_options_page, OptionsPage
+from picard.webservice import ratecontrol
+
+PLUGIN_NAME = 'Apiseeds Lyrics'
+PLUGIN_AUTHOR = 'Andrea Avallone'
+PLUGIN_DESCRIPTION = 'Fetch lyrics from Apiseeds Lyrics, which provides millions of lyrics from artist all around the world. ' \
+                     'Lyrics provided are for educational purposes and personal use only. Commercial use is not allowed. ' \
+                     'In order to use Apiseeds you need to get a free API key at <em>https://apiseeds.com</em>. ' \
+                     'Want to contribute? Check out the project page at <em>https://github.com/avalloneandrea/apiseeds-lyrics</em>!'''
```
phw: I think an empty string `''` sneaked in at the end of this line ;)

phw (changes requested):
Looks good so far. Only the `error` needs to be handled and some minor tweak mainly to make codacy happy

plugins/apiseeds_lyrics/apiseeds_lyrics.py
avalloneandrea: Added this:
```
if error:
    log.info('{}: lyrics NOT found for track {}'.format(PLUGIN_NAME, metadata['title']))
    album._requests -= 1
    album._finalize_loading(error)
    return
```

and this:
```
except KeyError:
```

but Picard crashes when the API is unable to find lyrics. This is the last thing I see in the console:
```
E: 11:24:18,237 webservice._handle_reply:383: Network request error for https://orion.apiseeds.com:443/api/music/lyric/artist/track?apikey=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
: Error transferring https://orion.apiseeds.com:443/api/music/lyric/artist/track?apikey=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 - server replied: Not Found (QT code 203, HTTP code 404)
```
How can I investigate the issue?

phw: You probably should change it back to catch all exceptions again, but actually add the stacktrace option. This way you will also see what caused your crashes.

avalloneandrea: I have double checked my code, this is working as expected now:
```
@staticmethod
def process_response(album, metadata, document, reply, error):

    if error:
        log.info('{}: lyrics found for track {}'.format(PLUGIN_NAME, metadata['title']))
        album._requests -= 1
        album._finalize_loading(None)
        return

    try:
        lyrics = document['result']['track']['text']
        metadata['lyrics'] = lyrics
        log.info('{}: lyrics found for track {}'.format(PLUGIN_NAME, metadata['title']))

    except KeyError:
        log.info('{}: lyrics NOT found for track {}'.format(PLUGIN_NAME, metadata['title']))

    finally:
        album._requests -= 1
        album._finalize_loading(None)
```

However I notice that those three lines of code are written twice, and I'm not really a fan of code duplications:
```
log.info('{}: lyrics found for track {}'.format(PLUGIN_NAME, metadata['title']))
album._requests -= 1
album._finalize_loading(None)
```

Therefore I'm more prone to _"ask forgiveness not permission"_: don't check `error`, let `document` be `None`, and catch the `TypeError` raised by the first access to `document`:
```
@staticmethod
def process_response(album, metadata, document, reply, error):

    try:
        lyrics = document['result']['track']['text']
        metadata['lyrics'] = lyrics
        log.info('{}: lyrics found for track {}'.format(PLUGIN_NAME, metadata['title']))

    except (TypeError, KeyError):
        log.info('{}: lyrics NOT found for track {}'.format(PLUGIN_NAME, metadata['title']))

    finally:
        album._requests -= 1
        album._finalize_loading(None)
```

What do you think about it?

phw: I think it should be fine. In case of `error != None` this error will be logged anyway also by the web service module.

It also depends a bit on what we expect this plugin to cause in case of an error. My original suggestion used `album._finalize_loading(error)`. This will cause the entire album loading to fail. Probably with a plugin like this we don't want to be so strict. So I'm totally fine with your solution.

phw (approved):
This is looking good now to me, great work. I'll merge this and it will be available once we update the website (which hopefully will be soon, we are working on some website fixes currently).

PR #213, opened 2019-03-09
instruments: Remove type annotations.

dseomn:
https://github.com/metabrainz/picard-plugins/pull/202#issuecomment-470806751
> Yes, we have an issue with the Picard website which cannot parse the
> plugin if it uses special Python 3 syntax.

### DIFF
--- a/plugins/instruments/instruments.py
+++ b/plugins/instruments/instruments.py
@@ -20,18 +20,18 @@ PLUGIN_DESCRIPTION = """\
   Adds a multi-valued tag of all the instruments (including vocals), for use in
   scripts.
   """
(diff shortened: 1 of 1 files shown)


phw:
Thanks a lot for this, but I have some hope we won't need it. We just merged https://github.com/metabrainz/picard-website/pull/150 , which upgrades our website with proper Python 3 support. If this works out as expected we will be able to update the website with your plugin included without applying any workarounds. But I'll keep your PR here open for now in case we get into some trouble with the website upgrade.

phw:
We deployed the fixes to the website, the plugin is now available.

https://picard.musicbrainz.org/plugins/

PR #219, opened 2019-05-06
Add album_level_tags plugin

Harakku:
New plugin to transform multiple file specific tags into single
strings which are usable in the file naming script via script
functions.


plugins/album_level_tags/album_level_tags.py
```diff
@@ -0,0 +1,360 @@
...
+#)
+from picard.metadata import register_album_metadata_processor
+from picard.script import register_script_function
+
+#
+#
+# START OF MONKEYPATCHED HOOKS
+#
+#
+
```
zas: In fact, it would be great to submit PRs to add those hooks to the core Picard code, so they'll be in next stable version. There's already a PR for a post file save hook (https://github.com/metabrainz/picard/pull/1186).

Harakku: I did that, removed the now unneeded code from the plugin and tested that it still works the same as before. Links to PRs: https://github.com/metabrainz/picard/pull/1190, https://github.com/metabrainz/picard/pull/1191

I bumped the needed plugin API version to "2.3" since the plugin doesn't work with the current builds.

I also read more about licensing and concluded that since the plugin imports functions from Picard, the plugin needs to be GPL2 too. No big deal though.

plugins/album_level_tags/album_level_tags.py
```diff
@@ -0,0 +1,290 @@
...
+def _file_removal_helper(tag_variable, album_mbid, metadata, metadata_key):
+    # Take a look at the notes at _file_addition_helper
+    try:
+        metadata_value = metadata[metadata_key]
+    except KeyError:
+        return
+
+    try:
+        tag_variable[album_mbid][metadata_value] -= 1
```
sambhav: This logic and the one above might be simplified using either collections.Counter or collections.defaultdict.

Harakku: That made a big difference and those changes kinda snowballed into making everything else cleaner too. :-)

There are now some formatting options in the script functions. Based on my manual testing, everything seems to work.

zas (approved):
LGTM

@phw: ok for you ?

Harakku:
I just realized how this plugin could be made much cleaner. Instead of scribing down tags as they come by, an album object can be saved as a global variable. Then the album object can be accessed when the script functions are being called. Duhhh.

Gonna make those changes soonish.

zas:
> Gonna make those changes soonish.

Should we merge as is or do you have changes to make before?

Harakku:
Almost forgot to update this. The plugin has some option pages now and quite big changes in the way it works. Didn't end up needing the new hooks after all. (Awkward.) The option editing is a bit rough but I hope it's somewhat easy to understand.

All in all, I'm happy with the plugin now.

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+        try:
+            album = parser.file.parent.album
+        except AttributeError:  # Happens with the file naming example box
+            return metadata_key.strip("~").upper()
+        if isinstance(album, NatAlbum):
+            raise TypeError("{0}: NAT-albums are not processed.".format(PLUGIN_NAME))
+
+        # Check if cached data is recent enough to be used (i.e. we can assume that the user hasn't modified the tags
+        # in the meantime) or remove other releases from the cache to keep it small
+        global tag_cache
```
zas, line 103: `global` keyword isn't strictly needed here I think.

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+        global last_run_time
+        album_cache = tag_cache.get(album.id, {})
+        if album_cache:
+            t = time.time()
+
+            cached_results = album_cache.get((metadata_key, option_profile))
+            if cached_results and t - last_run_time < 4:  # seconds
```
zas, line 110: I'd replace `4` with a explicit variable/constant (something like `DELAY_IN_SECONDS`) and remove the comment.

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+        if isinstance(album, NatAlbum):
+            raise TypeError("{0}: NAT-albums are not processed.".format(PLUGIN_NAME))
+
+        # Check if cached data is recent enough to be used (i.e. we can assume that the user hasn't modified the tags
+        # in the meantime) or remove other releases from the cache to keep it small
+        global tag_cache
+        global last_run_time
+        album_cache = tag_cache.get(album.id, {})
+        if album_cache:
+            t = time.time()
```
zas, line 107: Instead of `t`, I'd use `now` or `current_time` since it stores current time.

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+            options.update(DEFAULT_AVERAGE_OVERRIDE_VALUES)
+        try:
+            all_user_options = json.loads(album.tagger.config.setting["a_l_t_options"])
+            profile_options = all_user_options.get(option_profile, {})
+        except json.decoder.JSONDecodeError:
+            raise SyntaxError("{0}: Syntax error in options.".format(PLUGIN_NAME))
+
+        if option_profile is not None and option_profile not in all_user_options:
+            raise KeyError("{0}: Unknown option profile \"{1}\" used in a script.".format(PLUGIN_NAME, option_profile))
+        if profile_options.get(FORMATTER_WHEN_ONE_OPTION):
```
zas, line 129: Perhaps:
```python
formatter2_is_default = False if profile_options.get(FORMATTER_WHEN_ONE_OPTION) else True
```
Since you're using few of those below.

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+        formatter_when_one = formatter if formatter2_is_default else options[FORMATTER_WHEN_ONE_OPTION]
+        loaded_files_only = options[LOADED_FILES_ONLY_OPTION]
+        grouping_key = normalize_tagname(options[GROUPING_OPTION])
+        replacements_profile = options[REPLACEMENTS_PROFILE_OPTION]
+
+        # Get tags
+        group_by_medium = True if metadata_key == "media" else False
+        discnumber = 0
+        temp = defaultdict(list)
+        for t in album.tracks:
```
zas, line 148: Use `track` instead of `t`.

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+            discnumber += 1
+            if t.num_linked_files > 0:
+                for lf in t.linked_files:
+                    temp[lf.metadata.get(grouping_key)] += lf.metadata.getall(metadata_key)
+            elif not loaded_files_only:
+                temp[t.metadata.get(grouping_key)] += t.metadata.getall(metadata_key)
+
+        if average_mode:
+            tag_counter = Counter({ sum(map(float, tag_list)) / len(tag_list): 1
+                                    for tag_list in temp.values() if len(tag_list) > 0 })
```
zas, line 161: This expression deserves at least a comment, you may also cache `len(tag_list)`.

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+        # Rename tag values if that's wanted behaviour
+        try:
+            replacements = json.loads(album.tagger.config.setting["a_l_t_replacements"]).get(replacements_profile, {})
+        except json.decoder.JSONDecodeError:
+            raise SyntaxError(PLUGIN_NAME + ": Syntax error in replacements.")
+        tag_counter = Counter({ replacements.get(tag, tag): count for tag, count in tag_counter.items() })
+
+        # Effectively removes the None key from the counter
+        tag_counter[None] = 0
+        tag_counter = +tag_counter
```
zas, line 174: Typo? Or?

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+                label.setText("Option error! \"{0}\" expected {1}, got {2}.".format(
+                    option, str(type(defaultvalue)), str(type(value))))
+                return
+        label.setText("Success! Valid syntax. Valid options.")
+
+    def replacements_validation_func(self):
+        text_area = self.ui.plainTextEdit_raw_2
+        label = self.ui.label_valid_raw_2
+        try:
+            _ = json.loads(text_area.toPlainText())
```
zas, line 343: Please avoid the use of `_`, it is reserved for gettextization (see https://github.com/metabrainz/picard/blob/master/picard/i18n.py)

plugins/album_level_tags/__init__.py
```diff
@@ -0,0 +1,360 @@
...
+    normalize_tagname,
+    register_script_function,
+)
+from picard.ui.options import (
+    OptionsPage,
+    register_options_page,
+)
+from PyQt5 import QtGui
+from PyQt5.QtWidgets import QButtonGroup
+import time
```
zas, line 59: You might want to re-order imports to fit usual Picard import order (see https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md and search for `isort`).

zas (changes requested):

Sophist-UK:
I can't help but wonder whether it might be better for this plugin simply to calculate the lists and averages of various file metadata and place them in the album metadata as new variables.

Harakku:
Closing. It looks like these features were made available in the "Release tag aggregation functions" plugin: https://github.com/metabrainz/picard-plugins/pull/300

PR #221, opened 2019-06-21
Metabrainz/2.0

MetaTunes:
Version 2.0.4: Fix occasional regex backtracking crash. Make naming of movement tags consistent with Picard docs. Added an option to attempt to get works and movement info from title if there are no work relationships (requires title in form "work: movement"). If Muso-specific genre processing is selected (or XML reference file is provided including classical composers) and there is no composer (because of a lack of work relationship) then the plugin will check the listed artist against the reference list of classical composers and, if there is a match, will populate the composer metadata and set the genre to classical.


plugins/classical_extras/__init__.py
```diff
@@ -6935,26 +7056,45 @@ def publish_metadata(self, release_id, album, track):
                     tm,
                     'show work movement',
                     '1')  # for iTunes
+                self.append_tag(
+                    release_id,
+                    tm,
+                    'showmovement',
+                    '1')  # consistent with Picard tag docs
```
phw, line 7063: It's a bit co fusing that it both writes "showmovement" and "show work movement" and the comment states that the latter is targeting iTunes. "showmovement" was explicitly added for iTunes and MusicBee (which took the iTunes implementation and extended it to other formats). So it actually will be saved to the files in a way that makes it work in iTunes (for MP4) and other formats (MusicBee only), which is not true for "show work movement".

phw (commented):
I am a bit confused by the showmovement tag handling, but apart from this I think this is good to merge.

plugins/classical_extras/__init__.py
MetaTunes: "show work movement" pre-dates Picard's implementation of "showmovement" and worked with iTunes (but not MusicBee, I think). In this release I updated the plugin to include "showmovement" but left the earlier tag there for backwards compatibility (in case someone is using it directly rather than via iTunes). I will modify the comments to clarify this in my development copy so that it is in the next release, if that's OK (I'd rather not mess with the code for this release unless it is necessary for other reasons).
See discussion at https://community.metabrainz.org/t/classical-extras-2-0/394627/3 and https://community.metabrainz.org/t/movements-saved-in-txxx-id3v2-4-frames-instead-of-mvnm-ones/404300/3

MetaTunes: Code will read
`            if '~cwp_part_levels' in tm and int(tm['~cwp_part_levels']) > 0:
                  self.append_tag(
                        release_id,
                        tm,
                        'show work movement',
                        '1')  # original tag for iTunes, kept for backwards compatibility
                  self.append_tag(
                        release_id,
                       tm,
                        'showmovement',
                        '1')  # new tag for iTunes & MusicBee, consistent with Picard tag docs`
in next version.
BTW, if anyone really objects to having a "show work movement" tag, they can always unset it.

MetaTunes:
Any reason why Travis is still not completed?
Is this OK to merge now?

phw (approved):

phw:
@MetaTunes I have merged it, thank you very much for submitting. There was only a small merge conflict in the version line I fixed.

PR #223, opened 2019-07-05
Add hyphen_unicode plugin

alanswanson:
Cloned non_ascii_equivalents.py plugin to only replace unicode
character HYPHEN (U+2010) [0xE2 0x80 0x90] with typographically
identical HYPHEN-MINUS (U+002D) [0x2D] for fonts that do not
support HYPHEN and to prevent visually duplicate filenames
differentiated only by their hyphens.

Since text processing on music tags is rare so choice is purely
pedantic especially as keyboards only have HYPHEN-MINUS.

Discussion of hyphen issue on Musicbrainz forums;
https://community.metabrainz.org/t/correct-hyphen-unicode-hyphen-or-hyphen-minus/19610

### DIFF
--- a/plugins/hyphen_unicode/hyphen_unicode.py
+++ b/plugins/hyphen_unicode/hyphen_unicode.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (C) 2016 Anderson Mesquita <andersonvom@gmail.com>
+# Copyright (C) 2019 Alan Swanson <reiver@improbability.net>
+#
+# This program is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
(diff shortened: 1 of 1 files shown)


plugins/hyphen_unicode/hyphen_unicode.py
```diff
@@ -0,0 +1,79 @@
...
+"albumartist", "albumartists" and "albumartistsort" tags.'''
+
+# Based on Non-ASCII Equivalents plugin.
+# Musicbrainz form discussion on HYPHEN versus HYPHEN-MINUS at;
+# https://community.metabrainz.org/t/correct-hyphen-unicode-hyphen-or-hyphen-minus/19610
+
+CHAR_TABLE = {
+    # HYPHEN to HYPHEN-MINUS
+    "‐": "-",
+}
```
zas, line 47: Perhaps it would make sense to use [str.translate](https://docs.python.org/3.7/library/stdtypes.html#str.translate)

zas (approved):

Issue #222, opened 2019-07-02
Allow Remove Perfect Albums plugin to remove releases with multiple files per track


phw:
See https://community.metabrainz.org/t/split-albums-in-album-list-where-multiple-track-copies-are-found/433254

PR #224, opened 2019-07-09
Added option 'Remove perfect albums (allowing multiple file extensions)'

dsparkplug:
This option treats albums with multiple complete sets of track files but with different extensions as perfect.

See #222

### DIFF
--- a/plugins/remove_perfect_albums/remove_perfect_albums.py
+++ b/plugins/remove_perfect_albums/remove_perfect_albums.py
@@ -1,11 +1,14 @@
 PLUGIN_NAME = 'Remove Perfect Albums'
-PLUGIN_AUTHOR = 'ichneumon, hrglgrmpf'
+PLUGIN_AUTHOR = 'ichneumon, hrglgrmpf, dsparkplug'
 PLUGIN_DESCRIPTION = '''Remove all perfectly matched albums from the selection.'''
-PLUGIN_VERSION = '0.2'
+PLUGIN_VERSION = '0.3'
 PLUGIN_API_VERSIONS = ['0.15', '2.0']
 PLUGIN_LICENSE = "GPL-2.0"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
 
+import os
+import os.path
+from picard import log
 from picard.album import Album
 from picard.ui.itemviews import BaseAction, register_album_action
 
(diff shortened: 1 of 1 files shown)


plugins/remove_perfect_albums/remove_perfect_albums.py
```diff
@@ -19,4 +22,65 @@ def callback(self, objs):
               	and album.get_num_unsaved_files() == 0 and album.loaded == True):
                 self.tagger.remove_album(album)
 
+class RemovePerfectAlbumsAllowingMultipleFileExtensions(BaseAction):
+    NAME = 'Remove perfect albums (allowing multiple file extensions)'
+
+    def callback(self, objs):
+        for album in objs:
+            
+            if (not(isinstance(album, Album))):
```
zas, line 31: You don't need to add parenthesis here:

```python
if not isinstance(album, Album):
```

plugins/remove_perfect_albums/remove_perfect_albums.py
```diff
@@ -19,4 +22,65 @@ def callback(self, objs):
               	and album.get_num_unsaved_files() == 0 and album.loaded == True):
                 self.tagger.remove_album(album)
 
+class RemovePerfectAlbumsAllowingMultipleFileExtensions(BaseAction):
+    NAME = 'Remove perfect albums (allowing multiple file extensions)'
+
+    def callback(self, objs):
+        for album in objs:
+
```
zas, line 30: This whiteline should be removed. See PEP8 for Python style recommendations.

plugins/remove_perfect_albums/remove_perfect_albums.py
```diff
@@ -19,4 +22,65 @@ def callback(self, objs):
...
+    def callback(self, objs):
+        for album in objs:
+            
+            if (not(isinstance(album, Album))):
+                continue
+           
+            title = album.column('title')
+            log.info('Checking album perfection: %s', title)
+            
+            if (album.loaded != True):
```
zas, line 37: Useless parenthesis and test can be done this way:

```python
if album.loaded:
```

plugins/remove_perfect_albums/remove_perfect_albums.py
```diff
@@ -19,4 +22,65 @@ def callback(self, objs):
...
+                log.info('Album is not loaded')
+                continue
+
+            unsavedFiles = album.get_num_unsaved_files()
+            if (unsavedFiles > 0):
```
zas, line 42:
```python
if unsavedFiles:
```

plugins/remove_perfect_albums/remove_perfect_albums.py
```diff
@@ -19,4 +22,65 @@ def callback(self, objs):
...
+                log.info('Album has %s unsaved files', unsavedFiles)
+                continue
+            
+            unmatchedFiles = album.get_num_unmatched_files()
+            if (unmatchedFiles > 0):
```
zas, line 47:
```python
if unmatchedFiles:
```

plugins/remove_perfect_albums/remove_perfect_albums.py
```diff
@@ -19,4 +22,65 @@ def callback(self, objs):
...
+                continue
+
+            if (totalFiles % numberOfTracks != 0):
+                log.info('The number of files %s in album is not wholly divisible by the number of tracks %s', totalFiles, numberOfTracks)
+                continue
+
+            #get dictionary of file extension counts
+            extDict = {}
+            for file in album.iterfiles():
+                _, ext = os.path.splitext(file.filename)
```
zas, line 64: Do not use `_` as placeholder, it is used for translations (`_()` is a function), just use `unused`

zas (changes requested):
Please respect [PEP8](https://www.python.org/dev/peps/pep-0008/)

I didn't test the PR yet, but I will soon.

zas:
@dsparkplug any update?

Sophist-UK:
IMO this is trying to do in a plugin what should be done in Picard proper i.e. because we need the icons to reflect the perfect nature of the album or otherwise - and this PR will essentially remove albums which are shown as needing saving in the UI. We need Picard to recognise perfect albums which have multiple formats.

phw:
As this hasn't gone anywhere I'm going to close this for now. Can be re-opened if needed.

And I agree with @Sophist-UK, we could at least have an option to treat a release with multiple files per track as completed. Maybe at least if each track has the same amount of files and they are of different formats.

Issue #226, opened 2019-08-02
In "Generate Cuesheet", incorrect file extension used


benjaminglass1:
This issue is in reference to the following line:
https://github.com/metabrainz/picard-plugins/blob/31f290bd8cad703188308ac5e8191d7b439310fd/plugins/cuesheet/cuesheet.py#L187
The "Generate Cuesheet" plugin appears to be hardcoded to always list the file type as MP3, but the Cuesheet spec (at least according to the [HydrogenAudio wiki](http://wiki.hydrogenaud.io/index.php?title=Cue_sheet#Most_often_used)) has a few different file descriptors that may be used. Particularly, lossless files such as FLAC should be listed as having a WAV file type.

benjaminglass1:
Resolved by PR #227

PR #227, opened 2019-08-03
Add file type checking for CUE spec compliance

benjaminglass1:
Solves issue #226 . Adds a fairly rudimentary file extension checking system that should write the correct comments into the CUE files.

### DIFF
--- a/plugins/cuesheet/cuesheet.py
+++ b/plugins/cuesheet/cuesheet.py
@@ -182,14 +182,16 @@ class GenerateCuesheet(BaseAction):
                 t.set("INDEX", "01", "%02d:%02d:%02d" % (mm, ss, ff))
                 for file in track.linked_files:
                     audio_filename = file.filename
+                    extension = audio_filename.split(".")[-1].lower()
                     if os.path.dirname(filename) == os.path.dirname(audio_filename):
                         audio_filename = os.path.basename(audio_filename)
-                    if audio_filename.split(".")[-1].lower() in ["mp3", "mp2", "m2a"]:
-                        cuesheet.tracks[i].set("FILE", audio_filename, "MP3")
-                    elif audio_filename.split(".")[-1].lower() in ["aiff", "aif", "aifc"]:
-                        cuesheet.tracks[i].set("FILE", audio_filename, "AIFF")
+                    if extension in ["mp3", "mp2", "m2a"]:
+                        file_type = "MP3"
+                    elif extension in ["aiff", "aif", "aifc"]:
+                        file_type = "AIFF"
                     else:
-                        cuesheet.tracks[i].set("FILE", audio_filename, "WAV")
+                        file_type = "WAV"
+                    cuesheet.tracks[i].set("FILE", audio_filename, file_type)
 
             cuesheet.write()
 
--- a/plugins/cuesheet/cuesheet.py
+++ b/plugins/cuesheet/cuesheet.py
@@ -3,7 +3,7 @@
 PLUGIN_NAME = "Generate Cuesheet"
 PLUGIN_AUTHOR = "Lukáš Lalinský, Sambhav Kothari"
 PLUGIN_DESCRIPTION = "Generate cuesheet (.cue file) from an album."
-PLUGIN_VERSION = "1.1"
+PLUGIN_VERSION = "1.2"
 PLUGIN_API_VERSIONS = ["2.0"]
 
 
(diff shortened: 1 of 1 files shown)


zas (changes requested):
Looks ok to me but you should increase plugin version number too.

benjaminglass1:
@zas I amended the commit to add that. :+1:

zas (approved):

plugins/cuesheet/cuesheet.py
```diff
@@ -184,7 +184,12 @@ def callback(self, objs):
                     audio_filename = file.filename
                     if os.path.dirname(filename) == os.path.dirname(audio_filename):
                         audio_filename = os.path.basename(audio_filename)
-                    cuesheet.tracks[i].set("FILE", audio_filename, "MP3")
+                    if audio_filename.split(".")[-1].lower() in ["mp3", "mp2", "m2a"]:
```
zas: You can perhaps reduce the code redundancy a bit, with something like:

```python
extension = audio_filename.split(".")[-1].lower()
if extension in ["mp3", "mp2", "m2a"]:
    filetype = "MP3"
elif extension in ["aiff", "aif", "aifc"]:
    filetype = "AIFF"
else:
   filetype = "WAV"
cuesheet.tracks[i].set("FILE", audio_filename, filetype)
```

benjaminglass1: Got it. :+1:

zas (approved):

plugins/cuesheet/cuesheet.py
```diff
@@ -182,9 +182,16 @@ def callback(self, objs):
...
+                    extension = audio_filename.split(".")[-1].lower()
                     if os.path.dirname(filename) == os.path.dirname(audio_filename):
                         audio_filename = os.path.basename(audio_filename)
-                    cuesheet.tracks[i].set("FILE", audio_filename, "MP3")
+                    if extension in ["mp3", "mp2", "m2a"]:
+                        file_type = "MP3"
+                    elif extension in ["aiff", "aif", "aifc"]:
+                        file_type = "AIFF"
+                    else:
+                        file_type = "WAV"
```
phw, line 193: @benjaminglass1 Sorry for jumping in here so late, but this part puzzles me a bit. Why does it set everything else to WAV? I kind of get how this makes sense for FLAC (equaling WAV with lossless), but does it also make sense for thing like WMA and MP4?

benjaminglass1: According to the [archived cue sheet spec](https://web.archive.org/web/20160201021136/http://digitalx.org/cue-sheet/syntax/) and the HydrogenAudio wiki page for [cue sheets](http://wiki.hydrogenaud.io/index.php?title=Cue_sheet#Most_often_used), the acceptable 5 file types are WAVE (which I just realized I misspelled [facepalm]), MP3, AIFF, BINARY and MOTOROLA. BINARY and MOTOROLA were both specifically for some form of binary format that afaik doesn't exist anymore, which means all audio files get stuck in the other 3 categories. Based on the HydrogenAudio wiki, it appears to be the convention that files not fitting neatly into MP3 or AIFF should be categorized as WAVE.

phw: Ok, got it. Thanks for the clarification :) Will you submit a PR for the `WAVE` fix? I can also just push this minor fix directly if you prefer.

benjaminglass1:
I'd rather just have you push the fix directly, if you don't mind.




On Fri, Aug 9, 2019 at 7:34 AM -0500, "Philipp Wolfer" <notifications@github.com> wrote:












@phw commented on this pull request.





In plugins/cuesheet/cuesheet.py:
>                      if os.path.dirname(filename) == os.path.dirname(audio_filename):
                         audio_filename = os.path.basename(audio_filename)
-                    cuesheet.tracks[i].set("FILE", audio_filename, "MP3")
+                    if extension in ["mp3", "mp2", "m2a"]:
+                        file_type = "MP3"
+                    elif extension in ["aiff", "aif", "aifc"]:
+                        file_type = "AIFF"
+                    else:
+                        file_type = "WAV"



Ok, got it. Thanks for the clarification :) Will you submit a PR for the WAVE fix? I can also just push this minor fix directly if you prefer.

plugins/cuesheet/cuesheet.py
benjaminglass1: Push it directly, if you don't mind. Seems too small for a full pull request.

phw: Done in https://github.com/metabrainz/picard-plugins/commit/70b797d139d8d1a5089681e4530c4d04dda031ee

PR #232, opened 2019-08-24
Replace Forbidden Windows Symbols Plugin

AlexRustler:
Replace forbidden symbols :, \, /, ? etc with similar Unicode symbols.

### DIFF
--- a/plugins/replace_forbidden_symbols/replace_forbidden_symbols.py
+++ b/plugins/replace_forbidden_symbols/replace_forbidden_symbols.py
@@ -1,4 +1,4 @@
-﻿# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 
 # Copyright (C) 2019 Alex Rustler <alex_rustler@rambler.ru>
 #
--- a/plugins/replace_forbidden_symbols/replace_forbidden_symbols.py
+++ b/plugins/replace_forbidden_symbols/replace_forbidden_symbols.py
@@ -16,17 +16,21 @@
 # this program. If not, see <http://www.gnu.org/licenses/>.
 
 from picard import metadata
+from picard.script import register_script_function
 
 PLUGIN_NAME = "Replace Forbidden Symbols"
 PLUGIN_AUTHOR = "Alex Rustler <alex_rustler@rambler.ru>"
-PLUGIN_VERSION = "0.1"
+PLUGIN_VERSION = "0.2"
 PLUGIN_API_VERSIONS = ["0.9", "0.10", "0.11", "0.15", "2.0", "2.2"]
 PLUGIN_LICENSE = "GPL-3.0-or-later"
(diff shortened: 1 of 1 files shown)


zas (changes requested):
You may fix syntax (whitespaces, tabs, whitelines) according to PEP8.

plugins/replace_forbidden_symbols/replace_forbidden_symbols.py
```diff
@@ -0,0 +1,77 @@
...
+#
+# You should have received a copy of the GNU General Public License along with
+# this program. If not, see <http://www.gnu.org/licenses/>.
+
+from picard import metadata
+
+PLUGIN_NAME = "Replace Forbidden Symbols"
+PLUGIN_AUTHOR = "Alex Rustler <alex_rustler@rambler.ru>"
+PLUGIN_VERSION = "0.1"
+PLUGIN_API_VERSIONS = ["0.9", "0.10", "0.11", "0.15", "2.0"]
```
phw: You could add 2.2 here as well, that is the version of the next API version in the upcoming release. But your plugin would still run without this, this is just to be on the safe side for the future and to clearly indicate compatibility.

phw (approved):
Thanks, this is looking good for me.

Just a general suggestion: You could consider adding a scripting function `$replace_forbidden()` for this using `register_script_function`. This way the user have control to apply the replacement to any tag they want. E.g.

```
$set(composer,$replace_forbidden(%composer%))
```

See the keep or videotools plugins for examples

plugins/replace_forbidden_symbols/replace_forbidden_symbols.py
AlexRustler: Hello!
Thanks for review, but I need help for use register_script_function.
When I write this code:
`
@register_script_function
def replace_forbidden(value):
    return [fix_forbidden(x) for x in value]   `

and script in Picard:
`$set(composer,$replace_forbidden(%composer%))`

I got error:  replace_forbidden() takes 1positional argumnet but 2 were given.

I try this:
`
@register_script_function
def script_forbidden(parser, value):
    return replace_forbidden(value)`

and script in Picard:
`$set(composer,$script_forbidden(%composer%))`

and got error: sequence item 0: expected str instance, list found.

I don't understand reason of errors.

phw: The problem here is that your replace_forbidden expects a list, but tagger script operates on strings. The proper function to wrap would be the `fix_forbidden` function. I would add something like this:

```python
def script_replace_forbidden(parser, value):
    return fix_forbidden(value)

register_script_function(script_replace_forbidden, name="replace_forbidden")
```

zas (approved):

phw (approved):

PR #235, opened 2019-09-18
classical_extras: Fixed config access to ce_tab option

phw:
The plugin was doing some shenanigans to access the value of the `ce_tab` option and accessed internal structure of the config. This causes crashes with Picard 2.2 since config was refactored.

Changed now to use proper APIs and fixing the root cause (option was not defined).


phw:
I think we should rather merge https://github.com/metabrainz/picard-plugins/pull/236 to not create a fork from upstream code.

Issue #238, opened 2019-09-20
[Feature Request]: Better convertation from Unicode to ASCII


Nero-A:
In Picard there is the [option] (https://picard.musicbrainz.org/docs/options/) "Metadata" / "Convert Unicode punctuation characters to ASCII".
The converting is quite useful, especially if the local media player has lacks with Unicode punctuation  or mixed use of Unicode and ASCII (like Kodi).

First it's IMHO a little bit inaccurate for me because ASCII is actually a 7 bit code but the converting might be to an 8 bit code (e.g. with German umlauts). It seems to be an [extended ASCII](https://en.wikipedia.org/wiki/Extended_ASCII) but it's not clear for me, which code-page will be used, i.e. which characters will be translated. Strictly speaking the converting must depend on the used encoding (e.g.  Latin-1, Windows-1252, etc.).
It might be fine to get more details information about the converting.

The converting for Unicode to ASCII basically works fine but there are also some strange (ugly) results. E.g. [guillemets](https://en.wikipedia.org/wiki/Guillemet) ("»" and "«") will be translated to ">>" and "<<" (double greater-than or less-than sign).

It might be fine to have:

- more details documentation how the converting is done,
- more specific option for the converting (e.g. set the code-page/character set) and
- have the option to include and/or exclude characters individually (e.g. guillemet).

Nero-A:
I know/saw the plugin [Non-ASCII Equivalents](https://github.com/metabrainz/picard-plugins/tree/2.0/plugins/non_ascii_equivalents) but it's not the solution for me.

phw:
Please report issues regarding Picard to https://tickets.metabrainz.org/projects/PICARD/

Having said this I think the option is simply not named properly. The internal name for this is `convert_punctuation`, and that is what it actually does: It converts various forms of punctuation characters to their more generic forms as available on let's say keyboards. This is mostly for people who prefer this or need it for better compatibility (e.g. some devices might not have the fonts to display all these characters).

What this is not about is encoding, hence it does not make sense to have a encoding setting here. Encoding is a topic dealt with when saving to files, and Picard will usually work with Unicode except for ID3 where you can choose the encoding.

PR #240, opened 2019-09-20
Create re-convert_guillemets.py

Nero-A:
[re-]change metadata: change "<<" and ">>" to '«' and '»' (guillemets).
Workaround for my problem with replaced guillemets, see Issue #238


plugins/re-convert/re-convert.py
```diff
@@ -0,0 +1,98 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+PLUGIN_NAME = "re-convert"
```
Sophist-UK: This title is too generic. It needs to be called something like "Convert guillemots" in order to explain to users what it does.

Nero-A: I don't want to limit it to guillemots only. Maybe there will be more characters in future.
But maybe it's not worth to publish it.

Sophist-UK: Better for it to do something simple - and be well named.

Nero-A:
"Codacy/PR Quality Review — Not up to standards."
Why?
It follows the examples of plugins (https://picard.musicbrainz.org/docs/plugin-api/)

phw:
Code-wise this looks fine for me. But I wonder if we need another simple conversion plugin like this which operates on a limited set of fields for a very specific replacement. This basically can be done by a simple script like:

```
$set(album,$replace(%album%,<<,«))
$set(album,$replace(%album%,>>,»))
$set(title,$replace(%title%,<<,«))
$set(title,$replace(%title%,>>,»))
```

Now I get it, this is a lot of repetitive code, and once you want to do a lot of replacements on a lot of tags this gets pretty cumbersome. 

But I feel having a single plugin for every possible conversion does also not scale well. We already have some plugins that IMHO are too specific, as they do some not too complex conversion for only a specific set of hard coded tags.

This plugin is totally fine as a plugin that solves your specific problem, I have myself created occasionally plugins I never shared because they are too specific and solved a specific thing I was interested in, sometimes only for tagging a few files. This is really part of the power of Picard to allow users to solve their issues. Maybe this would be better suited as an example on https://community.metabrainz.org/ for others to see how to solve issues like this?

What I would find interesting is a plugin where the user can define custom replacements and the tags they are applied to in a small UI.

Sophist-UK:
> Code-wise this looks fine for me. But I wonder if we need another simple conversion plugin like this which operates on a limited set of fields for a very specific replacement. This basically can be done by a simple script like:
...
> What I would find interesting is a plugin where the user can define custom replacements and the tags they are applied to in a small UI.

Now that Philipp has raised this I tend to agree with him - this functionality can be done by a script and a more general Replace plugin with an Options / Plugins / Replace UI page where users can defined the From / To strings and the fields to operate on would be very useful.

phw:
@zas Any opinion on this? My opinion is that a plugins with a hard coded set of replacements that does this replacement for a hard coded set of tags is useful for a specific use case. But we can fill up the plugin list with an endless amount of such plugins.

We already have replace_forbidden_symbols, non_ascii_equivalents and mayve others with a very similar functionality.

In many cases I think it would be more useful to have a script function doing these changes, instead of applying them to a hard coded list of tags. Or a plugin like this, but where you can configure the replacements and the tags on which they should be performed.

zas:
> In many cases I think it would be more useful to have a script function doing these changes, instead of applying them to a hard coded list of tags. Or a plugin like this, but where you can configure the replacements and the tags on which they should be performed.

Yes, something more generic would be better, I'd prefer a script function.

PR #243, opened 2019-10-13
Add additional_artists_variables to main Picard plugins repository

rdswift:
Finally got around to submitting this plugin to the official list.

This plugin provides specialized album and track variables for use in naming scripts. It is based on the "Album Artist Extension" plugin, but expands the functionality to also include track artists. Note that it
cannot be used as a direct drop-in replacement for the "Album Artist Extension" plugin because the variables are provided with different names.

If this is accepted, I wonder if we should remove the "Album Artist Extension" plugin?

### DIFF
--- a/plugins/additional_artists_variables/additional_artists_variables.py
+++ b/plugins/additional_artists_variables/additional_artists_variables.py
@@ -31,7 +31,7 @@ this plugin.
 Please see the <a href="https://github.com/rdswift/picard-plugins/blob/2.0_RDS_Plugins/plugins/additional_artists_variables/docs/README.md">user guide</a> on GitHub for more information.
 '''
 PLUGIN_VERSION = '0.5'
-PLUGIN_API_VERSIONS = ['2.0']
+PLUGIN_API_VERSIONS = ['2.0', '2.1', '2.2']
 PLUGIN_LICENSE = 'GPL-2.0-or-later'
 PLUGIN_LICENSE_URL = 'https://www.gnu.org/licenses/gpl-2.0.html'
 
@@ -150,19 +150,16 @@ def process_artists(album_id, source_metadata, destination_metadata, source_type
         destination_metadata['~artists_{0}_all_sort_primary'.format(source_type,)] = sort_pri_artist
     if artist_count:
         destination_metadata['~artists_{0}_all_count'.format(source_type,)] = artist_count
-    return None
 
 
 def make_album_vars(album, album_metadata, release_metadata):
     album_id = release_metadata['id'] if release_metadata else 'No Album ID'
     process_artists(album_id, release_metadata, album_metadata, 'album')
-    return None
 
 
 def make_track_vars(album, album_metadata, track_metadata, release_metadata):
     album_id = release_metadata['id'] if release_metadata else 'No Album ID'
     process_artists(album_id, track_metadata, album_metadata, 'track')
-    return None
 
 
 def metadata_error(album_id, metadata_element, metadata_group):
--- a/plugins/additional_artists_variables/additional_artists_variables.py
+++ b/plugins/additional_artists_variables/additional_artists_variables.py
@@ -19,6 +19,7 @@
 
 PLUGIN_NAME = 'Additional Artists Variables'
 PLUGIN_AUTHOR = 'Bob Swift (rdswift)'
+PLUGIN_USER_GUIDE_URL = 'https://github.com/rdswift/picard-plugins/blob/2.0_RDS_Plugins/plugins/additional_artists_variables/docs/README.md'
 PLUGIN_DESCRIPTION = '''
 This plugin provides specialized album and track variables for use in
 naming scripts. It is based on the "Album Artist Extension" plugin, but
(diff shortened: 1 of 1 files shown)


rdswift:
I tried to apply a `.format()` to the `PLUGIN_DESCRIPTION` string so that I only needed to enter the link to the user guide once.  This seemed to work fine during testing here, but caused a build error in one of the tests.  See [Travis Build Job 615.1](https://travis-ci.org/metabrainz/picard-plugins/jobs/597865539) for details.  I assume that this is because of the way that the information is extracted from the file for display on the website.  Is this a constraint that should be documented for plugin developers?

plugins/additional_artists_variables/additional_artists_variables.py
```diff
@@ -0,0 +1,177 @@
...
+expands the functionality to also include track artists. Note that it
+cannot be used as a direct drop-in replacement for the "Album Artist
+Extension" plugin because the variables are provided with different
+names.  This will require changes to existing scripts if switching to
+this plugin.
+<br /><br />
+Please see the <a href="https://github.com/rdswift/picard-plugins/blob/2.0_RDS_Plugins/plugins/additional_artists_variables/docs/README.md">user guide</a> on GitHub for more information.
+'''
+PLUGIN_VERSION = '0.5'
+PLUGIN_API_VERSIONS = ['2.0']
```
phw: Best change this to `PLUGIN_API_VERSIONS = ['2.0', '2.1', '2.2']`

Should we drop API version '2.0' for some reason in a new release this plugin will continue to work as it also claims compatibility with '2.1' and '2.2'. And since there is nothing special here this would be fine.

plugins/additional_artists_variables/additional_artists_variables.py
```diff
@@ -0,0 +1,177 @@
...
+        destination_metadata['~artists_{0}_all_sort'.format(source_type,)] = sort_artist
+    if std_artist_list:
+        destination_metadata['~artists_{0}_all_std_multi'.format(source_type,)] = std_artist_list
+    if cred_artist_list:
+        destination_metadata['~artists_{0}_all_cred_multi'.format(source_type,)] = cred_artist_list
+    if sort_pri_artist:
+        destination_metadata['~artists_{0}_all_sort_primary'.format(source_type,)] = sort_pri_artist
+    if artist_count:
+        destination_metadata['~artists_{0}_all_count'.format(source_type,)] = artist_count
+    return None
```
phw: `return None` is not really needed here. For one the return values are never used, and also in  `var = process_artists(...)` the `var` variable would be `None` if there is not return statement at all.

Same is true of course for the two functions below.

plugins/additional_artists_variables/additional_artists_variables.py
```diff
@@ -0,0 +1,177 @@
...
+this plugin.
+<br /><br />
+Please see the <a href="https://github.com/rdswift/picard-plugins/blob/2.0_RDS_Plugins/plugins/additional_artists_variables/docs/README.md">user guide</a> on GitHub for more information.
+'''
+PLUGIN_VERSION = '0.5'
+PLUGIN_API_VERSIONS = ['2.0']
+PLUGIN_LICENSE = 'GPL-2.0-or-later'
+PLUGIN_LICENSE_URL = 'https://www.gnu.org/licenses/gpl-2.0.html'
+
+PLUGIN_USER_GUIDE_URL = 'https://github.com/rdswift/picard-plugins/blob/2.0_RDS_Plugins/plugins/additional_artists_variables/docs/README.md'
```
phw, line 38: I like that. Could also be very useful for complex plugin like classical extras. We could add a help button / link to the plugins.

phw (commented):
Thanks a lot. Just minor nitpicking from my side :)

plugins/additional_artists_variables/additional_artists_variables.py
rdswift: That never occurred to me.  Thanks.

plugins/additional_artists_variables/additional_artists_variables.py
rdswift: I have no idea what I what I was thinking there.

plugins/additional_artists_variables/additional_artists_variables.py
rdswift: That's sort of what I had in mind.  I know there's a plan to upgrade the plugin system at some point, and I was thinking this might be a way to provide information for a 'Help' or 'User Guide' link from both the Picard Plugins web site as well as from the description from the plugins window in Picard itself.

Classical extras would be a perfect use case.

rdswift:
> Thanks a lot. Just minor nitpicking from my side :)

I very much appreciate your 'nitpicking' comments.  I learn a bit more from each one, and they help me improve my code, both for this case and in the future.  Thank-YOU.

rdswift:
This plugin basically replaces the **AlbumArtist Extension** plugin, but adds some additional variables and extends it to track artists.  Unfortunately, it provides the information using different variable names.  Do you think there would be any benefit to adding the variable names from **AlbumArtist Extension** to make this extension backwards compatible with it, and then retiring the **AlbumArtist Extension** plugin?

plugins/additional_artists_variables/additional_artists_variables.py
phw: https://tickets.metabrainz.org/browse/PICARD-1649

phw (approved):

phw:
I think I would rather keep it simple and consistent inside this plugin, it is easy enough to switch from one variables to the others. What about adding a section to the documentation explaining the migration?

rdswift:
> What about adding a section to the documentation explaining the migration?

That makes sense to me.  I'll do that.  Thanks.

PR #246, opened 2019-11-25
A typo and a new language for the decode_cyrillic

NeNpO:


NeNpO:
Should i make the commits separately?

plugins/bpm/__init__.py
```diff
@@ -17,7 +17,7 @@
 # PLUGIN_INCOMPATIBLE_PLATFORMS = [
 #    'win32', 'cygwyn', 'darwin', 'os2', 'os2emx', 'riscos', 'atheos']
 
-from aubio import source, tempo
+from audio import source, tempo
```
phw: This actually is not a typo, the module is called [aubio](https://pypi.org/project/aubio/). Please drop this commit from the pull request ;)

plugins/decode_greek_cyrillic/decode_greek1253.py
```diff
@@ -0,0 +1,106 @@
...
+#   instead, try the common variations, and show a dialog to the user,
+#   allowing him to select the correct transcoding. See 2cyr.com for example.
+# - also see http://stackoverflow.com/questions/23326531/how-to-decode-cp1252-string
+
+class DecodeGreek(BaseAction):
+    NAME = "Unmangle Greek metadata"
+
+    def unmangle(self, tag, value):
+        try:
+            print(value, value.encode('latin1'))
```
phw: Please drop this `print` statement or convert it to a `log.debug` message (but with some proper output, e.g. `log.debug("%s: %s => %r", PLUGIN_NAME, value, value.encode('latin1'))`)

I know this is in the original plugin, but it should be fixed there as well ;)

plugins/decode_greek_cyrillic/decode_greek1253.py
```diff
@@ -0,0 +1,106 @@
...
+
+            for tag in _decode_tags:
+                if not (tag in cluster.metadata):
+                    continue
+
+                cluster.metadata[tag] = self.unmangle(tag, cluster.metadata[tag])
+
+            log.debug("cluster name is %s by %s" % (cluster.metadata['album'], cluster.metadata['albumartist']))
+
+            for _name, file in enumerate(cluster.files):
```
phw: Why `_name`? Due to enumerate this actually is just an integer index. But I think there is no need to use enumerate here at all, `for file in cluster.files:` should be just fine.

plugins/decode_greek_cyrillic/decode_greek1253.py
```diff
@@ -0,0 +1,106 @@
...
+
+_decode_tags = [
+    'title',
+    'albumartist',
+    'artist',
+    'album',
+    'artistsort'
+]
+# _from_encoding = "latin1"
+# _to_encoding   = "cp1251"
```
phw: If these are unused they could be removed.

plugins/decode_greek_cyrillic/decode_greek1253.py
```diff
@@ -0,0 +1,106 @@
...
+from picard.cluster import Cluster
+from picard.ui.itemviews import BaseAction, register_cluster_action
+
+_decode_tags = [
+    'title',
+    'albumartist',
+    'artist',
+    'album',
+    'artistsort'
+]
```
phw, line 52: There is artist and artistsort here, but only albumartist. I think this should at least also list albumartistsort.

plugins/decode_greek_cyrillic/decode_greek1253.py
```diff
@@ -0,0 +1,106 @@
...
+# _from_encoding = "latin1"
+# _to_encoding   = "cp1251"
+
+
+# TODO:
+# - extend to support multiple codepage decoding, not just cp1251->latin1
+#   instead, try the common variations, and show a dialog to the user,
+#   allowing him to select the correct transcoding. See 2cyr.com for example.
```
phw: The above comment does not fully apply to this case. Remove?

phw (changes requested):
A few comments, most of which would also apply to decode_cyrillic. I think we can get this cleaned up there and then take the changes over to decode_cyrillic as well.

I just kind of wonder if it would make sense to combine both plugins, e.g. having one plugin (e.g. "fix_misencoding") that offers both cyrillic and greek encoding fixes? But I'm fine with merging this as a separate plugin once it is cleaned up.

plugins/bpm/__init__.py
NeNpO: Sorry about that i didn't know.
i just saw the following error and thought that was a typo. I should researched first.

```
> "bpm/__init__.py", line 20, in <module>
>     from aubio import source, tempo
> ModuleNotFoundError: No module named 'aubio'"
```

NeNpO:
First of all i am a noob with GitHub and as of the code i'm still learning.

> A few comments, most of which would also apply to decode_cyrillic. I think we can get this cleaned up there and then take the changes over to decode_cyrillic as well.

This is possible and i'm more than happy to do it.
 

> I just kind of wonder if it would make sense to combine both plugins, e.g. having one plugin (e.g. "fix_misencoding") that offers both cyrillic and greek encoding fixes? But I'm fine with merging this as a separate plugin once it is cleaned up.

I was toying with the idea to make it to auto recognize the encoding of the characters.
So after that i will try to make the fix_missencoding plugin but maybe i will start from the simple stuff and just ask the user about the encoding.

plugins/decode_greek_cyrillic/decode_greek1253.py
NeNpO: Ok, i tested it and it seems to work like you said.

plugins/decode_greek_cyrillic/decode_greek1253.py
NeNpO: Done

plugins/decode_greek_cyrillic/decode_greek1253.py
NeNpO: Also added  "comment:" and "comment:ID3v1 Comment".

NeNpO:
Ok I think that I corrected all the problems that you mentioned and now I am working on to add the same function on files.

phw (approved):
Thanks a lot

NeNpO:
I should thank you for your hard work.
So thank you!
If you need any help just let me know.

PR #253, opened 2020-03-03
TheAudioDB: Updates to API key

phw:
Update the TheAudioDB with various changes:

1. New API key (they have issued new keys)
2. Remove rate limiting
3. Add some notes about contributing and their patreon campaign to the description in the plugin's settings.

See also discussion at https://www.theaudiodb.com/forum-new/viewtopic.php?f=6&t=14818

### DIFF
--- a/plugins/theaudiodb/__init__.py
+++ b/plugins/theaudiodb/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2015-2019 Philipp Wolfer
+# Copyright (c) 2015-2020 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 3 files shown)


zag2me:
Looks good to me, let me know if Picard could use any new API features?

We support loads of different artwork types including 3D covers now which could be cool in the future if the tagging could support them somehow.

Example 3D Album cover:
qqstxu1558126687

Example glossy Album thumb:
xrxuxu1558126677

phw:
@zag2me Thanks for the feedback. We'll let you know. Current state is that cover art in Picard is still a bit lacking, there are a lot of improvements we could / should do. So for now the basic function is sufficient.

Btw, the register link on the homepage (in the text above the Patreon link) is currently broken.

zas (approved):

PR #254, opened 2020-04-22
Make test.py a real unittest

mineo:
It's not too much work and it's always nicer to use a standard test setup.

Also, this creates a `test` folder that other test can be put into in the
future.

Also, this creates `test/test_doctest.py` that keeps the list of doctests to run.


mineo:
See #255

Issue #256, opened 2020-04-27
ValueError Exception in Format Performer Tags


lonemadmax:
From https://community.metabrainz.org/t/picard-plugin-format-performer-tags/402265/15
```
__init__.py, line 69, in rewrite_tag
   instruments.remove(instrument)
ValueError: list.remove(x): x not in list
```
(rest of the traceback in an image in the forum)

Four examples are given:
https://musicbrainz.org/release/6fa0ff27-1b59-3e67-9e73-213193e6e360
https://musicbrainz.org/release/357298ea-eaa2-49c8-9559-2f48019081b4
https://musicbrainz.org/release/3cca7500-d82c-4ad6-aa57-f27107a9cc72
https://musicbrainz.org/release/3f06ade0-d2ef-4cca-aae0-6254de73d96b

Keep in mind that I'm not the reporter and I haven't had that issue (yet). I just stumbled upon that thread when looking for info on some plugins, I have not even run any file through Picard yet! That and I don't know how the MB db works.

Looking at the code, I don't get what https://github.com/metabrainz/picard-plugins/blob/1d8784b9488e47102518b96392faffd062ceb058/plugins/format_performer_tags/__init__.py#L64-L69 might be trying to do. If I understand this correctly (big if) `subkey` may have something like 'background vocals and drums', that would give `instruments = ['background vocals', 'drums']` and then that is searched for guest/solo/additional items? But that last search is what https://github.com/metabrainz/picard-plugins/blob/1d8784b9488e47102518b96392faffd062ceb058/plugins/format_performer_tags/__init__.py#L72-L75 does. Even if we had 'guitar and solo guitar', we would get `instruments = ['guitar', 'solo guitar']`, with 'solo' not being an item.

One thing the four releases given have in common is at least one track with 'guest performer' and no instrument (at least as shown in web). If that is reflected as `subkey = 'guest'`, we would do something in the loop, and that something would be raise the exception: `instruments = ['guest']`, then for whatever reason https://github.com/metabrainz/picard-plugins/blob/1d8784b9488e47102518b96392faffd062ceb058/plugins/format_performer_tags/__init__.py#L68 does `instruments[0] = "guest guest"`, and the next line tries to remove the value just overwritten.

rdswift:
@lonemadmax 

Thanks for your work in tracking down the problem.  I've added a check in the code to make sure that the instrument exists in the list before trying to remove it.  PR #257

rdswift:
[...]

PR #257, opened 2020-04-28
Issue #256: ValueError Exception in Format Performer Tags

rdswift:
Test that instrument is in the list before trying to remove it.  Set `subkey` to "unspecified" if it is one of the types (i.e.: "guest", "solo" or "additional")

### DIFF
--- a/plugins/format_performer_tags/__init__.py
+++ b/plugins/format_performer_tags/__init__.py
@@ -46,6 +46,7 @@ from picard.plugins.format_performer_tags.ui_options_format_performer_tags impor
 performers_split = re.compile(r", | and ").split
 
 WORD_LIST = ['guest', 'solo', 'additional']
+UNSPECIFIED_INSTRUMENT = 'unspecified'
 
 
 def get_word_dict(settings):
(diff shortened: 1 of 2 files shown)


plugins/format_performer_tags/__init__.py
```diff
@@ -59,14 +59,17 @@ def rewrite_tag(key, values, metadata, word_dict, settings):
     mainkey, subkey = key.split(':', 1)
     if not subkey:
         return
+    if subkey in WORD_LIST:
+        subkey = 'unspecified'
```
phw: Shouldn't this be `if subkey not in WORD_LIST`? Or do I misunderstand the intention?

phw (changes requested):

plugins/format_performer_tags/__init__.py
rdswift: I don't think so.  It was intended to avoid the situation when there is no instrument provided but the relationship shows as one or more of the qualifiers from `WORD_LIST` so that the qualifier isn't considered to be the instrument.  In all honesty, I'm not 100% sure that situation would ever come up.

rdswift: For example, look at the performer relationships on the first track of https://musicbrainz.org/release/357298ea-eaa2-49c8-9559-2f48019081b4 which shows a "Guest Performer" (no instrument specified) for Cliff Richard.  This triggers the "Unspecified" instrument.

phw: So for a case where it just would have a value of `performer:guest`, right? Thanks for clarifying. Not sure if it can happen, but it is no mistake to take care of this.

phw (approved):

plugins/format_performer_tags/__init__.py
rdswift: Although that example uncovered another problem, in that it drops the "Guest" qualifier.  I'll have to look at that a bit closer.

rdswift: I need to make one small change to make it display properly.  I'll upload the revision right away.

rdswift: My "one small change" wasn't as small as I thought.  (Are they ever?)  It now properly handles cases where there is no performer type specified, combining then under a type of "unspecified".

rdswift: I moved this PR back to "Draft" status because I think there's a better way of handling this without having to use the "unspecified" instrument tag.

rdswift: I ended up cleaning up the word parsing code and removed the need for adding the "unspecified" instrument type.  Sorry for all the changes making it harder to review.  I tested it with all four releases listed in Issue #256 and it processed them all properly.

phw (approved):

PR #260, opened 2020-06-03
Indentation problem

jas01:
Problem with the indentation so calling

      mvt = title_split[1]

  in all case would cause a exit error when the size of the split are only 1 and
  i'm guessing the test colon >1 are to prevent that


MetaTunes:
This is superseded by #260

phw:
#261 has been merged instead, closing

PR #262, opened 2020-06-24
Add Search Engine Lookup plugin

rdswift:
This PR is to add the Search Engine Lookup plugin to the list of Picard Plugins available.

### DIFF
--- a/plugins/search_engine_lookup/__init__.py
+++ b/plugins/search_engine_lookup/__init__.py
@@ -27,13 +27,11 @@ PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
 
 from urllib.parse import quote_plus
 
-from PyQt5.QtWidgets import QMessageBox
-
 from picard import config, log
 from picard.cluster import Cluster
 from picard.plugins.search_engine_lookup.ui_options_search_engine_lookup import Ui_SearchEngineLookupOptionsPage
 from picard.ui.itemviews import BaseAction, register_cluster_action
-from picard.ui.mainwindow import MainWindow
+from picard.ui.mainwindow import MainWindow, QtWidgets
 from picard.ui.options import OptionsPage, register_options_page
 from picard.util.webbrowser2 import open as _open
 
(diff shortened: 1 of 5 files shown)


plugins/search_engine_lookup/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+            if isinstance(cluster, Cluster):
+                parts = []
+                if 'albumartist' in cluster.metadata and cluster.metadata['albumartist']:
+                    parts.extend(str(cluster.metadata['albumartist']).split())
+                if 'album' in cluster.metadata and cluster.metadata['albumartist']:
+                    parts.extend(str(cluster.metadata['album']).split())
+                if parts:
+                    if config.setting["search_engine_lookup_extra_words"]:
+                        parts.extend(str(config.setting["search_engine_lookup_extra_words"]).split())
+                    url = ENGINES[config.setting["search_engine_lookup_provider"]] + quote_plus(' '.join(parts))
```
zas, line 59: I'd move `config.setting[*]' outside of the loop, storing values in local variables.

plugins/search_engine_lookup/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+        # Settings for search engine provider
+        temp = config.setting["search_engine_lookup_provider"]
+        if temp not in ENGINES:
+            temp = 'Google'
+        if temp == 'Bing':
+            self.ui.rb_bing.setChecked(True)
+        elif temp == 'DuckDuckGo':
+            self.ui.rb_duckduckgo.setChecked(True)
+        else:
+            self.ui.rb_google.setChecked(True)
```
zas, line 110: Generating radio buttions would permit to make code more generic and remove hardcoded references to search engines. `ENGINES` could be used (and extended).

zas (approved):
LGTM.

phw (approved):

plugins/search_engine_lookup/__init__.py
rdswift: These are both excellent suggestions.  I'll chip away at them when I have time over the next couple of weeks and submit a pull request with the changes.  Do you see storing the engines in a persistent dictionary (which can be edited from the ui), and then building the radio buttons from there?

PR #264, opened 2020-07-19
Search Engine Lookup: Add provider add, edit, delete functions.  Redesign config UI.

rdswift:
Building on the comments provided by the reviewers of the original Search Engine Lookup plugin, this update adds the ability for the user to manage the list of search engine providers (add, edit and remove).  The configuration UI has been redesigned to accommodate the additional functionality.  This also adds one more key to the settings dictionary to hold the list of providers.

image

Example of editing an existing provider.  The "Title" box is read-only.  Note that both the Title and URL pass the validation tests (showing a checkmark to the right of the text boxes) thus the "Save" button is enabled.

image

Example of adding a new provider.  The title passes validation (two or more characters and different from all existing providers), but the URL validation fails because it does not contain `%search%`.  Thus the 'X' and the "Save" button is disabled.

image

Example of adding a new provider.  The title fails because it is the same as an existing provider, thus the "X" and the "Save" button is disabled.  The URL passes validation.

image

Example of adding a new provider.  Both the title and URL pass the validations, thus both are shown with a checkmark and the "Save" button is enabled.

image

Example with the new provider saved and added to the list.

image

Confirmation dialog displayed when deleting a provider.

image

Error dialog displayed when trying to delete the currently selected (checked) provider.

image


rdswift:
I did not include a confirmation dialog box before resetting to the "factory defaults" because I thought that the user could always abort the changes by not clicking the "Make It So!" button.  Any thoughts on this?  Should I include a confirmation dialog for the sake of consistency?

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
-        if temp not in ENGINES:
-            temp = 'Google'
-        if temp == 'Bing':
-            self.ui.rb_bing.setChecked(True)
-        elif temp == 'DuckDuckGo':
-            self.ui.rb_duckduckgo.setChecked(True)
-        else:
-            self.ui.rb_google.setChecked(True)
+        self.provider = config.setting[KEY_PROVIDER]
+        if self.provider not in self.providers.keys():
```
zas: `.keys()` isn't needed with a dict (it defaults to use keys).

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
+            self.provider = list(self.providers.keys())[0]
```
zas: Not sure what you are trying to do, but it's unlikely to work (it may work ... but....)
Order of dict keys isn't guaranteed before Python 3.7.

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
+        self.ui.pb_delete.clicked.connect(self.delete_provider)
+        self.ui.pb_test.clicked.connect(self.test_provider)
+        self.ui.pb_reset.clicked.connect(self.reset_settings)
 
     def load(self):
-        # Enable external link
-        self.ui.format_description.setOpenExternalLinks(True)
+        # Settings for search engine providers
+        self.providers = {}
+        self.fill_providers_dict_from_list(config.setting[KEY_PROVIDERS])
```
zas: Since `fill_providers_dict_from_list()` is doing `self.providers = {}` previous line can be removed.

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
+        self.additional_words = config.setting[KEY_EXTRA]
+        self.ui.le_additional_words.setText(self.additional_words)
+
+        self.update_list()
+        self.setup_actions()
+
+    def select_provider(self, list_item):
+        if list_item.checkState() == QtCore.Qt.Checked:
+            # New provider selected
+            self.provider = list_item.text()
```
zas: It's always a bad idea to use the item text directly, because you may want a different label at some point while you want to store a unique identifier in config.
Example: Google (label) -> google (identifier)
So if Google becomes Gogole, you can easily change the label, and change nothing in load/save/config.

See https://doc.qt.io/qt-5/qlistwidgetitem.html#data

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
+        self.edit_provider = edit_provider
+        self.edit_url = edit_url
+        self.add_flag = not edit_provider
+
+        provider_data = config.setting[KEY_PROVIDERS]
+        if not provider_data:
+            provider_data = DEFAULT_PROVIDERS
+        self.providers = []
+        for item in provider_data:
+            self.providers.append(item.split('|')[0])
```
zas: Why not use a dict or named tuple?

```python
providers = [
  {'label': 'Google', 'searchquery': 'https://www.google.com/search?q=%search%', 'id': 'google'},
  ...
]
```
It would be much better than splitting strings etc...

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
+        parts = ('The Beatles Abby Road ' + self.additional_words).strip().split()
+        url = self.providers[current_item.text()].replace(r'%search%', quote_plus(' '.join(parts)))
+        _open(url)
+
+    def fill_providers_dict_from_list(self, provider_list):
+        self.providers = {}
+        if not provider_list:
+            provider_list = DEFAULT_PROVIDERS
+        for provider_record in provider_list:
+            provider_name, provider_url = provider_record.split('|', 2)
```
zas: Once again, I don't see the point of splitting strings when the language provides structured data.

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
+            self.providers[provider_name] = provider_url
+
+    def update_list(self, current_item=None):
+        current_row = -1
+        self.ui.list_providers.clear()
+        provider_list = sorted(self.providers.keys())
```
zas: You should sort on labels for display, but do not use labels as identifiers.
If you keep this, `.keys()` does nothing: `sorted(self.providers)` will sort keys of the dict.

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
+        for counter, provider_key in enumerate(provider_list):
+            item = QtWidgets.QListWidgetItem()
+            item.setFlags(item.flags() | QtCore.Qt.ItemIsUserCheckable)
+            item.setText(provider_key)
```
zas: I guess `provider_key` is supposed to be unique and permanent, but here you use it as label (which may change or be translated).
Note that I insist on this, because it will bite you at some point, but code is ok otherwise.

plugins/search_engine_lookup/__init__.py
```diff
@@ -78,56 +91,226 @@ def show_popup(title, content):
...
 
-        # Process search engine provider settings
-        if self.ui.rb_bing.isChecked():
-            settings["search_engine_lookup_provider"] = 'Bing'
-        elif self.ui.rb_duckduckgo.isChecked():
-            settings["search_engine_lookup_provider"] = 'DuckDuckGo'
+        if self.providers:
+            temp = []
+            for key in self.providers.keys():
+                temp.append('{0}|{1}'.format(key, self.providers[key],))
```
zas: Again this format... ;)

zas (commented):
UI-wise, much better.
Code-wise, few comments ;)

I should review the whole plugin, and I may have misunderstood few things, I'll re-review later.

plugins/search_engine_lookup/__init__.py
rdswift: I keep forgetting that.  I'll fix it.  Thanks.

plugins/search_engine_lookup/__init__.py
rdswift: The only thing I was doing was to ensure that there was always one (existing / valid) item selected.  This arbitrarily selected the first item in the list if there were no items selected.

plugins/search_engine_lookup/__init__.py
rdswift: Good catch.  Thanks.

plugins/search_engine_lookup/__init__.py
rdswift: Okay, I'll rework this part.

plugins/search_engine_lookup/__init__.py
rdswift: I had it all originally coded using a dictionary, but ran into problems trying to save it to the settings and restore it later.  I'm sure that it works to save/restore a list of tuples to the settings, so I'll do that.

rdswift: Oops, I was thinking of something else.  In this case, I was just getting a list of the providers (keys) to check that the same provider name was not used for a new provider.  A dictionary wasn't really necessary for the lookup.  There are better ways of doing this, so I'll change it.

plugins/search_engine_lookup/__init__.py
zas: >  This arbitrarily selected the first item in the list if there were no items selected.

Ok then it works ;)

BTW, using `.keys()` is totally correct, but dicts are returning keys when iterate, so `list(d)` or `list(d.keys())` will produce same output (but last one is a bit slower). Since 3.5, we can use another way, but it's new and not well-known yet: `[*d]`.
This is the fastest and less readable. For readibility, `.keys()` is imho the best (but it is also the slowest).
Too many ways to do something in Python ;)

```python
>>> d = {'c':3, 'a': 1, 'b': 2}
>>> list(d)
['c', 'a', 'b']
>>> list(d.keys())
['c', 'a', 'b']
>>> [*d]
['c', 'a', 'b']
>>> import timeit
>>> timeit.timeit('[*d]', setup="d = {'c':3, 'a': 1, 'b': 2}", number=10000000)
0.609008648985764
>>> timeit.timeit('[*d]', setup="d = {'c':3, 'a': 1, 'b': 2}", number=10000000)
0.6115269220026676
>>> timeit.timeit('list(d)', setup="d = {'c':3, 'a': 1, 'b': 2}", number=10000000)
1.0752794349973556
>>> timeit.timeit('list(d)', setup="d = {'c':3, 'a': 1, 'b': 2}", number=10000000)
1.0832242340256926
>>> timeit.timeit('list(d)', setup="d = {'c':3, 'a': 1, 'b': 2}", number=10000000)
1.0784355590003543
>>> timeit.timeit('list(d.keys())', setup="d = {'c':3, 'a': 1, 'b': 2}", number=10000000)
1.4061465539853089
>>> timeit.timeit('list(d.keys())', setup="d = {'c':3, 'a': 1, 'b': 2}", number=10000000)
1.403430454985937
```

plugins/search_engine_lookup/__init__.py
zas: >  ran into problems trying to save it to the settings and restore it later.

Weird, please share what you found, current config stuff is able to save dicts afaik, but you may have hit some bug.

rdswift: It may simply be that I was using the wrong option class for the dictionary.  As I read it, the available classes (from picard.config) are: TextOption, BoolOption, IntOption, FloatOption and ListOption.  I tried saving the dictionary as a ListOption.  Qt didn't like that.  What class should I be using for the dictionary?

rdswift: Saving the dictionary directly would eliminate quite a bit of code.

zas: >  What class should I be using for the dictionary?

Just Option.
See https://github.com/metabrainz/picard/blob/dd2acaf6ee1128ee11a23fcba559cc480530e956/test/test_config.py#L332 for an example with a `set` (but it works with `dict`)
Or https://github.com/metabrainz/picard/blob/dd2acaf6ee1128ee11a23fcba559cc480530e956/picard/ui/options/interface_colors.py#L91 (with a dict)

rdswift: THANK-YOU!

plugins/search_engine_lookup/__init__.py
```diff
@@ -214,102 +197,100 @@ def load(self):
...
                 QtWidgets.QMessageBox.Ok,
                 QtWidgets.QMessageBox.Ok
             )
         else:
             if QtWidgets.QMessageBox.warning(
                 self,
-                'Confirm Deletion',
-                'You are about to permanently delete the search provider "{0}".  Continue?'.format(provider,),
+                _('Confirm Deletion'),
+                _('You are about to permanently delete the search provider "') + provider + _('".  Continue?'),
```
zas: From i18n point of view, the previous form was much better (read: concatenating doesn't play well with translations).
And the best is to use named placeholders:

```python
_('You are about to permanently delete the search provider "{provider_name}".  Continue?').format(provider_name=provider)
```

plugins/search_engine_lookup/__init__.py
```diff
@@ -59,7 +61,10 @@ class SearchEngineLookupTest(BaseAction):
 
     def callback(self, cluster_list):
         extra = config.setting[KEY_EXTRA].split()
-        base_url = get_selected_url()
+        if config.setting[KEY_PROVIDER] in config.setting[KEY_PROVIDERS]:
+            base_url = config.setting[KEY_PROVIDERS][config.setting[KEY_PROVIDER]]['url']
```
zas: Perhaps more readable if written:

```python
provider = config.setting[KEY_PROVIDER]
providers = config.setting[KEY_PROVIDERS]
if provider in providers:
   base_url = providers[provider]['url']
```

plugins/search_engine_lookup/__init__.py
```diff
@@ -125,10 +113,8 @@ def __init__(self, parent=None, edit_provider='', edit_url=''):
         self.ui.setupUi(self)
         self.ui.le_title.setText(self.edit_provider)
         self.ui.le_url.setText(self.edit_url)
-        if not self.add_flag:
-            self.ui.le_title.setReadOnly(True)
 
-        self.setWindowTitle('{0} Search Engine Provider'.format('Add' if self.add_flag else 'Edit',))
+        self.setWindowTitle((_('Add') if not edit_provider else _('Edit')) + _(' Search Engine Provider'))
```
zas: This is a bad practice when it comes to i18n (hint: do not concatenate).

For translation it's much better to have full phrases.

With marking first (`N_()`), then translation (`_()`):

```python
if edit_provider:
   window_title = N_('Edit Search Engine Provider')
else:
   window_title = N_('Add Search Engine Provider')
self.setWindowTitle(_(window_title))
```

Or you can simplify (in this case):

```python
if edit_provider:
   self.setWindowTitle(_('Edit Search Engine Provider'))
else:
   self.setWindowTitle(_('Add Search Engine Provider'))
```

plugins/search_engine_lookup/__init__.py
```diff
@@ -139,10 +125,7 @@ def setup_actions(self):
         self.ui.pb_cancel.clicked.connect(self.reject)
 
     def check_validation(self):
-        if self.add_flag:
-            valid_title = bool(re.match(RE_VALIDATE_TITLE, self.edit_provider)) and bool(self.edit_provider not in self.providers)
-        else:
-            valid_title = True
+        valid_title = bool(re.match(RE_VALIDATE_TITLE, self.edit_provider)) and bool(self.edit_provider not in self.providers)
```
zas: In Python, `bool()` isn't needed (usually).

plugins/search_engine_lookup/__init__.py
```diff
@@ -214,102 +197,100 @@ def load(self):
...
-
-    def edit_provider_dialog(self, provider='', url=''):
-        dialog = SearchEngineEditDialog(self, provider, url)
+        provider_id = current_item.data(QtCore.Qt.UserRole)
+        url = self.providers[provider_id]['url']
+        self.edit_provider_dialog(provider_id, provider, url)
+
+    def edit_provider_dialog(self, provider_id='', provider='', url=''):
+        titles = [x['name'] for x in self.providers.values()]
+        titles.remove(provider)
```
zas: I think we can do it in one pass using `if` in list comprehension:

```python
titles = [x['name'] for x in self.providers.values() if x['name'] != provider]
 ```

zas (changes requested):
Good work, code is much simpler, few more comments but overall looks good to me.

phw:
> I did not include a confirmation dialog box before resetting to the "factory defaults" because I thought that the user could always abort the changes by not clicking the "Make It So!" button. Any thoughts on this? Should I include a confirmation dialog for the sake of consistency?

This is about the "Reset" button below the list, right? I think it is not really needed, there is not much lost and as you said the user can cancel. Is there a big need for the "Reset" button at all? It does not do much less then the already existing "Restore defaults" button.

rdswift:
> This is about the "Reset" button below the list, right? I think it is not really needed, there is not much lost and as you said the user can cancel. Is there a big need for the "Reset" button at all? It does not do much less then the already existing "Restore defaults" button.

It is totally redundant and I'll remove it.  I never used the "Restore defaults" button before so didn't realize what it did.

plugins/search_engine_lookup/__init__.py
rdswift: Okay, I'm (slowly) beginning to understand the translation process.  Thanks.

plugins/search_engine_lookup/__init__.py
rdswift: Definitely.  Thanks.

plugins/search_engine_lookup/__init__.py
rdswift: Actually, there's no good reason to change the window title at all, so I'm removing this part entirely.  Good explanation of the i18n stuff though.  Thanks.

plugins/search_engine_lookup/__init__.py
rdswift: I found that I still needed it in one spot to keep Qt from crashing, but I inserted a comment explaining why it's there.

plugins/search_engine_lookup/__init__.py
rdswift: Worked perfectly.  Thanks.

zas (approved):

phw (approved):

PR #267, opened 2020-08-18
Add help URLs to some plugin option pages

phw:
Those plugins by @rdswift provide some extended documentation, I think it makes sense to link to it.

Classical extras is another candidate, but since it is mostly developed in @MetaTunes's repository I'd rather have this submitted to upstream and merge this later.


zas (approved):
Lgtm

PR #268, opened 2020-09-01
theaudiodb: Add option to use high resolution images

phw:
Ads an option to use high resolution images for cover art from TheAudioDB.

One example with such cover art available is:

https://www.theaudiodb.com/album/2349438
https://musicbrainz.org/release-group/2587b239-ca88-47ec-907c-33d0b22ff915

### DIFF
--- a/plugins/theaudiodb/__init__.py
+++ b/plugins/theaudiodb/__init__.py
@@ -20,7 +20,7 @@
 PLUGIN_NAME = 'TheAudioDB cover art'
 PLUGIN_AUTHOR = 'Philipp Wolfer'
 PLUGIN_DESCRIPTION = 'Use cover art from TheAudioDB.'
-PLUGIN_VERSION = "1.1.1"
+PLUGIN_VERSION = "1.2"
 PLUGIN_API_VERSIONS = ["2.0", "2.1", "2.2", "2.3", "2.4"]
 PLUGIN_LICENSE = "GPL-2.0-or-later"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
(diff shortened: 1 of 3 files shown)


zag2me:
Cool thanks, I was thinking about making this some kind of premium feature in the future but for now its all good :) Hopefully it will encourage more to upload 4K covers.
Bandwidth is quite expensive...

phw:
@zag2me I also couldn't find much about it, after long digging I found one example (but I already lost the URL again). But let me know if I should restrict this in the plugin or nag people more. It has not yet been released to the official plugin downloads :D

Sophist-UK:
In my experience a significant proportion of people who want 4k covers will generally upload if they find one isn't there - so IMO making it possible / easy to tag your music with 4K covers is likely to encourage uploads as you suggest. That said, I think that TheAudioDB charging for access to 4K pictures will negate much of this, and I would prefer to see MB offering / encouraging 4K album art via e.g. Cover Art Archive.

So if it were me, I would:

a. Put something on the Album Art tab for Releases on MB to either show the 4K art if MB already has it or to encourage people to upload it if they have it themselves.

b. If a Picard user has enabled 4K cover art, then put something in Picard to encourage / help people to find and upload the 4K cover art to MB.

Bandwidth for 4K covers might be significant when you are bulk-downloading the covers whilst tagging, but still relatively small by comparison with e.g. video streaming (even at 720p).

I would be more concerned about having this enabled by default when (I am guessing that) the majority of people won't want 4K pictures (at least not for a few years anyway).

zag2me:
No need to restrict for now, I'm happy to provide this feature and let it grow. 

As @Sophist-UK  says maybe it should be "opt in" check box to control the growth.

I only mentioned it as I need to keep an eye on the bandwidth as it may not be sustainable in the long term 👍  We already serve terabytes a day.

PR #273, opened 2020-11-25
Add Deezerart plugin

livingsilver94:
This PR adds Deezerart to the plugin collection. Part of the plugin development has been discussed in [this thread](https://community.metabrainz.org/t/i-developed-deezerart/499526).


plugins/deezerart/__init__.py
```diff
@@ -0,0 +1,160 @@
+PLUGIN_NAME = "Deezer cover art"
+PLUGIN_AUTHOR = "Fabio Forni <livingsilver94>"
+PLUGIN_DESCRIPTION = "Fetch cover arts from Deezer"
+PLUGIN_VERSION = '1.1.0'
+PLUGIN_API_VERSIONS = ['2.2', '2.3', '2.4', '2.5']
```
Sophist-UK: Can we fix the redirect issue and limit this to 2.5 and above?

plugins/deezerart/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class OptionsPage(providers.ProviderOptions):
+    NAME = 'Deezer'
+    TITLE = 'Deezer'
+    options = [config.TextOption('setting', 'deezerart_size', obj.CoverSize.BIG.value)]
+    _options_ui = Ui_Form
+
+    def load(self):
+        for s in obj.CoverSize:
```
Sophist-UK, line 43: For the UI, to be compatible with Cover Art Archive options page, can we display the cover art sizes in px rather than the Deezer Small / Medium / Big or XL.

plugins/deezerart/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+
+__version__ = PLUGIN_VERSION
+
+
+def redirected_url(url: str) -> str:
+    """
+    Find where a URL is redirecting to.
+    """
+    parsed = urlsplit(url)
+    conn = http.HTTPSConnection(parsed.netloc, 443) if parsed.scheme == 'https' else http.HTTPConnection(parsed.netloc, 80)
```
Sophist-UK: Can we not use the QT network functions for this and avoid having to support two different methods?

Can you use the existing Picard network routines to make this call?

This is particularly important if this code is synchronous and blocking as it appears to be because it does not provide a call-back.

plugins/deezerart/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+                return
+            artist = self._artist()
+            album = self.metadata['album']
+            for result in results:
+                if not isinstance(result, obj.Track):
+                    continue
+                if not is_similar(artist, result.artist.name) or not is_similar(album, result.album.title):
+                    continue
+                cover_url = result.album.cover_url(obj.CoverSize(config.setting['deezerart_size']))
+                if picard.PICARD_VERSION < (2, 5):
```
Sophist-UK: As previously suggested, can we not just restrict this plugin to be used by Picard 2.5 or above?

plugins/deezerart/deezer/__init__.py
```diff
@@ -0,0 +1 @@
+from .client import Client, SearchOptions
```
Sophist-UK, line 1: I am not sure whether it is beneficial or not to split the code for this plugin across so many files. Can
```
./deezer/__init__.py
./deezer/client.py
./deezer/obj.py
```
not simply be included in the main file?

Or is this something specific to do with this being a cover art plugin?

plugins/deezerart/deezer/obj.py
```diff
@@ -0,0 +1,95 @@
...
+    Cover size selector.
+    """
+    THUMBNAIL = 'small'
+    SMALL = ''
+    MEDIUM = 'medium'
+    BIG = 'big'
+    LARGE = 'xl'
+
+
+class Album(Object):
```
Sophist-UK: Since we don't appear to be using namespaces, is there any conflict in use of e.g. an Album class with any other similarly named classes in Picard itself or other plugins?

plugins/deezerart/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+    def _queue_from_url(self, album: obj.Object, error: QtNet.QNetworkReply.NetworkError):
+        self.album._requests -= 1
+        try:
+            if error:
+                self.error('could not get Deezer API object: {}'.format(error))
+                return
+            if not isinstance(album, obj.Album):
+                self.error('API object is not an album')
+                return
+            cover_url = album.cover_url(obj.CoverSize(config.setting['deezerart_size']))
```
Sophist-UK, line 101: What if you want XL covers but they don't exist?

Do we want code to fall back to Big in this event etc.?

plugins/deezerart/deezer/obj.py
```diff
@@ -0,0 +1,95 @@
+import enum
+import json
+from typing import Any, List, Mapping, Optional, Union
+
+# This module is a perfect target for dataclasses.
+# Picard though supports Python 3.5+, so we cannot
+# use dataclasses for compatibility reasons.
+
+
+class Object:
```
Sophist-UK: The class Object is already the root class in Python!!!!

Sophist-UK (changes requested):
Sorry to have so many comments, but I have some general comments also:

1. Would it be better to handle the Deezer lookup slightly differently in multiple steps:

a. If it doesn't already exist in Album metadata, search for the Deezer Album ID and store it in Album metadata.

b. Retrieve the album data from Deezer including the cover art URLs.

c. Retrieve the cover art itself.

For compatibility with similar functionality elsewhere, we might want an option for storing the metadata and another for deleting it if it already exists.

2. The Deezer API talks about authentication. Obviously this plugin does not use personal data, so I cannot see that it is necessary to authenticate just to get cover art, but I just wanted to check that this wasn't necessary?

plugins/deezerart/__init__.py
livingsilver94: Sure!

plugins/deezerart/__init__.py
livingsilver94: I could but I cannot guarantee the label -> pixel association is stable. The Deezer docs only talk about the label: https://developers.deezer.com/api/album

plugins/deezerart/__init__.py
livingsilver94: This is irrelevant. If you want me to only support Picard 2.5+, this piece of code will be gone.

plugins/deezerart/__init__.py
livingsilver94: Again, sure!

plugins/deezerart/deezer/__init__.py
livingsilver94: This is for readability and good coding style. Nothing more, nothing less. Why are you against splitting the codebase into multiple files?

plugins/deezerart/deezer/obj.py
livingsilver94: You may not be using namespaces but this is `deezerart/deezer` module. It's an inner module located in a plugin. I think it's safe to say it'll never be used anywhere else but Deezerart.

plugins/deezerart/__init__.py
livingsilver94: A fallback would actually be useful but the Deezer docs don't say some cover sizes may be unavailable. Anyway, I can implement a fallback logic to play it safe, yeah.

plugins/deezerart/deezer/obj.py
livingsilver94: So? Again, this is the `deezerart/deezer` module, an internal one. Also, namespaces are a good practice to use.

plugins/deezerart/__init__.py
phw: Yes, I would also be in favor of limiting this to 2.5 and later.

We could also refactor the workaround to use the webservice module and be non-blocking, but it complicates things and I don't feel like it is really worth it.

plugins/deezerart/__init__.py
phw: I'd stay with the label, I think it is better if Deezer does not clearly communicate the sizes.

plugins/deezerart/deezer/obj.py
phw: Yes, `deezerart.deezer.obj.Album` is clearly distinct from `picard.album.Album`. Actually they are namespaced by module. Issues only arrive if you import same named objects into a modules local namespace, but that's not even happening here outside of deezerart.deezer and if can be cleanly handled anyway.

plugins/deezerart/deezer/obj.py
phw: But I'd still avoid naming the base class `Object` here, as this directly overwrites Python's `Object` in this module. It's unlikely to cause an issue in this case, but things like this are often the cause for subtle bugs. E.g. later down that file you want to use Python's `Object` for something real, and it then uses the local `Object` instead. So maybe rename it to something else, e.g. `Entity`.

plugins/deezerart/deezer/__init__.py
phw: I actually think it is beneficial to have the deezer specific stuff in it's own module or package. Not sure whether it needs to be split further, e.g. the extra obj.py for the model classes is a bit a matter of taste. But I really don't mind having it split like this. I personally would have named the file `model.py`, but yeah, no strong opinion either way.

plugins/deezerart/deezer/obj.py
Sophist-UK: I am prepared to bet that this Object or Entity is not completely generic and has something a bit more specific that sets it apart - in which case I am sure that a better name can be found.

plugins/deezerart/deezer/obj.py
Sophist-UK: How about DeezerAlbum?

phw: You really don't need to have this extensive prefixing of names in Python. I know it is quite common in some other languages to prefix every module and data type, but in Python you have everything namespaced by module names already.

Also this is a data type only used locally in the deezerart.deezer package. It is usually good practice to keep more locally used names short, more widely used names more verbose.

If it would really be used in a separate module and you import it into your local module namespace you can easily avoid name conflicts with something like `from deezerart.deezer import Album as DeezerAlbum` or `import deezerart.deezer as deezer`, then use `deezer.Album`.

Sophist-UK: I realise it is unnecessary BUT it is an object which represents the Album data held by Deezer, and not an Album as loaded by Picard. So calling it DeezerAlbum would help distinguish the content for maintainers.

livingsilver94:
@Sophist-UK r.e.

> The Deezer API talks about authentication. Obviously this plugin does not use personal data, so I cannot see that it is necessary to authenticate just to get cover art, but I just wanted to check that this wasn't necessary?

Deezer's API documentation states, in the introduction:

>Unlimited Access, without stress, without identification. Deezer Simple API provides a nice set of services to build up web applications allowing the discovery of Deezer's music catalogue.

plugins/deezerart/__init__.py
livingsilver94: I double checked the docs here (IDK if you have a Deezer account to see the page): https://developers.deezer.com/api/album

All `cover_*` fields are listed as non-optional. There's no note anywhere that hints they may be absent, so I'd just say that if an XL cover doesn't "exist", Deezer upscales smaller covers to bigger sizes (and no, there's no way I could know if it's original sized or upscaled).

Sophist-UK: I cannot see anywhere which says which fields are mandatory or optional - and URL could be empty. Further research needed  to find examples.

livingsilver94: I contacted the Deezer support but they replied they don't have the information I requested. I'll try to get in touch with Deezer devs in some other way.

phw: Thanks.I would be fine with merging this for now as is. If there is new information on this the plugin can be extended, but if it for now looks like all sizes are always available let's go with this.

For the TheAudioDB I chose to fall back to lower resolution, see #268. But there the situation is very different, high resolution images are a real exception there (for now). So without this fallback this plugin would not be well usable.

phw (approved):

PR #280, opened 2020-11-25
[Standardise_Performers] Sort performers to avoid false changes

Sophist-UK:
The main intent of this update is to sort Standardised Performers so that the order of performers remains the same even if the data is split in a different sequence, thus avoiding false track metadata changes when the source MB metadata is essentially the same.

In addition, I have taken the opportunity to remove code that is specific to Picard v1.

### DIFF
--- a/plugins/standardise_performers/standardise_performers.py
+++ b/plugins/standardise_performers/standardise_performers.py
@@ -21,9 +21,10 @@ Performer [slide guitar]: Eric Stewart
 Performer [synthesizer]: Lol Creme
 Performer [tambourine]: Graham Gouldman
 </pre>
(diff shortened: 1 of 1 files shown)


Sophist-UK:
As an example:
```
Performer [acoustic guitar, bass, dobro, electric guitar and tambourine]: Graham Gouldman
Performer [acoustic guitar, electric guitar, grand piano and synthesizer]: Lol Creme
```
might result in either:
```
Performer [acoustic guitar]: Graham Gouldman; Lol Creme
Performer [acoustic guitar]: Lol Creme; Graham Gouldman
```
depending on the order that the original Performer tags were processed (which is **not** a defined order e.g. alphabetical). Sorting the contents of each Performer tag will ensure future consistency.

phw (approved):

PR #281, opened 2020-11-26
Sort homepages for consistency

Sophist-UK:
Some artists e.g. Andrew Lloyd Webber have multiple Artist Homepages, and in this event the order of these can be random, making metadata change when source MB data is the same.

This fix sorts the results in order to ensure consistency.


Sophist-UK:
Amy Winehouse is another example.

phw (approved):

Issue #282, opened 2020-11-26
[acousticbrainz_tonal-rhythm] Rate limit issues


Sophist-UK:
When loading albums I have been receiving error messages in the log like:
```
E: 23:37:19,087 webservice\__init__._handle_reply:445: Network request error for http://acousticbrainz.org:80/api/v1/high-level?recording_ids=f2ada1c9-fa39-4d15-aa11-9bd798a885a9&map_classes=true: Error transferring http://acousticbrainz.org:80/api/v1/high-level?recording_ids=f2ada1c9-fa39-4d15-aa11-9bd798a885a9&map_classes=true - server replied: TOO MANY REQUESTS (QT code 299, HTTP code 429)
E: 23:37:19,134 webservice\__init__._handle_reply:445: Network request error for http://acousticbrainz.org:80/f2ada1c9-fa39-4d15-aa11-9bd798a885a9/low-level: Error transferring http://acousticbrainz.org:80/f2ada1c9-fa39-4d15-aa11-9bd798a885a9/low-level - server replied: TOO MANY REQUESTS (QT code 299, HTTP code 429)
```
Looking at the code for this plugin, the rate limit appears to be hard coded whereas the [AcousticBrainz API manual](https://acousticbrainz.readthedocs.io/api.html#rate-limiting) seems to suggest that rate limiting is variable and is reported in the response headers.

IMO this plugin needs to be enhanced to process these response headers and feed them back to the webservice rate limiting functionality. In the mean time I will put forward a small PR which will increase the rate limit from 50ms to 100ms in order to try to avoid these sorts of errors (at the cost of slower lookups).

phw:
The updated acousticbrainz plugin will replace the tonal rhythm plugin, as it integrates the functionality of it. It also somewhat mitigates the rate limiting issue as it is doing less requests overall (it batches requests for mulltiple recordings of one release, so ideally it ends up with one request per release).

It is no full support for the dynamic rate limiting of the AB server, though. Proper support for this would best be added to Picard's web service module, though. See PICARD-1681

phw:
Closing this as this now is working well by the acousticbrainz plugin, which does less requests per album.

PR #283, opened 2020-11-26
[acousticbrainz_tonal-rhythm] Increase rate delay

Sophist-UK:
Avoid getting rate limit errors with AcousticBrainz:
```
E: 23:37:19,087 webservice\__init__._handle_reply:445: Network request error for http://acousticbrainz.org:80/api/v1/high-level?recording_ids=f2ada1c9-fa39-4d15-aa11-9bd798a885a9&map_classes=true: Error transferring http://acousticbrainz.org:80/api/v1/high-level?recording_ids=f2ada1c9-fa39-4d15-aa11-9bd798a885a9&map_classes=true - server replied: TOO MANY REQUESTS (QT code 299, HTTP code 429)
E: 23:37:19,134 webservice\__init__._handle_reply:445: Network request error for http://acousticbrainz.org:80/f2ada1c9-fa39-4d15-aa11-9bd798a885a9/low-level: Error transferring http://acousticbrainz.org:80/f2ada1c9-fa39-4d15-aa11-9bd798a885a9/low-level - server replied: TOO MANY REQUESTS (QT code 299, HTTP code 429)
```
by increasing the hardcoded delay time from 50ms to 100ms.

See #282 for a better suggested approach following the acousticbrainz API rate limiting model.

### DIFF
--- a/plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
+++ b/plugins/acousticbrainz_tonal-rhythm/acousticbrainz_tonal-rhythm.py
@@ -27,7 +27,7 @@ from the AcousticBrainz database.<br/><br/>
 '''
 PLUGIN_LICENSE = "GPL-2.0-or-later"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
-PLUGIN_VERSION = '1.1.3'
+PLUGIN_VERSION = '1.1.4'
 PLUGIN_API_VERSIONS = ["2.0"]  # Requires support for TKEY which is in 1.4
 
 from json import JSONDecodeError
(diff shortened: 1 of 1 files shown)


Sophist-UK:
I just tested this and it seems to have stopped the errors occurring.

Since my acousticbrainz calls take c. 80ms, and it was getting errors with a rate limit of 50ms (i.e. next call was effectively immediate), it looks like an increase to 100ms is only a 20% decrease in rate rather than 50% one, whilst the scope for tweaking this to find a slightly lower value is far lower - consequently I do not propose to try to optimise this value further.

phw:
Thanks, I guess this works for a fix for now.

About the dynamic rate limit see also PICARD-1681 . I think it would make sense to have this implemented as part of the networking code. But maybe it needs some configuration, as servers might behave a little different. zas already cautioned that e.g. the rate limit headers of mb.org are considered experimental, so we probably should stick with the fixed limit there for now. Maybe we could refactor the rate limiting into separate classes, which can then be set per request.

phw (approved):

PR #288, opened 2020-12-01
[albumartist_website] Fix Picard crash if artist has no urls

Sophist-UK:


Sophist-UK:
I never quite got to grips as to why it was crashing.

Sophist-UK:
And I have not found a single clue as to why the test builds failed - no explanation just a statement that they failed.

phw (approved):
Github Actions was having issues at the time with their runners.

PR #290, opened 2021-01-12
Fix and improve ReplayGain

Sophist-UK:
1. Make it work at all on Py3.
2. Capture output to Picard debug log.
3. Improve the description and provide details of where to get needed executables. 

Although it runs the commands correctly, there are some issues documented in the description.

There is further work that can be done, however Philipp and I agreed to a major update along completely different lines, so this version is provided as an interim warts-and-all version.

### DIFF
--- a/plugins/replaygain/__init__.py
+++ b/plugins/replaygain/__init__.py
@@ -1,8 +1,55 @@
 # -*- coding: utf-8 -*-
 PLUGIN_NAME = "ReplayGain"
-PLUGIN_AUTHOR = "Philipp Wolfer"
-PLUGIN_DESCRIPTION = """Calculate ReplayGain for selected files and albums."""
-PLUGIN_VERSION = "0.1"
+PLUGIN_AUTHOR = "Philipp Wolfer, Sophist"
+PLUGIN_DESCRIPTION = """<h2>Calculate ReplayGain for selected files and albums.</h2>
+<p>
+This plugin calculates the ReplayGain values for Albums and / or Tracks.
+These values are stored in tags and music players use this information
+to adjust the played volume so that all tracks have a similar volume
+and you don't get tracks that you can barely hear and need to turn up the volume for
+followed by tracks with a mucg louder volume so that you are then deafened.
+</p><p>
+This plugin needs separate external executables to be run to calculate the replay gains.
+You need to download these executables and then configure the ReplayGain plugin in
+Options / Plugins / ReplayGain with the path and filename of the executable.
+</p><p>
+As executables, they are probably best stored in a directory which is normally read-only for normal users
+and requires administrative write access to store the executable (e.g. the Picard executable directory),
+but you can store them in the plugins directory if you wish.
+</p><hr/>
+<p>Whilst this version works, it has some issues:</p>
+<ul>
+<li>It doesn't work with file paths / names containing unicode characters</li>
+<li>On Windows it opens an empty window whilst processing</li>
+<li>Tags are stored directly in the files, are not displayed in Picard until the files are reloaded and will be overwritten if Picard saves the files</li>
+<li>You need to add the replay gain tags to Options / Tags / Preserve these tags from being overwritten with MusicBrainz data (because Musicbrainz does not provide this data)</li>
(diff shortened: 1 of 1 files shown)


phw (approved):
I think that works as an intermediate update for getting this back into a somewhat working sate. Thanks for looking into it.

plugins/replaygain/__init__.py
```diff
@@ -1,16 +1,64 @@
...
+As executables, they are probably best stored in a directory which is normally read-only for normal users
+and requires administrative write access to store the executable (e.g. the Picard executable directory),
+but you can store them in the plugins directory if you wish.
+</p><hr/>
+<p>Whilst this version works, it has some issues:</p>
+<ul>
+<li>It doesn't work with file paths / names containing unicode characters</li>
+<li>On Windows it opens an empty window whilst processing</li>
+<li>Tags are stored directly in the files, are not displayed in Picard until the files are reloaded and will be overwritten if Picard saves the files</li>
+<li>You need to add the replay gain tags to Options / Tags / Preserve these tags from being overwritten with MusicBrainz data (because Musicbrainz does not provide this data)</li>
```
mineo, line 25: super minor super late nitpick: This is not true if Clear existing tags is not enabled (which is the default).

plugins/replaygain/__init__.py
```diff
@@ -1,16 +1,64 @@
...
+You need to download these executables and then configure the ReplayGain plugin in
+Options / Plugins / ReplayGain with the path and filename of the executable.
+</p><p>
+As executables, they are probably best stored in a directory which is normally read-only for normal users
+and requires administrative write access to store the executable (e.g. the Picard executable directory),
+but you can store them in the plugins directory if you wish.
+</p><hr/>
+<p>Whilst this version works, it has some issues:</p>
+<ul>
+<li>It doesn't work with file paths / names containing unicode characters</li>
```
mineo, line 22: Is this a Windows-only restriction? It works just fine on https://musicbrainz.org/release/871f12ec-6259-4e93-b440-9108559c38b1.

plugins/replaygain/__init__.py
```diff
@@ -39,9 +87,17 @@ def calculate_replay_gain_for_files(files, format_, tagger):
...
-                         decode_filename(b' '.join(file_list)))
-        check_call([command] + options + file_list)
+        call = [command] + options + file_list
+        tagger.log.debug(call)
+        with subprocess.Popen(call, stdout=subprocess.PIPE, stderr=subprocess.STDOUT) as process:
+            (output, null) = process.communicate()
+            output = output.splitlines()
+            for line in output:
+                if line:
+                    tagger.log.debug(line)
```
mineo, line 97: It would be great if the lines were collected and logged differently depending on `rc`. I'd expect error messages to be logged on level error.

mineo (commented):
> Make it work at all on Py3.

I'm confused - I've been using the replaygain plugin (the .zip file I have is from feb 2018). What didn't work?

Sophist-UK:
> I'm confused - I've been using the replaygain plugin (the .zip file I have is from feb 2018). What didn't work?

@mineo The line that was broken was:
```
        tagger.log.debug('%s %s %s', command, ' '.join(options),
                         decode_filename(b' '.join(file_list)))
 ```
Removing the b specifier for the string made it work again. I then started to sort out other issues, and was part way through when @phw suggested that we should make this work entirely differently. I cannot now recall where that conversation was - if Philipp can remember perhaps he can add a link.

plugins/replaygain/__init__.py
Sophist-UK: If we weren't going to replace this with something significantly better, I would do this. It is easy to get the rc and change logging accordingly.

plugins/replaygain/__init__.py
Sophist-UK: Yes - it is a windows issue.

phw:
> I cannot now recall where that conversation was - if Philipp can remember perhaps he can add a link.

Good question, I think it was on the forums. It also had been discussed in the past on IRC at least twice :)

Sophist-UK:
@mineo Wieland - The conversation was [here](https://community.metabrainz.org/t/help-with-replaygain-plugin/446469).

PR #291, opened 2021-01-19
"ReplayGain" instead of "gain"

Sophist-UK:
"gain" is incorrect terminology and this change corrects this and makes it clear that these actions are from this plugin.

### DIFF
--- a/plugins/replaygain/__init__.py
+++ b/plugins/replaygain/__init__.py
@@ -103,7 +103,7 @@ def calculate_replay_gain_for_files(files, format_, tagger):
 
 
 class ReplayGain(BaseAction):
-    NAME = N_("Calculate Track Replay&Gain...")
+    NAME = N_("Calculate track Replay&Gain...")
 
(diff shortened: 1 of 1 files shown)


plugins/replaygain/__init__.py
```diff
@@ -103,7 +103,7 @@ def calculate_replay_gain_for_files(files, format_, tagger):
 
 
 class ReplayGain(BaseAction):
-    NAME = N_("Calculate replay &gain...")
+    NAME = N_("Calculate Track Replay&Gain...")
```
phw: Let's have this consistent and either lower case "track" here or uppercase "album" in the other string. I'm in favor of "Calculate track Replay&Gain..."

phw (changes requested):
Just minor consistency change, otherwise LGTM

plugins/replaygain/__init__.py
Sophist-UK: Done.

phw (approved):

PR #293, opened 2021-03-07
happidev_lyrics: Convert former apiseeds_lyrics to Happi.dev API

phw:
Apiseed is shutting down and gets integrated into [Happi.Dev](https://happi.dev). If you are logged in into the Apiseeds website it shows this message:

> We are migrating to Happi.Dev. Apiseeds will be shut down soon. If you want use our services please go to Happi.Dev.

Happi.dev provides a Music API with lyrics search. As I underdstand it the API seeds data might get integrated there, but the API itself is a bit different. Hence this PR builds upon the Apiseeds Lyrics plugin, but adopts the code to match the new API. In particular this now requires two requests to fetch the lyrics.

See:
- PICARD-2112
- https://community.metabrainz.org/t/apiseeds-lyrics-offline/518636


plugins/happidev_lyrics/happidev_lyrics.py
```diff
@@ -0,0 +1,171 @@
...
+            album._finalize_loading(None)
+            return
+
+        try:
+            lyrics_url = response['result'][0]['api_lyrics']
+            log.debug('{}: lyrics found for track {} at {}'.format(
+                PLUGIN_NAME, metadata['title'], lyrics_url))
+            path = urlparse(lyrics_url).path
+            album._requests += 1
+            self._request(album.tagger.webservice, path,
```
Sophist-UK, line 95: I recommend that lyrics requests are put at the front FIFO of the non-priority queue so that when you load an album (or several albums) the lyrics requests are not queued up behind the search requests.

plugins/happidev_lyrics/happidev_lyrics.py
```diff
@@ -0,0 +1,171 @@
...
+            log.debug('{}: {}'.format(PLUGIN_NAME, error))
+            callback(None, None, error)
+            return
+
+        queryargs['apikey'] = apikey
+        ws.get(self.happidev_host, self.happidev_port, path, callback,
+            parse_response_type='json', priority=True, queryargs=queryargs)
+
+    def process_search_response(self, album, metadata, response, reply, error):
+        if error or (response and (not response.get('success', False) or not response.get('length', 0))):
```
Sophist-UK, line 82: To save half the requests if you Reload an album, I recommend caching the search responses (which are small - but not the lyrics responses which are large).

plugins/happidev_lyrics/happidev_lyrics.py
```diff
@@ -0,0 +1,171 @@
...
+        super().__init__()
+        ratecontrol.set_minimum_delay(
+            (self.happidev_host, self.happidev_port), self.happidev_delay)
+
+    def process_metadata(self, album, metadata, track, release):
+
+        artist = metadata['artist']
+        if not artist:
+            log.debug(
+                '{}: artist is missing, please provide a valid value'.format(PLUGIN_NAME))
```
Sophist-UK: "please provide a valid value"???

How about "artist is required to obtain happi.dev lyrics"?

plugins/happidev_lyrics/happidev_lyrics.py
```diff
@@ -0,0 +1,171 @@
...
+            return
+
+        title = metadata['title']
+        if not title:
+            log.debug(
+                '{}: title is missing, please provide a valid value'.format(PLUGIN_NAME))
```
Sophist-UK: "please provide a valid value"???

How about "title is required to obtain happi.dev lyrics"?

plugins/happidev_lyrics/happidev_lyrics.py
```diff
@@ -0,0 +1,171 @@
...
+from picard.webservice import ratecontrol
+
+PLUGIN_NAME = 'Happi.dev Lyrics'
+PLUGIN_AUTHOR = 'Andrea Avallone, Philipp Wolfer'
+PLUGIN_DESCRIPTION = 'Fetch lyrics from Happi.dev Lyrics, which provides millions of lyrics from artist all around the world. ' \
+                     'Lyrics provided are for educational purposes and personal use only. Commercial use is not allowed.<br /><br />' \
+                     'In order to use Happi.dev you need to get a free API key at <a href="https://happi.dev">happi.dev</a>'
+PLUGIN_VERSION = '2.0'
+PLUGIN_API_VERSIONS = ['2.0', '2.1', '2.2', '2.3', '2.4', '2.5', '2.6']
+PLUGIN_LICENSE = 'MIT'
```
Sophist-UK, line 30: I don't have an issue with MIT license, however since this is a rewrite replacement not an edit, IMO we can change to a different license if we wish e.g. GPL.

Sophist-UK (changes requested):
Nice to see it only takes 2 API calls per track - I was worried that it might take 3 or more.

plugins/happidev_lyrics/happidev_lyrics.py
phw: Not sure I follow. Would the suggestion be to queue the lyrics requests with `important=True` (which would put them at the top of the queue)?

plugins/happidev_lyrics/happidev_lyrics.py
phw: I'd keep the license. I did basically only take the apiseeds lyrics and ran a search and replace over it, then modified the request / response handling as necessary.

plugins/happidev_lyrics/happidev_lyrics.py
Sophist-UK: Yes - `important=True` for lyric requests, `important=False` for search requests.

plugins/happidev_lyrics/happidev_lyrics.py
Sophist-UK: Or even better:
```
        artist = metadata['artist']
        title = metadata['title']
        if not (artist and title):
            log.debug(
                '{}: both artist and title are required to obtain happi.dev lyrics'.format(PLUGIN_NAME))
            return
```
as this avoid a situation where they get a message about needing artist, provide the artist and then get a second error about needing title.

plugins/happidev_lyrics/happidev_lyrics.py
```diff
@@ -0,0 +1,171 @@
...
+                PLUGIN_NAME, metadata['title']))
+            album._requests -= 1
+            album._finalize_loading(None)
+            return
+
+        try:
+            lyrics = response['result']['lyrics']
+            metadata['lyrics'] = lyrics
+            log.debug('{}: lyrics loaded for track {}'.format(
+                PLUGIN_NAME, metadata['title']))
```
Sophist-UK: Since both artist and title are required, how about:
```
            log.debug('{}: lyrics loaded for track {} by {}, '.format(
                PLUGIN_NAME, metadata['title'], metadata['artist']))
```

phw: Yes makes sense.

Technically artist might not be required, but I think I'd keep it as it is now. For the Apiseeds API it was clear, because both artist and title where a required parameter for the API endpoint.

But with the Happi.dev API it is different, as we use a search endpoint. The structured parameters only tell the endpoint to return tracks with lyrics, but otherwise there is just a freeform search field (the q parameter). There is no documentation on how this works, but I found searching for `"ARTIST" "TITLE"`  seemed to work reasonably well.

Technically that means just the title would also work. But looking for artist + title at least seems to be a bit more reliable.

plugins/happidev_lyrics/happidev_lyrics.py
phw: Yes, let's change this.

But technically this shouldn't happen, if the data is not available at the time the plugin runs something has gone wrong. I think these comments had been there more for debugging than for anything else.

plugins/happidev_lyrics/happidev_lyrics.py
Sophist-UK: I suspect that there are many examples of completely different songs with the same title by different artists.

I think it is far less likely that an artist has two completely different songs with the same title.

Given that it is reasonable to expect Picard's MB based metadata to include both artist and title, I think it is absolutely the correct thing to do to require both to avoid false positives.

plugins/happidev_lyrics/happidev_lyrics.py
Sophist-UK: hmmm. I think you are probably right that the plugin only runs after successful download of data from MB, and MB should never have a situation where a track doesn't have an Artist or Title. But Sods Law says that there will be an example at some point.

plugins/happidev_lyrics/happidev_lyrics.py
phw: Addressed in b0ff1c874e16e1abb424693148032b4477bb0d62

plugins/happidev_lyrics/happidev_lyrics.py
phw: I improved the logging to include the artist in 0def93a7a9a4c7761d67cae19aac4074143979dd

plugins/happidev_lyrics/happidev_lyrics.py
phw: Done in 0def93a7a9a4c7761d67cae19aac4074143979dd

plugins/happidev_lyrics/happidev_lyrics.py
phw: I'll merge this now as is as I want to have a working plugin deployed and my main concern is to replace apiseeds with a functionally equivalent version for happi.dev, otherwise we will receive continuous issue reports about apiseeds ceasing to work.

Someone can address caching the results later.

Sophist-UK: Please feel free to plagiarise my caching code in albumartistwebsite plugin.

PR #294, opened 2021-03-08
Add primary album artist tags as a multi-value variable

rdswift:
Update the **Additional Artists Variables** plugin to provide a multi-value variable that lists the top 5 tags assigned to the primary album artist for a release.  Only tags with a rating higher than zero are included.

### DIFF
--- a/plugins/additional_artists_variables/additional_artists_variables.py
+++ b/plugins/additional_artists_variables/additional_artists_variables.py
@@ -30,12 +30,14 @@ this plugin.
 <br /><br />
 Please see the <a href="https://github.com/rdswift/picard-plugins/blob/2.0_RDS_Plugins/plugins/additional_artists_variables/docs/README.md">user guide</a> on GitHub for more information.
 '''
-PLUGIN_VERSION = '0.5'
+PLUGIN_VERSION = '0.6'
 PLUGIN_API_VERSIONS = ['2.0', '2.1', '2.2']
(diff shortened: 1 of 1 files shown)


plugins/additional_artists_variables/additional_artists_variables.py
```diff
@@ -91,6 +93,12 @@ def process_artists(album_id, source_metadata, destination_metadata, source_type
                     temp_sort_name = artist_credit['artist']['sort-name']
                 else:
                     metadata_error(album_id, 'artist-credit.artist.sort-name', source_type)
+                tag_list = []
+                if 'tags' in artist_credit['artist']:
```
phw: It's actually a bit more complicated. You can have different fields here, possible are `genres`, `tags`, `user-genres` and `user-tags`. In theory all 4 could be present, but Picard only requests one of them from the server. Which one depends on the setting:

- If the setting `only_my_genres` is active, the `user-*` variants are used
- If the setting `folksonomy_tags` ("Use folksonomy tags as genre") is active, then `tags` or `user-tags` is used, otherwise it is `genres` or `user-genres`

See also https://github.com/metabrainz/picard/blob/master/picard/dataobj.py#L48

rdswift: **THANK-YOU**.  That's the part I couldn't find when I was browsing the Picard code.

I'll update the plugin to try to accommodate all conditions.  Stay tuned...  :wink:

phw (approved):

PR #295, opened 2021-03-09
Add Data Dumper Plugin

rdswift:
Add a utility plugin to dump the `album` and `track` metadata provided to plugins to a text file.  This is intended to help plugin developers by showing the information provided as arguments to their plugins.


phw:
Not sure about this really. I see it's value both as an example and possible debug tool. But having this in the official plugin list I kind of fear users being confused by it or expecting it to do something it is not supposed to do.

rdswift:
Good point.  Perhaps this is something better referenced in the Picard User Guide if and when I actually add a tutorial about developing plugins.  I'll cancel this PR but leave the plugin available from my clone of the repository.

PR #296, opened 2021-03-15
PICARD-2141: AcousticBrainz Mood-Genre: Save existing genres before overwriting

Sophist-UK:
Save existing genre metadata (i.e. MB genres if "Use genres from MusicBrainz" is enabled) so that scripts can put it back or merge genre values.

### DIFF
--- a/plugins/acousticbrainz/acousticbrainz.py
+++ b/plugins/acousticbrainz/acousticbrainz.py
@@ -19,11 +19,12 @@
 PLUGIN_NAME = 'AcousticBrainz Mood-Genre'
 PLUGIN_AUTHOR = 'Andrew Cook, Sambhav Kothari'
 PLUGIN_DESCRIPTION = '''Uses AcousticBrainz for mood and genre.
+Any existing genres saved in %_mbgenre%.
 
 WARNING: Experimental plugin. All guarantees voided by use.'''
 PLUGIN_LICENSE = "GPL-2.0"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
-PLUGIN_VERSION = "1.2.1"
+PLUGIN_VERSION = "1.2.2"
 PLUGIN_API_VERSIONS = ["2.0"]
 
 from functools import partial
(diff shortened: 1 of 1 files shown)


phw:
I kind of feel that's the wrong way to go. At least we would need to add this to other plugins as well, ideally without breaking things. Also it does not do what one would expect if you have more than one genre plugin enabled (which causes mostly random behavior anyway, another problem). But if you e.g. also enable the last.fm plugin you might end up with `~mbgenre` actually being the last.fm genres and not the MB genres.

What about a different approach to solve the issue, that one wants the mood, but not the genre: Add a config page for AcousticBrainz Mood-Genre to enable using mood and/or genre?

For solving the issue of being able to access the MB genres independent of any other genre plugin I'd say we could add a `~musicbrainz_genres`  variable to Picard directly.

The actual full solution would be to have the ability to have multiple genre plugins that can collect genres, but a single set of settings that handles the final list. E.g. this could also combine genres from multiple sources. There was a short discussion about this at https://community.metabrainz.org/t/acousticbrainz-mood-genre-plugin/404057/7. I thought we also had a ticket, but can't find any right now.

Sophist-UK:
@phw Philipp - I agree with you that genres needs an overhaul (see [PICARD-2054](https://tickets.metabrainz.org/browse/PICARD-2054)).

As and when Picard has the supporting functionality, then plugins like this and lastFm can hook into that.

But right now, I want AcousticBrainz Mood and don't want AcousticBrainz Genre and this is the easiest backwards-compatible solution.

However see also [PICARD-2145](https://tickets.metabrainz.org/browse/PICARD-2145) - if that is implemented, perhaps with `~musicbrainz_genres` then this would be superfluous.

phw:
Ah yes, PICARD-2054 was the ticket :)

I don't see much that speaks against adding the genre variables for artist, album and track as you suggested. That's probably even better than a single `~musicbrainz_genres`. We want to do a beta 3 release soon and soon after the 2.6 final to get this finally out, but maybe we could even include those genre parameters still with 2.6 or at least 2.6.1 as a first step to improve genre handling.

Sophist-UK:
Hmm ... should we use singular "genre" or plural "genres"?

I am in favour of having three variables holding the genre data specific to artist, release and track **and also** an overall variable holding the original Picard metadata value of the tag "genre" (which depending on options selected will be some form of combination of the first three).

phw:
I'd be for "genres" in those tags, as they are supposed to hold multiple. And yes, `genre` itself would of course stay.

Sophist-UK:
On reflection, I think that genre plugins should always add genres to any existing list rather than replace them.

Sophist-UK:
See also [lastFm+ v2 PR](https://github.com/metabrainz/picard-plugins/pull/170).

PR #299, opened 2021-04-01
Handle missing field in the Wikidata response

Gabrielcarvfer:
Issue reported in https://community.metabrainz.org/t/picard-closes-when-matching-a-specific-release-using-the-wikidata-plugin/527185

### DIFF
--- a/plugins/wikidata/__init__.py
+++ b/plugins/wikidata/__init__.py
@@ -226,14 +226,17 @@ class Wikidata:
                             for tmp in genre_entries:
                                 if tmp == node1.attribs.get('about'):
                                     list1 = node1.children.get('name')
-                                    for node2 in list1:
-                                        if node2.attribs.get('lang') == 'en':
-                                            genre = node2.text.title()
-                                            if not matches_ignored(self.ignore_these_genres_list, genre):
-                                                genre_list.append(genre)
-                                                log.debug('New genre has been found and ALLOWED: %s' % genre)
-                                            else:
-                                                log.debug('New genre has been found, but IGNORED: %s' % genre)
+                                    if not list1:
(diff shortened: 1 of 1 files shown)


plugins/wikidata/__init__.py
```diff
@@ -226,14 +226,17 @@ def parse_wikidata_response(self, item, item_id, genre_source_type, response, re
                             for tmp in genre_entries:
                                 if tmp == node1.attribs.get('about'):
                                     list1 = node1.children.get('name')
-                                    for node2 in list1:
```
Sophist-UK, line 229: How about:
```
for node in (list1 if list1 else []):
```

Gabrielcarvfer: I have no idea. I think logging the issue happened is a good idea, as if it fail silently you won't have the data and won't know why it is missing.

Sophist-UK: Yes - it probably is.

plugins/wikidata/__init__.py
```diff
@@ -226,14 +226,17 @@ def parse_wikidata_response(self, item, item_id, genre_source_type, response, re
...
-                                        if node2.attribs.get('lang') == 'en':
-                                            genre = node2.text.title()
-                                            if not matches_ignored(self.ignore_these_genres_list, genre):
-                                                genre_list.append(genre)
-                                                log.debug('New genre has been found and ALLOWED: %s' % genre)
-                                            else:
-                                                log.debug('New genre has been found, but IGNORED: %s' % genre)
+                                    if not list1:
+                                    	log.debug('WIKIDATA: Response does not contain a name field')
```
Sophist-UK: Probably best logged as a warning rather than debug though.

phw (approved):

PR #300, opened 2021-04-08
Release tag aggregation functions

phw:
This plugin introduces a couple of scripting functions for getting common values for tags across all files of a single release or cluster. There are functions to get the minimum, maximum or average value, to get a single value if all files share the same value, to get the value with the most occurrences or to get a multi-value list of all distinct values.

For details about the functions see https://community.metabrainz.org/t/determining-whether-condition-applies-to-multiple-files/505506/4

### DIFF
--- a/plugins/releasetag_aggregations/releasetag_aggregations.py
+++ b/plugins/releasetag_aggregations/releasetag_aggregations.py
@@ -26,15 +26,17 @@ PLUGIN_DESCRIPTION = ('Add functions to aggregate tags on a release:'
                       '<li>‎$releasetag_max(name, precision=2)</li>'
                       '<li>‎$releasetag_min(name, precision=2)</li>'
                       '<li>‎$releasetag_mode(name)</li>'
+                      '<li>‎$releasetag_distinct(name, separator=; )</li>'
                       '<li>‎$releasetag_multi_avg(name, precision=2)</li>'
                       '<li>‎$releasetag_multi_max(name, precision=2)</li>'
                       '<li>‎$releasetag_multi_min(name, precision=2)</li>'
                       '<li>‎$releasetag_multi_mode(name)</li>'
+                      '<li>‎$releasetag_multi_distinct(name, separator=; )</li>'
                       '</ul>'
                       '<b>The functions work only in file naming scripts and '
-                      'the files should be either be part of a release or cluster!</b>')
-PLUGIN_VERSION = "0.2"
-PLUGIN_API_VERSIONS = ["2.4", "2.5", "2.6"]
+                      'the files should either be part of a release or cluster!</b>')
+PLUGIN_VERSION = "0.3"
+PLUGIN_API_VERSIONS = ["2.5", "2.6"]
 PLUGIN_LICENSE = "GPL-2.0-or-later"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
 
@@ -113,8 +115,7 @@ def format_number(value, precision=2):
 
 
 def mode(values):
-    """Returns the mode (the value with the most occurrences) from values.
-    """
+    """Returns the mode (the value with the most occurrences) from values."""
     l = list(values)
     if not l:
         return ''
@@ -123,6 +124,7 @@ def mode(values):
 
 def average(values, precision=2):
     """Returns the arithmetic average of all numeric elements in values.
+
     Non-numeric elements are ignored.
     """
     numbers = list(try_iter_numeric(values, skip_non_numeric=True))
@@ -132,14 +134,12 @@ def average(values, precision=2):
 
 
 def natsort_min(values, precision=2):
-    """Returns the smallest value from values, treats numeric strings as numbers.
-    """
+    """Returns the smallest value from values, treats numeric strings as numbers."""
     return format_number(max(try_iter_numeric(values)), precision=precision)
 
 
 def natsort_max(values, precision=2):
-    """Returns the largest value from values, treats numeric strings as numbers.
-    """
+    """Returns the largest value from values, treats numeric strings as numbers."""
     return format_number(max(try_iter_numeric(values)), precision=precision)
 
 
@@ -160,8 +160,7 @@ def func_releasetag_all(parser, name):
                 common_value = value
             if common_value != value:
                 return ''
-        else:
-            return common_value
+        return common_value
 
     return aggregate_release_tags(parser, name, aggregate_func)
 
--- a/plugins/releasetag_aggregations/releasetag_aggregations.py
+++ b/plugins/releasetag_aggregations/releasetag_aggregations.py
@@ -33,14 +33,15 @@ PLUGIN_DESCRIPTION = ('Add functions to aggregate tags on a release:'
                       '</ul>'
                       '<b>The functions work only in file naming scripts and '
                       'the files should be either be part of a release or cluster!</b>')
-PLUGIN_VERSION = "0.1"
-PLUGIN_API_VERSIONS = ["2.4"]
+PLUGIN_VERSION = "0.2"
+PLUGIN_API_VERSIONS = ["2.4", "2.5", "2.6"]
 PLUGIN_LICENSE = "GPL-2.0-or-later"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
 
 from functools import partial
 
 from picard.cluster import Cluster
+from picard.metadata import MULTI_VALUED_JOINER
 from picard.script import script_function
 from picard.script.parser import normalize_tagname
 
@@ -142,6 +143,10 @@ def natsort_max(values, precision=2):
     return format_number(max(try_iter_numeric(values)), precision=precision)
 
 
+def distinct(values, separator=MULTI_VALUED_JOINER):
+    return separator.join(set(values))
+
+
 @script_function(documentation="""`$releasetag_all(name)`
 
 Returns the value of the tag name if all files on the album have the same
@@ -234,4 +239,23 @@ Similar to $releasetag_avg(), but considers each value of multi-value tags separ
 **Only works in File Naming scripts.**""")
 def func_releasetag_multi_avg(parser, name, precision="2"):
     aggregate_func = partial(average, precision=precision)
+    return aggregate_release_tags(parser, name, aggregate_func, multi=True)
+
+
+@script_function(documentation="""`$releasetag_distinct(name, separator=; )`
+
+Returns a multi-value tag with all distinct values of tag across all the files on the release.
+**Only works in File Naming scripts.**""")
+def func_releasetag_distinct(parser, name, separator=MULTI_VALUED_JOINER):
+    aggregate_func = partial(distinct, separator=separator)
+    return aggregate_release_tags(parser, name, aggregate_func)
+
+
+@script_function(documentation="""`$releasetag_multi_distinct(name, separator=; )`
+
+Returns a multi-value tag with all distinct values of tag across all the files on the release.
+Similar to $releasetag_distinct(), but considers each value of multi-value tags separately.
+**Only works in File Naming scripts.**""")
+def func_releasetag_multi_distinct(parser, name, separator=MULTI_VALUED_JOINER):
+    aggregate_func = partial(distinct, separator=separator)
     return aggregate_release_tags(parser, name, aggregate_func, multi=True)
--- a/plugins/releasetag_aggregations/releasetag_aggregations.py
+++ b/plugins/releasetag_aggregations/releasetag_aggregations.py
@@ -96,6 +96,10 @@ def aggregate_release_tags(parser, name, aggregate_func, multi=False):
 
 
 def format_number(value, precision=2):
+    try:
+        precision = int(precision)
+    except (TypeError, ValueError):
+        precision = 0
     if isinstance(value, int):
         return str(value)
     elif isinstance(value, float):
--- a/plugins/releasetag_aggregations/releasetag_aggregations.py
+++ b/plugins/releasetag_aggregations/releasetag_aggregations.py
@@ -1,0 +1,233 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 Philipp Wolfer
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


Sophist-UK:
@phw Philipp - Thanks for picking up on this as it looks similar to something I recently suggested.

A. I do wonder, however, if we can shorten the function names a bit. the `$releasetag_` prefix is a bit cumbersome IMO, particularly with some of the longer suffixes. Could I suggest the use of `$album_` as the prefix instead? 

B. I suspect that you created these as file-naming-script only functions for a reason, however it seems to me that it might be possible to defines these in a way that they would also work in tagging scripts. Could you explain your reasoning for this choice so that we can try to find a way to make it work in tagging scripts?

phw:
> Could I suggest the use of `$album_` as the prefix instead?

Yep, fine with me. I'll update it.


> Could you explain your reasoning for this choice so that we can try to find a way to make it work in tagging scripts?

It's not really a choice, it is currently simply not possible to have a script function do this except for file naming. Scripts don't have access to the file or album. That's also the reason why `$matchedtracks` and `$is_complete` do only work for file naming. The only situation where there is access to a file is in the renaming script. I have written a bit about this and what would be needed to solve this on https://community.metabrainz.org/t/determining-whether-condition-applies-to-multiple-files/505506/7

But having slept over this I was a bit too optimistic. Simply making the file available in tagger scripts would be trivial, but would not work as expected. The reason is that a script runs when a file gets attached, so any scripting relying on data from all files will not provide the expected results except for the very last file added. This is non-trivial to solve, and also the reason why I decided against providing the file to the parser when I added support for file metadata in tagger scripts.

What we could do rather easily is having a way to access the album, track or cluster for which a script runs. Then e.g. functions like the ones added here could at least work on all tracks of an album. They shouldn't rely on the files attached to those entities though unless we are in a renaming context.

phw:
I renamed all functions to `$album_*`. Indeed much shorter and nicer to read

phw:
I renamed all functions to `$album_*`. Indeed much shorter and nicer to read

Sophist-UK:
> It's not really a choice, it is currently simply not possible to have a script function do this except for file naming. Scripts don't have access to the file or album. That's also the reason why `$matchedtracks` and `$is_complete` do only work for file naming. The only situation where there is access to a file is in the renaming script. I have written a bit about this and what would be needed to solve this on https://community.metabrainz.org/t/determining-whether-condition-applies-to-multiple-files/505506/7
> 
> But having slept over this I was a bit too optimistic. Simply making the file available in tagger scripts would be trivial, but would not work as expected. The reason is that a script runs when a file gets attached, so any scripting relying on data from all files will not provide the expected results except for the very last file added. This is non-trivial to solve, and also the reason why I decided against providing the file to the parser when I added support for file metadata in tagger scripts.
>
> What we could do rather easily is having a way to access the album, track or cluster for which a script runs. Then e.g. functions like the ones added here could at least work on all tracks of an album. They shouldn't rely on the files attached to those entities though unless we are in a renaming context.

I think this probably needs unpacking:

a. I am unclear why we need to look at files rather than tracks. Yes - there is some metadata that can only be gathered from a file (like bitrate or format), but most data people will want to analyse the downloaded MB tags. So I would personally prefer to see functions that summarise the track metadata which can be used in tagging scripts, and if necessary a separate set of functions to access metadata that is only available in a file.

b. It would probably be helpful to have a track variable `%_attached_files%` that shows the number of files associated with a track.

c. I think that it IS possible to access attached file metadata from a tagging script, but I do also see the difficulties of analysing file-only metadata in a tagging script if the tagging script runs when each file is attached. Of course, there are different use cases, and the situation is different for each. For example, after a lookup or a multi-file drag&drop, all files are attached to tracks automatically and it should be possible to attach all files first and then run the tagging scripts afterwards. Of course, this only works in some use cases and not others, which means results will not be consistent, and short of running all tagging scripts again every time another file is attached - which may have other side effects - and I am unclear how we could fix this.

d. Maybe my decaying knowledge of the code is worse than I thought. Don't we store a link to the Album object in the Track Object and a list of links to Track objects in the Album, and similarly a link to a Track object in the File object and a list of links to File objects in the Track so we can navigate between these objects?

phw:
> a. I am unclear why we need to look at files rather than tracks.

Because for the use case in https://community.metabrainz.org/t/determining-whether-condition-applies-to-multiple-files/505506 files work better: All the files are the data that gets actually stored, and they provide file data like bitrate. In general at time of saving there is not much difference between file and track metadata except the technical variables and maybe some odd cases with specific scripting. If this would also work in normal tagger script it would most likely be working on tracks, but as this does not work anyway right now there is no real point in using tracks instead of files.

> b. It would probably be helpful to have a track variable %_attached_files% that shows the number of files associated with a track.

Sure. But issue to the same limitations. This currently would only reliable work in renaming scripts, see my earlier comments and below.

> c. I think that it IS possible to access attached file metadata from a tagging script,

That needs more details. First, no, currently it is **not** possible, plain and simple. All the script parser gets to work with is a metadata object. Only for file naming scripts a reference to the file is provided, which in turn allows access to tracks and albums etc.

It would be **possible** to extend Picard to have access to that, please read my comments on the forums and above. What I think we should have is providing references to track, album and / or cluster for scripts in general. Then script functions could operate on this data.

It would still not be save to rely on other files, though. When scripts get executed other files may or may not be present. It's a similar situation as with plugins: I really don't recommend a plugin that uses data from files. The Classical Extras I think does it, but that only works under specific circumstances.

Solving this requires some serious refactoring in the way Picard works. Currently if you lookup let's say 1000 single files each file is processed individually and might be matched to a release. Once a file got matched, the release gets loaded and when the loading finished the file gets matched to a track and the metadata gets updated (including running the scripts). At that point one, some or all files might be attached to the release. That means to reliable finish loading an album with all files attached you'd need to wait until all 1000 files have been processed. Even worse, a user can at any time add or remove files for a release (either manually or by other lookups). So maybe Picard would need to re-evaluate the scripts for all files on the release whenever a file gets added/removed to/from the release.

> d. Maybe my decaying knowledge of the code is worse than I thought. Don't we store a link to the Album object in the Track Object and a list of links to Track objects in the Album, and similarly a link to a Track object in the File object and a list of links to File objects in the Track so we can navigate between these objects?

Sure, that is how the plugin navigates around: It goes from the file via track to the album, to all the files attached to tracks on that album.

Sophist-UK:
> > a. I am unclear why we need to look at files rather than tracks.
> 
> Because for the use case in https://community.metabrainz.org/t/determining-whether-condition-applies-to-multiple-files/505506 files work better

Yes - as I said, there are some use cases which need file data only for file renaming. But more use cases for MB data in file naming and tagging scripts.

I accept that file data cannot be summarised reliably in tagging scripts unless we run the tagging scripts on all tracks / files whenever any file is attached or detached - but if we grouped file attachments to tracks where possible, overhead of running tagging scripts multiple times would be much less of an issue.

I guess I am suggesting either:

a. Fix Picard first so that it groups file attachments / detachments and then runs tagging scripts on all tracks / files whenever a new file is attached / detached; or

b. Have separate but similar `$album_*` to work on track MB data and working in both tagging and renaming scripts PLUS similar `$files_*` which works on file data and only in file naming scripts.

>All the files are the data that gets actually stored, and they provide file data like bitrate. 

Yes - some additional file-only data. But all other data that get's stored is in track metadata.

> In general at time of saving there is not much difference between file and track metadata except the technical variables and maybe some odd cases with specific scripting.

Exactly.

> If this would also work in normal tagger script it would most likely be working on tracks, but as this does not work anyway right now there is no real point in using tracks instead of files.

Unclear why we cannot summarise Track data in an album in tagger scripts. This does NOT change when files are attached, so no issues of attaching one file and running tagging script on that file but not running tagging scripts on other tracks which might then use out of date summaries.

> > b. It would probably be helpful to have a track variable %_attached_files% that shows the number of files associated with a track.
> 
> Sure. But issue to the same limitations. This currently would only reliable work in renaming scripts, see my earlier comments and below.

Tagging scripts are run again when the value of this variable changes, so it can be used singly in a tagging script for this track. But unless tagging scripts are run on all tracks every time a file is attached / detached, then summarising this would only be reliable in a file naming script.

> > c. I think that it IS possible to access attached file metadata from a tagging script,
> 
> That needs more details. First, no, currently it is **not** possible, plain and simple. All the script parser gets to work with is a metadata object. Only for file naming scripts a reference to the file is provided, which in turn allows access to tracks and albums etc.
>
> It would be **possible** to extend Picard to have access to that, please read my comments on the forums and above. What I think we should have is providing references to track, album and / or cluster for scripts in general. Then script functions could operate on this data.
 
Ah - ok. This **is** a limitation of Picard currently - but I agree that we should add a parameter so that the track object is also passed with navigation to album / file metadata from there.

> It would still not be save to rely on other files, though. When scripts get executed other files may or may not be present. It's a similar situation as with plugins: I really don't recommend a plugin that uses data from files. The Classical Extras I think does it, but that only works under specific circumstances.

It would be safe **if** we reran tagging scripts on all tracks/file whenever a file is attached / detached.

But Track / Album metadata plugins run only once when data is downloaded from MB and not on file attachment / detachment, hence issues with accessing file data from a plugin.

Perhaps Picard also needs events for plugins to hook into to run before / after a tagging script is run.

> Solving this requires some serious refactoring in the way Picard works. Currently if you lookup let's say 1000 single files each file is processed individually and might be matched to a release. Once a file got matched, the release gets loaded and when the loading finished the file gets matched to a track and the metadata gets updated (including running the scripts). At that point one, some or all files might be attached to the release. That means to reliable finish loading an album with all files attached you'd need to wait until all 1000 files have been processed. 

Yes - I agree that if you lookup 1,000 files individually, files are added to releases singly rather than in groups and this might cause a LOT of additional tagging script runs. Waiting for the end of the 1,000 files is not acceptable, nor is the n^2 performance hit of running a tagging script on n tracks n times (once for every file as it is added).

> Even worse, a user can at any time add or remove files for a release (either manually or by other lookups). So maybe Picard would need to re-evaluate the scripts for all files on the release whenever a file gets added/removed to/from the release.

IMO this is less of a problem, because the limitation here is going to be the speed with which a user can drag and drop individual or groups of files.

> > d. Maybe my decaying knowledge of the code is worse than I thought. Don't we store a link to the Album object in the Track Object and a list of links to Track objects in the Album, and similarly a link to a Track object in the File object and a list of links to File objects in the Track so we can navigate between these objects?
> 
> Sure, that is how the plugin navigates around: It goes from the file via track to the album, to all the files attached to tracks on that album.

Ok - thanks for the scripting parameters explanation which makes it clear what the current Picard limitation is.

## Summary
I believe that there is benefit in adding functionality to Picard first as discussed above, and then making this new plugin work only with that version of Picard or later so that you get tagging script functionality.

phw:
> Perhaps Picard also needs events for plugins to hook into to run before / after a tagging script is run.

There are plugin hooks for when files are added / removed from releases. I don't really see a use case for before / after scripts, but if there is one we could add such hooks.

> Waiting for the end of the 1,000 files is not acceptable, nor is the n^2 performance hit of running a tagging script on n tracks n times (once for every file as it is added).

Which turns this into a hard problem to solve. I honestly would exclude the case of providing access to files to scripts completely for now. Rather I would focus on providing album / track / cluster to the scripts.

> I believe that there is benefit in adding functionality to Picard first as discussed above, and then making this new plugin work only with that version of Picard or later so that you get tagging script functionality.

Agreed on the extending Picard part. But I don't see why we should not make the plugin available as is. The situation is basically that a user is asking for a use case A, and we say that's not possible with Picard right now but a plugin could do it. Now we have a plugin that solves the requested use case A, why should we hold it back until a future Picard version which would allow the plugin to also solve a different use case B that was not even asked for?

I mean the goal was to solve use case A in current Picard releases.

phw:
I don't get what Codacy is complaining about at https://app.codacy.com/gh/metabrainz/picard-plugins/pullRequest?prid=7228675 . That looks bugged on the linter part.

Sophist-UK:
> There are plugin hooks for when files are added / removed from releases. I don't really see a use case for before / after scripts, but if there is one we could add such hooks.

Given that scripts are run when files are added, I think you are right.

> Rather I would focus on providing album / track / cluster to the scripts.

I agree with this strategy.

> > I believe that there is benefit in adding functionality to Picard first as discussed above, and then making this new plugin work only with that version of Picard or later so that you get tagging script functionality.
> 
> Agreed on the extending Picard part. But I don't see why we should not make the plugin available as is. The situation is basically that a user is asking for a use case A, and we say that's not possible with Picard right now but a plugin could do it. Now we have a plugin that solves the requested use case A, why should we hold it back until a future Picard version which would allow the plugin to also solve a different use case B that was not even asked for?

Because once you provide the plugin, it is difficult to change approach. 

I do indeed see that there is a file-based use case, but if we want to issue this plugin which is file naming only and looks at file metadata, then can we rename the functions as `$files_*` in order to describe where the data is being taken from and to leave an opening for track metadata using `$album_*` or `$track_*` which more clearly relate to MB data.

phw:
Plugins are much easier to change then if this was added into Picard itself. That's the reason why I would only want to have this as a plugin for now at the first place. Once we are fully satisfied with the functionality we can choose to include it into Picard.

My plan actually was to update the functions to handle track metadata when run in tagger script mode once this is possible. I must say I find `$files_*` highly confusing, the intention was to have a function that aggregates data across a release / album, `$files_*` doesn't tell me that.

Sophist-UK:
> My plan actually was to update the functions to handle track metadata when run in tagger script mode once this is possible. I must say I find `$files_*` highly confusing, the intention was to have a function that aggregates data across a release / album, `$files_*` doesn't tell me that.

If we can have a common set of functions that (at a later date) will handle aggregation across both Track and File metadata (and be run in tagger scripts for Track metadata), then great - we don't need to reserve names for future functionality. :-)

But if we use `$album_*` for functions that will only ever run against File metadata in file renaming scripts, and then we implement other functions for Track metadata which will run in either Tagging or File Naming scripts, then I think we are going to cause confusion.

Of course this is just one persons view - and not a consensus.

I trust your judgement on this - so please go ahead with whatever approach you think is best.

phw:
> If we can have a common set of functions that (at a later date) will handle aggregation across both Track and File metadata (and be run in tagger scripts for Track metadata), then great - we don't need to reserve names for future functionality. :-)

At least I think that should turn out.

It is clear to me at that point that we should not yet integrate this into Picard itself. But for me plugins are the perfect place to put in functionality that is focused on very specific use cases and maybe is rather opinionated. It's also a way to put out code that can be iterated on comparable quickly. If it turns out well enough we should integrate this into Picard.

Sophist-UK:
> ... plugins are the perfect place to put in functionality that is focused on very specific use cases and maybe is rather opinionated. It's also a way to put out code that can be iterated on comparable quickly. If it turns out well enough we should integrate this into Picard.

Agree completely.

zas:
> I don't get what Codacy is complaining about at https://app.codacy.com/gh/metabrainz/picard-plugins/pullRequest?prid=7228675 . That looks bugged on the linter part.

No idea, I tried to run PEP257 check tool locally, and it doesn't complain on the same code.

PR #303, opened 2021-04-16
happidev_lyrics: Fix possible endless recursion if no api key was configured

phw:
If no API key was configured, the plugin would increase album._requests and call album._finalize_loading without actually doing an asynchronous request. Since the track metadata processors get run inside album._finalize_loading this would cause an endless recursion.


phw:
This is actually another sign that we need to provide a proper interface for plugins registering their requests, this is a design flaw. But for now the plugin should be fixed.

zas (approved):

Issue #269, opened 2020-10-30
AcousticBrainz Mood-Genre plugin should have a settings-page like the last-fm plugin


natrius:
To set things like 

"Ignore genre" (alternatively split to "AcousticBrainz Mood" and "AcousticBrainz Genre",
"join multiple values with: '; / \'" or something like that. Let us choose like other plugins are doing currently
"Ignore genres" with a list separated by new line for some weird genres or moods i don't like. 

Some people just want to get moods as tags and get the genre with the last.fm plugin. Because its nice to get something like that.

phw:
There is now a settings page in the rewritten AcousticBrainz 2.0 plugin. You can choose whether you want mood, genre or both. It even incorporates the functionality of acousticbrainz_tonal-rhythm and allows you to fetch the data from AB to more custom tags if needed.

By default it behaves as the previous version and is configured to load both mood and genre.

There are no options for joining or excluding specific genres. The idea here is that these options should be centralized in Picard itself, and that it is possible to combine multiple genre plugins.

phw:
Closing this as the acusticbrainz plugin now provides a settings page.

PR #304, opened 2021-05-10
Add a full AcousticBrainz highlevel tagger plugin

wargreen:
First version of the acousticbrainz fullmood tagger plugin
Rename the acousticbrainz mood-genre file & directory to help distinguish the two plugins

In this new plugin, i propose a way to name theses tags. I hope it's the good way !


Sophist-UK:
What will be the impact on anyone who has the existing acousticbrainz plugin installed after the directory changes? Will they be offered the new version to download? If they download it will they still have the old version installed or will it be deleted?

wargreen:
Is Picard verify if a file with the same hash is already in the local plugin folder ?
What is the "ID" of a plugin ? The folder name, the name or the hash ?

plugins/acousticbrainz_fullmood/acousticbrainz_fullmood.py
```diff
@@ -0,0 +1,88 @@
...
+        data = load_json(data).get(metadata["musicbrainz_recordingid"])
+        if data and "0" in data and "highlevel" in data["0"]:
+            data = data["0"]["highlevel"]
+            f, c = 0, 0
+            for classifier, v in data.items():
+                if "all" in v:
+                    c += 1
+                    for feature, proba in v["all"].items():
+                        f += 1
+                        metadata["AB.{}.{}".format(classifier, feature)] = dump_json(proba)
```
phw: I would suggest not using dots in the name. Dots are not valid for Picard scripting variable names, hence one could not easily write `%AB.foo.bar%`, but you would have to use the less convenient `$get(AB.foo.bar)` instead. Maybe use underscores or colon instead?

phw (changes requested):
Sorry for the delay, but here is some feedback.

I think in general this is fine. But I would prefer to have this as part of the acousticbrainz plugin, adding a UI to enable storing all those additional fields. Having it as a separate plugin means twice the number of requests if the user enables both the existing acousticbrainz plugin and the new one.

So my proposal would be to add the functionality to the existing plugin, but add a UI where the user can explicitly enable writing all the additional features to tags.

The Wikidata plugin is an example of a plugin with an additional options page. Ideally the UI itself is created as a .ui file using Qt Designer, and compiled into a .py file using Qt's uic. That's how it is handled for wikidata and most (if not all) other plugins with own UI. The script I added at https://github.com/metabrainz/picard-plugins/pull/305 can help compiling the .ui files.

If wanted I could assist in adding the UI, we could resolve https://github.com/metabrainz/picard-plugins/issues/269 that way also.

An alternative to the UI could be to instead write all those features to hidden tags (staring with an underscore). That way they are available for scripting, but don't get written to the tags by default. Users wanting to make use of the data could do so by appropriate scripting.

> Rename the acousticbrainz mood-genre file & directory to help distinguish the two plugins

Please don't do that, it will break updates of existing installs. Also a single file plugin is expected to be in a folder of same name. That's only a convention, but if we name plugins we should follow it. Or actually convert this plugin into a proper python module and name it's file `__init__.py`.

plugins/acousticbrainz_fullmood/acousticbrainz_fullmood.py
```diff
@@ -0,0 +1,88 @@
...
+    if error:
+        album._requests -= 1
+        album._finalize_loading(None)
+        return
+
+    try:
+        data = load_json(data).get(metadata["musicbrainz_recordingid"])
+        if data and "0" in data and "highlevel" in data["0"]:
+            data = data["0"]["highlevel"]
+            f, c = 0, 0
```
phw: Can we get some more expressive names for those variables?

wargreen:
Thank you for the feedback.
I will try to integrate my plugin in the previous acousticbrainz plugin, with a GUI. With my python level it can take a while but i will try !
Do you think that it's a good idea to integrate also the acousticbrainz_tonal-rythm plugin ?

plugins/acousticbrainz_fullmood/acousticbrainz_fullmood.py
wargreen: Since foo and bar can have underscore inside, i think colon is better.
I will change it.

phw:
> Do you think that it's a good idea to integrate also the acousticbrainz_tonal-rythm plugin ?

Overall yes, but for now I'd say no. As this accesses a different API endpoint integrating both into a single plugin requires some more refactoring. Let's focus on this API endpoint first.

rdswift:
I see that you've also made changes to another plugin (fanart.tv) in this PR.  Those should be removed and placed in a separate PR.

wargreen:
Sorry error when rebasing... Fixing that now.

wargreen:
Here the first V2 plugin.
We have implement low and hi level enpoint directly for simplify further developpement. We've think that start to dev for only highlevel make lot more work.
The GUI allow to select the previous behavior (by default) and/or add the tonal-rhythm, all the highlevel data, a subset of the lowlevel data.
I hope it's good !

plugins/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,356 @@
...
+from json import dumps as dump_json, JSONDecodeError
+from functools import partial
+from picard import config, log
+from picard.webservice import ratecontrol
+from picard.util import load_json
+from picard.ui.options import register_options_page, OptionsPage
+from picard.metadata import register_track_metadata_processor
+from picard.plugins.acousticbrainz.ui_options_acousticbrainz_tags import Ui_AcousticbrainzOptionsPage
+
+ratecontrol.set_minimum_delay((ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT), 50)
```
phw: I know this was exactly the same on the old plugin, but we have issues with the rate limit https://github.com/metabrainz/picard-plugins/issues/282 . Unfortunately Picard does not yet support dynamic rate limit. But https://acousticbrainz.readthedocs.io/api.html#rate-limiting says "We typically set the limit to 10 queries every 10 seconds per IP address". That means, unfortunately, that setting this to 1000 is the better value :( Maybe we can get away with a bit lower?

@alastair What is your suggestion how we should handle this?

Ideally we should extend Picard to support dynamic rate limits, maybe by passing a function instead of just an int. I had already added a ticket for Picard to do so. But for now we need to deal with a static rate limit. But since we need should support also older Picard versions we need have a solution with a sensible static value as well.

plugins/acousticbrainz/options_acousticbrainz_tags.ui
```diff
@@ -0,0 +1,103 @@
...
+    <property name="geometry">
+     <rect>
+      <x>10</x>
+      <y>90</y>
+      <width>341</width>
+      <height>25</height>
+     </rect>
+    </property>
+    <property name="text">
+     <string>Add all  highlevel Acousticbrainz tags</string>
```
phw: There are two spaces here between "all  highlevel".

plugins/acousticbrainz/options_acousticbrainz_tags.ui
```diff
@@ -0,0 +1,103 @@
...
...gt;&lt;p&gt;The low level subset include :&lt;/p&gt;&lt;ul style=&quot;margin-top: 0px; margin-bottom: 0px; margin-left: 0px; margin-right: 0px; -qt-list-indent: 1;&quot;&gt;&lt;li style=&quot; margin-top:12px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;rhythm:bpm&lt;/li&gt;&lt;li style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;tonal:chords_change_rate&lt;/li&gt;&lt;li style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;tonal:chords_key&lt;/li&gt;&lt;li style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;tonal:chords_scale&lt;/li&gt;&lt;li style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;tonal:key_key&lt;/li&gt;&lt;li style=&quot; margin-top:0px; margin-bottom:12px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;tonal:key_scale&lt;/li&gt;&lt;/ul&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
phw: A superluous space between text and colon in "subset include :"

plugins/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,356 @@
...
+# =============================================================================
+# (define plugin options and link with user interface)
+
+class AcousticbrainzOptionsPage(OptionsPage):
+    NAME = "AcousticBrainz"
+    TITLE = "AcousticBrainz tags"
+    PARENT = "plugins"
+    
+    options = [
+        config.BoolOption("setting", "acousticbrainz_add_simplemood", True),
```
phw, line 343: Let's have separate options for mood and genre. We recently had this discussion in PICARD-2141, and that would resolve this partially. We don't need to do more genre options here, I plan to refactor the genre handling of plugins soon (once I get some other important refactorings out of the way)

plugins/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,356 @@
...
+be configured to include all highlevel data.
+<br/><br/>
+Based on code from Andrew Cook, Sambhav Kothari
+<br/><br/>
+<b>WARNING:</b> Experimental plugin. All guarantees voided by use.'''
+
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
+PLUGIN_VERSION = "2.0"
+PLUGIN_API_VERSIONS = ["2.0"]
```
phw: Can you set all of 2.0, 2.1, 2.2, 2.3, 2.4, 2.5 and 2.6 here?

The reason is that we e.g. might decide that a new Picard release no longer will support 2.0, 2.1 and 2.2. Then this plugin would stop working without an update.

phw (changes requested):
Thank you very much for the rewrite, a really big improvement. And opens the door for future improvements. Tested it locally and it works really well. Overall just minor comments on the code and functionality.

Codacy has some complaints, see https://app.codacy.com/gh/metabrainz/picard-plugins/pullRequest?prid=7402909

I don't know exactly why it suddenly insists on the docstrings, actually those tests are marked as disabled. But addressing at least the other issues would be good.

plugins/acousticbrainz/__init__.py
alastair: We're a little bit more lenient in our checking that what is listed in the documentation, although you're right, 50 is definitely too low. Do you have a general idea of how many queries you might make? Does it always depend on how many items a user wants to process with picard?

Remember that this endpoint is a bulk lookup endpoint, so you can request up to 25 items, separated by a `;`. Will picard let you run a plugin over many items at once rather than per-track?

phw: I guess this could be refactored to batch request the data for all recordings of a release. But I'd like to have this done in a separate PR, this one looks good so far in replicating the existing plugins functionality and only needs some minor cleanup.

Apart from that we cannot really say how much is expected, it all depends on how many releases the user loads and how many tracks those have. So currently it would be one request per track, if we would batch it it would be one request per 25 tracks.

plugins/acousticbrainz/__init__.py
wargreen: Now there are two option : mood and genre

plugins/acousticbrainz/__init__.py
wargreen: I've set 250ms for now, i hope it is good.

phw: Seems ok, let's go with this

phw (approved):
Thanks a lot for the changes and for the plugin. This works really well. I pushed some small fixes to make Pylint happy and some small UI tweaks for the option page. Let's merge this!

phw:
@wargreen Merged, the plugin will be automatically made available from the Picard website sometime the next 12 hours. Would be cool if you could introduce this new AB plugin on the forums.

Also if you want to you could look into the improvement alastair suggested. It should be possible I think to convert this plugin to a album_metadata_processor, then call the endpoints with a single request, providing all recording IDs. That should bring a big improvement to loading times and less trouble with rate limit :)

wargreen:
So cool if our work can be useful !
For the convertion to a album processor, i'm not sure to be the good guy, i have been helped by a friend (pistache) who is gone on a boat... Maybe on day if i can !
I will talk about this plugin in the Beets irc for sure, to motivate them to implement the same tool, and what's the forum you mean ?

phw:
Allright. If nobody takes this up I will see I can tackle this once I have fnished my work-in-progress patches for Picard and we are on track for the 2.7 release.

> and what's the forum you mean ?

The community forums at https://community.metabrainz.org

pstch:
> Also if you want to you could look into the improvement alastair suggested. It should be possible I think to convert this plugin to a album_metadata_processor, then call the endpoints with a single request, providing all recording IDs. That should bring a big improvement to loading times and less trouble with rate limit :)

I will submit a new PR for this. Thanks for the suggestion @alastair.

phw:
@pstch This particular refactoring has already been addressed in https://github.com/metabrainz/picard-plugins/pull/308

It really speeds using this plugin up a lot. Together with your changes here this makes this a really cool plugin. Thanks a lot again :)

PR #306, opened 2021-06-07
Haikuattrs: Support Media:Length and Audio:Bitrate attributes

phw:
To fill these attributes:

image

With this all attributes supported for audio files by Haiku should also be supported by the plugin.

### DIFF
--- a/plugins/haikuattrs/haikuattrs.py
+++ b/plugins/haikuattrs/haikuattrs.py
@@ -151,7 +151,8 @@ if be:
                     if tag == 'date':
                         value = value[:4]
                     elif tag == '~length':
-                        value = file.orig_metadata.length
+                        # Haiku expects the Media:Length attribute in nanoseconds
+                        value = file.orig_metadata.length * 1000
                     elif tag == '~rating':
                         try:
                             value = int(value) * 2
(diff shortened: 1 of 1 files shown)


plugins/haikuattrs/haikuattrs.py
```diff
@@ -136,23 +141,35 @@ def write_attr(fd, attr, attr_value):
...
-                    elif tag == '~rating':
-                        try:
-                            value = int(value) * 2
-                        except ValueError:
-                            log.warning(
-                                'haikuattrs: rating %r for %s is not an integer value',
-                                value, file.filename)
-                            continue
                     if not write_attr(fd, attr, value):
                         log.error('haikuattrs: setting %s=%s for %s failed' % (
```
Sophist-UK, line 174:
```
                if value and not write_attr(fd, attr, value):
```
?

phw: No, that would break the else:

```
if value:
   # write the attr
else:
  # remove the attr
```

Sophist-UK: Oh - sorry - that was cut off by the diff. D'oh!

phw: Actually when I rewrote this I first tried it exactly like you proposed before I realized the else case ;)

PR #308, opened 2021-07-05
AcousticBrainz batch requests

### LINKED ISSUE #233: acousticbrainz: query multiple recordings in one request
As discussed in #228 the AcousticBrainz data allows querying data for multiple recordings (up to 25) with one request. The acousticbrainz plugins should leverage this to do only one request per album loaded.

phw:
Refactor the (new) AcousticBrainz plugin to batch requests for 25 recordings.

Batching is done per album, so at minimum there will be one request per album and details level (low, high). For the majority of cases that means 1 request per album per level. Special handling for non-album tracks, as those do not trigger the album metadata hook.

Discussion: It would of course also be possible to batch across albums. But that would make the handling much more complicated. I think for now this provides a good compromise.

Fixes https://github.com/metabrainz/picard-plugins/issues/233

### DIFF
--- a/plugins/acousticbrainz/__init__.py
+++ b/plugins/acousticbrainz/__init__.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 # AcousticBrainz plugin for Picard
-# Copyright (C) 2021 Wargreen
+#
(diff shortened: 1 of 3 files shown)


phw:
@wargreen This is the finished refactoring

PR #309, opened 2021-08-06
amazon: fall back to try loading cover only by ASIN

phw:
Having the full Amazon URL is preferrable, as the correct image path can be deduced from the Amazon host. But if relationships are not available, try all the different hosts.

See also discussion at https://community.metabrainz.org/t/does-the-amazon-coverart-plugin-still-work/545285


zas (approved):
LGTM

PR #310, opened 2021-08-27
PICARD-2268: Sort URLs to avoid random order

Sophist-UK:
Random order causes albums to be non-perfect if this is the only data change.


phw (approved):
Yep, that makes sense :) Thanks!

PR #312, opened 2021-10-01
Add cover art lookup for selected track or album

rdswift:
This change to the Search Engine Lookup plugin adds a right-click option on album and track lines in the right-hand pane to open a search in their browser for cover art associated with the selected album or track.  It also incorporates some refactoring to avoid code duplication.

This is in a response to an enhancement request entered at https://github.com/rdswift/picard-plugins/issues/3


phw (approved):

phw:
Thanks :)

PR #314, opened 2021-10-04
Add new plugin to submit ISRCs found in audio files metadata

rdswift:
This plugin adds a right click option on an album to submit the ISRCs to the MusicBrainz server specified in the Option settings.

To use this function, you must first match your files to the appropriate tracks for a release.  Once this is done, but before you save your files (if you have Picard set to overwrite the `isrc` tag in your files), right-click the release and select "Submit ISRCs" in the "Plugins" section.  For each file that has a single valid ISRC in its metadata, the ISRC will be added to the recording on the release if it does not already exist.  Once all tracks for the release have been processed, the missing ISRCs will be submitted to MusicBrainz.

If a file's metadata contains multiple ISRCs, such as if the file has already been tagged, then no ISRCs will be submitted for that file.

If one of the files contains an invalid ISRC, or if the same ISRC appears in the metadata for two or more files, then a notice will be displayed and the submission process will be aborted.

When ISRCs have been submitted, a notice will be displayed showing whether or not the submission was successful.


plugins/submit_isrc/__init__.py
```diff
@@ -0,0 +1,262 @@
...
+                '' if self.isrc_count == 1 else 's',
+            ))
+            return
+
+        # Decode response if necessary.
+        xml_text = str(document, 'UTF-8') if isinstance(document, (bytes, bytearray, QtCore.QByteArray)) else str(document)
+
+        # Build error text message from returned xml payload
+        err_text = ''
+        matches = re.findall(r'<text>(.*?)</text>', xml_text)
```
phw, line 249: `document` should actually be a parsed XML document. Not quite sure which structure the response has, but you can do things like `document.metadata[0].error[0].text[0]` here. Don't know if that is simpler in this specific case, though.

phw (commented):
Looks good overall I think.

My only concern would be more wrong submissions, don't know how much we can trust ISRCs in tags. But I guess with digital becoming more and more common we cannot rely on CD submissions that much.

Ideally both ISRC submission from CD and from files would be added to Picard. For file submission we should look into a way to streamline data submission, with the user having the ability to inspect what gets submitted and what not. That could include all of AcoustID, AcousticBrainz, ISRC, barcode, ratings and even genres.

Until we have this this plugin looks good to me for the ISRC case.

phw:
I tested this on https://musicbrainz.org/release/e6be3b4a-e254-4541-ae98-a82a1d076d79, works well. I'm fine merging this.

Not sure if using the document as an XML object as I suggested would simplify anything or would rather make it more complex. The current way of checking the response is rather simple and straight forward.

rdswift:
> Not sure if using the document as an XML object as I suggested would simplify anything or would rather make it more complex.

It actually didn't work, and raised the exception:

```
AttributeError: 'QByteArray' object has no attribute 'metadata'
```

That leads me to believe that `document` is not being returned as a parsed XML document as we thought.  Either that or I'm misunderstanding something.

plugins/submit_isrc/__init__.py
rdswift: See discussion below.

Issue #179, opened 2018-11-09
Tango.info plugin returns no data


phw:
The Tango.info plugin does not work as it receives empty results when fetching the tango.info HTML pages. See https://github.com/metabrainz/picard-plugins/pull/172/ for details

phw:
Looked at this a bit more. The server responds with a content size of 0 and no body data.

For a request to https://tango.info/08427328130110 we get a 200 response with the following headers:

```
Date: Tue, 13 Apr 2021 13:17:43 GMT
Server: Apache
Vary: User-Agent
X-Content-Type-Options: nosniff
X-Frame-Options: sameorigin
Referrer-Policy: strict-origin-when-cross-origin
X-XSS-Protection: 1; mode=block
Content-Security-Policy: default-src 'none'; frame-ancestors 'self'; frame-src 'self' https://www.google.com https://calendar.google.com https://accounts.google.com https://www.youtube.com; img-src 'self' https://anna.info/x/ https://*.openstreetmap.org; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; media-src 'self';
Content-Length: 0
Keep-Alive: timeout=5, max=98
Connection: Keep-Alive
Content-Type: text/html; charset=UTF-8
```

The same request works with curl:

```
curl -v -A "MusicBrainz-Picard/2.6.1.dev1 (Linux-5.8.0-48-generic-x86_64-with-glibc2.32;CPython-3.8.6)" https://tango.info/08427328130110
```

Gives the following headers:

```
*   Trying 178.33.33.36:443...
* TCP_NODELAY set
* Connected to tango.info (178.33.33.36) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
[...]

PR #317, opened 2021-10-18
tangoinfo: remove plugin as it is unmaintained and unusuable

phw:
The requests to the server currently all return an empty response for unknown reason, see #179

As this is currently not working and nobody has come up with a fix I think it is better to remove the plugin. We can bring it back if there is interest and someone produces a working version.

@ix5 Pinging you as you originally wrote that plugin. Let me know if you are interested in trying to fix this. Currently I am not sure what is happening, but to me it looks like they intentionally try to block someone from scraping their data. But changing the user agent did not make a difference in my tests.


zas (approved):
Yes, agreed

ix5:
Thanks for notifying me @phw.
I had a look at https://github.com/metabrainz/picard-plugins/pull/172 and the issue appears to be something network- or Qt-related, which I don't know how to fix.

ix5:
See https://github.com/metabrainz/picard-plugins/pull/323

PR #318, opened 2021-11-06
Fix python string conversion error in abbreviate_artistsort.py

bleggett:
This fixes runtime errors logged in Picard due to invalid Python string conversion.


phw (approved):
Thanks

PR #320, opened 2021-12-28
Add multi-value variables for artist types and join phrases

rdswift:
As discussed in the [Community Forum](https://community.metabrainz.org/t/artist-variables-plugin/409544/13) this update provides the following new multi-value variables:

* **_artists_album_all_types** - All album artist types, as a multi-value
* **_artists_album_all_join_phrases** - All album artist join phrases, as a multi-value
* **_artists_track_all_types** - All track artist types, as a multi-value
* **_artists_track_all_join_phrases** - All track artist join phrases, as a multi-value

### DIFF
--- a/plugins/additional_artists_variables/additional_artists_variables.py
+++ b/plugins/additional_artists_variables/additional_artists_variables.py
@@ -30,7 +30,7 @@ this plugin.
 <br /><br />
 Please see the <a href="https://github.com/rdswift/picard-plugins/blob/2.0_RDS_Plugins/plugins/additional_artists_variables/docs/README.md">user guide</a> on GitHub for more information.
 '''
-PLUGIN_VERSION = '0.8'
+PLUGIN_VERSION = '0.8.1'
 PLUGIN_API_VERSIONS = ['2.0', '2.1', '2.2', '2.7']
 PLUGIN_LICENSE = 'GPL-2.0-or-later'
 PLUGIN_LICENSE_URL = 'https://www.gnu.org/licenses/gpl-2.0.html'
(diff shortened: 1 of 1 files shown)


rdswift:
I had to add a default value of **'[blank]'** for join phrases consisting of an empty string because Picard removes empty strings from a multi-value in [line 436 of picard.metadata.py](https://github.com/metabrainz/picard/blob/6ac0501f55f2df6dd5e5f032dbedfb226467665f/picard/metadata.py#L432-L441)::

    def set(self, name, values):
        name = self.normalize_tag(name)
        if isinstance(values, str) or not isinstance(values, Iterable):
            values = [values]
        values = [str(value) for value in values if value or value == 0]     # This is the line #
        if values:
            self._store[name] = values
            self.deleted_tags.discard(name)
        elif name in self._store:
            del self[name]

This seems like a bit of a hack, but can work in scripts if the user does an appropriate `$replace()` for the default value.

@zas / @phw do you know off hand if changing this to:

    def set(self, name, values):
        name = self.normalize_tag(name)
        if isinstance(values, str) or not isinstance(values, Iterable):
            values = [values]
        values = [str(value) for value in values if value or value == 0 or value == '']
        if values and (len(values) > 1 or values[0]):
            self._store[name] = values
            self.deleted_tags.discard(name)
        elif name in self._store:
            del self[name]

would break things, other than the two tests that I know fail (`test_cmd_setmulti_custom_splitter_string` and `test_cmd_setmulti_will_remove_empty_items`)?  Is there another (better) way to preserve empty string elements in a multi-value variable, or is that something that we just do **not** want to do?

rdswift:
Moved to PR #326 to avoid local branch sync issues.

PR #323, opened 2022-01-29
Re-add tangoinfo plugin

ix5:
This plugin has been ported to Picard 2.x.

It currently needs a workaround for a server misconfiguration for tango.info

### DIFF
--- a/plugins/tangoinfo/tangoinfo.py
+++ b/plugins/tangoinfo/tangoinfo.py
@@ -1,0 +1,360 @@
+# -*- coding: utf-8 -*-
+PLUGIN_NAME = "Tango.info Adapter"
+PLUGIN_AUTHOR = "Felix Elsner, Sambhav Kothari, Philipp Wolfer"
+PLUGIN_DESCRIPTION = """
+<p>Load genre, date and vocalist tags for latin dance music
+from <a href="https://tango.info">tango.info</a>.</p>
+"""
+PLUGIN_VERSION = "0.2.0"
+PLUGIN_API_VERSIONS = ["2.6", "2.7"]
+
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
+
+import re
+from functools import partial
+
+from picard import log
+from picard.util import LockableObject
+from picard.metadata import register_track_metadata_processor
+
+
+table_regex = re.compile(
+       r'<h2><a href="\/tracks?">Tracks?<\/a><\/h2>(?!<\/table>)(.+?)<\/table>'
+)  # Match the 'tracks'/'track' <table>
+tr_regex = re.compile(r"<tr>((?!</tr>).+?)</tr>")  # Match <tr> elements
+td_regex = re.compile(r"<td[^>]*>((?!</td>).+?)</td>")  # Match <td> elements
+tint_regex = re.compile(r"TINT:([0-9]+-[0-9]{1,2}-[0-9]{1,2})")
+
+TANGO_INFO_HOST = "tango.info"
+TANGO_INFO_PORT = 443
+
+
+# FIXME Remove all this
+# This is all just a hack to get around server issues at tango.info
+# The server will reply with an empty body if the Host header is set to
+# "tango.info:443", but just "tango.info" is fine.
+# Picard 2.8+ will by default not append the port to the Host header and thus
+# not need this workaround
+from picard import PICARD_VERSION
+from picard.version import Version
+host_workaround_needed = False
+if PICARD_VERSION < Version(2, 8, 0, 'dev', 1):
+    log.warning("%s: Picard version is older than 2.8, using workaround for "
+                "tango.info Host header issue", PLUGIN_NAME)
+    host_workaround_needed = True
+if host_workaround_needed:
+    from picard.webservice import WSGetRequest
+    from PyQt5.QtCore import QUrl
+    class WrappedWSGetRequest(WSGetRequest):  # noqa
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            url_without_port = QUrl(self.url().url())
+            # -1 is QUrl defaultPort
+            url_without_port.setPort(-1)
+            self.setUrl(url_without_port)
+
+
+class TangoInfoTagger:
+
+    class TangoInfoScrapeQueue(LockableObject):
+
+        def __init__(self):
+            LockableObject.__init__(self)
+            self.queue = {}
+
+        def __contains__(self, name):
+            return name in self.queue
+
+        def __iter__(self):
+            return self.queue.__iter__()
+
+        def __getitem__(self, name):
+            self.lock_for_read()
+            value = self.queue.get(name)
+            self.unlock()
+            return value
+
+        def __setitem__(self, name, value):
+            self.lock_for_write()
+            self.queue[name] = value
+            self.unlock()
+
+        def append(self, name, value):
+            self.lock_for_write()
+            if name in self.queue:
+                self.queue[name].append(value)
+                value = False
+            else:
+                self.queue[name] = [value]
+                value = True
+            self.unlock()
+            return value
+
+        def pop(self, name):
+            self.lock_for_write()
+            value = None
+            if name in self.queue:
+                value = self.queue[name]
+                del self.queue[name]
+            self.unlock()
+            return value
+
+    def __init__(self):
+        self.albumpage_cache = {}
+        self.albumpage_queue = self.TangoInfoScrapeQueue()
+
+    def add_tangoinfo_data(self, album, track_metadata, track, release):
+
+        # Look for BARCODE or barcode tag
+        if track_metadata.get("barcode"):
+            barcode = str(track_metadata["barcode"])
+        elif track_metadata.get("BARCODE"):
+            barcode = str(track_metadata["BARCODE"])
+        else:
+            # Abort if no barcode in track_metadata
+            return
+
+        # https://tango.info/wiki/TINT = <TINP>-<Side#>-<Track#>
+        # https://tango.info/wiki/TINP - "Tango Info Number for a Product" is a
+        # 14-digit numeric tango.info code used by tango.info and others.
+        # Example: TINT:00743216335725-1-5
+        # This plugin normalizes TINTs by stripping all zeros from the left
+        tint = "%s-%s-%s" % (
+            barcode.lstrip("0"),
+            str(track_metadata.get("discnumber", "1")),
+            str(track_metadata.get("tracknumber")),
+        )
+
+        if barcode in self.albumpage_cache:
+            if self.albumpage_cache[barcode].get(tint):
+                for field in ("genre", "date", "vocal"):
+                    # Do no overwrite with empty data
+                    if not self.albumpage_cache[barcode][tint].get(field):
+                        continue
+                    track_metadata[field] = \
+                        self.albumpage_cache[barcode][tint][field]
+            else:
+                log.debug(
+                    "%s: No information on tango.info for barcode %s",
+                    PLUGIN_NAME, barcode)
+        else:
+            self.website_add_track(album, album._new_tracks[-1], barcode, tint)
+
+    def website_add_track(self, album, track, barcode, tint, zeros=0):
+        """
+        :param zeros: Number of zeros that have been prepended to the barcode
+        """
+        self.album_add_request(album)
+
+        if self.albumpage_queue.append(barcode, (track, album, tint)):
+
+            # Barcode, zero-padded as needed
+            path = "/%s" % ("0" * zeros) + barcode
+
+            log.debug("%s: Downloading from tango.info: track %s, album %s "
+                      "with TINT %s from %s",
+                      PLUGIN_NAME, str(track), str(album), tint,
+                      "https://%s%s" % (TANGO_INFO_HOST, path))
+            # FIXME: Remove this
+            if host_workaround_needed:
+                ws = album.tagger.webservice
+                def get_patched(self, *args, **kwargs):  # noqa
+                    request = WrappedWSGetRequest(self, *args, **kwargs)
+                    return ws.add_request(request)
+                ws.get_patched = get_patched
+                return ws.get_patched(
+                    TANGO_INFO_HOST,
+                    TANGO_INFO_PORT,
+                    path,
+                    partial(self.website_process, barcode, zeros),
+                    priority=False,
+                    important=False,
+                    parse_response_type=None,
+                    queryargs=None,
+                )
+            # Call website_process() as a partial func
+            return album.tagger.webservice.get(
+                TANGO_INFO_HOST,
+                TANGO_INFO_PORT,
+                path,
+                partial(self.website_process, barcode, zeros),
+                priority=False,
+                important=False,
+                parse_response_type=None,
+                queryargs=None,
+            )
+
+    def website_process(self, barcode, zeros, response_bytes, reply, error):
+        """
+        response_bytes: PyQt5.QtCore.QByteArray, equals reply.readAll()
+        reply:          PyQt5.QtNetwork.QNetworkReply
+        error:          PyQt5.QtNetwork.QNetworkReply.NetworkError (optional)
+        """
+
+        if error:
+            log.warning("%s: Network error retrieving info for barcode %s",
+                        PLUGIN_NAME, barcode)
+            track_triple = self.albumpage_queue.pop(barcode)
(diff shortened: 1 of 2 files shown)


plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,346 @@
...
+
+        albuminfo = dict(genre=None, date=None, vocal=None)
+        page_structure_warned = False  # Ratelimit warnings
+
+        for tr in trcontent:
+            # Content inside of <td> elements
+            trackinfo = [m.groups()[0] for m in td_regex.finditer(tr)]
+
+            # Example of expected structure:
+            """
```
Sophist-UK: Better as a multi-line comment?

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,346 @@
...
+# FIXME Remove all this
+from picard.webservice import WSGetRequest
+from PyQt5.QtCore import QUrl
+class WrappedWSGetRequest(WSGetRequest):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        my_url = QUrl(self.url().url())
+        # HACK: Setting port=defaultPort will cause Host header sent to
+        # tango.info to change from `tango.info:443` to `tango.info`
+        my_url.setPort(-1)
```
phw: We could also omit default ports 80 / 433 from HTTP / HTTPS URLs in Picard already.

So in https://github.com/metabrainz/picard/blob/master/picard/util/__init__.py#L583-L589 just set the scheme based on the port, and only explicitly set the port if it differs from the default port.

@zas @Sophist-UK Do you see any downside doing this?

Apart from that the workaround here should stay for now if the plugin shall be compatible with existing released Picard versions, unless this gets fixed server side. It's a bit of a hack, but it is small code and cleanly done :)

phw (commented):
Thanks a lot for digging into this and figuring out this odd behavior caused by the explicit port.

Sophist's note about using a comment for the example data should be addressed, but otherwise this looks good to merge to me. Including the workaround. We can remove this anytime should this be fixed on tango.info.

plugins/tangoinfo/tangoinfo.py
Sophist-UK: The code in Picard that Philipp referred to is:
```
def build_qurl(host, port=80, path=None, queryargs=None):
    """
    Builds and returns a QUrl object from `host`, `port` and `path` and
    automatically enables HTTPS if necessary.
    Encoded query arguments can be provided in `queryargs`, a
    dictionary mapping field names to values.
    """
    url = QtCore.QUrl()
    url.setHost(host)
    url.setPort(port)

    if host in MUSICBRAINZ_SERVERS or port == 443:
        url.setScheme("https")
        url.setPort(443)
    else:
        url.setScheme("http")

    if path is not None:
        url.setPath(path)
    if queryargs is not None:
        url_query = QtCore.QUrlQuery()
        for k, v in queryargs.items():
            url_query.addQueryItem(k, str(v))
        url.setQuery(url_query)
    return url
```
I would argue that it is this code that is causing this problem. http / https implicitly define the port to be used and I don't think we need to set the port explicitly if they are standard. So if it was me I would:

1. Make the port an optional value.
2. Either only call `setport(port)` `if port is not None` (or if this causes problems `setport(-1)` `if port is None`.

My quick take on the Picard code would therefore be:
```
def build_qurl(host, port=None, path=None, queryargs=None):
    """
    Builds and returns a QUrl object from `host`, `port` and `path` and
    automatically enables HTTPS if necessary.
    Encoded query arguments can be provided in `queryargs`, a
    dictionary mapping field names to values.
    """
    url = QtCore.QUrl()
    url.setHost(host)
    if port and port not in (80, 443):
        url.setPort(port)
    if host in MUSICBRAINZ_SERVERS or port == 443:
        url.setScheme("https")
    else:
        url.setScheme("http")
    if path is not None:
        url.setPath(path)
    if queryargs is not None:
        url_query = QtCore.QUrlQuery()
        for k, v in queryargs.items():
            url_query.addQueryItem(k, str(v))
        url.setQuery(url_query)
    return url
``` 
I think that this would avoid the need for a hack completely.

ix5: I'm happy with whatever you decide. But note that for backward compatibility, the "hack" would need to stay in (behind a guard).
I'll try to verify which versions of Picard this plugin can run on successfully  and adjust accordingly.

One question, probably not all too important: Are there any services that will be accessed encrypted (https), but not on port 443?

plugins/tangoinfo/tangoinfo.py
ix5: My understanding was that python doesn't really have multi-line comments, requiring a `#` as a prefix for _each_ line.
So many sources recommend using empty strings. But I can see that linters would object.

ix5:
@phw would you mind cherry-picking the Github CI commit disabling py3.5 (or take whatever other action you deem appropriate)?
I don't think it should be part of this PR.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,357 @@
...
+# FIXME Remove all this
+# This is all just a hack to get around server issues at tango.info
+# The server will reply with an empty body if the Host header is set to
+# "tango.info:443", but just "tango.info" is fine.
+# Picard 2.8+ will by default not append the port to the Host header and thus
+# not need this workaround
+from picard import PICARD_VERSION
+from picard.version import Version
+host_workaround_needed = False
+#if PICARD_VERSION < Version(2, 8, 0, 'final', 0):  # noqa
```
ix5: @phw @Sophist-UK let me know which variant of version sniffing you prefer

plugins/tangoinfo/tangoinfo.py
ix5: In any case, changed to utilize `#`.

plugins/tangoinfo/tangoinfo.py
Sophist-UK, line 42: My opinion on this is probably not that relevant but since you ask...

`if PICARD_VERSION < Version(2, 8, 0, 'final', 0):`

Sophist-UK: My opinion on this is probably not that relevant but since you ask...

`if PICARD_VERSION < Version(2, 8, 0, 'final', 0):`

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,357 @@
...
+        trcontent = [match.groups()[0] for match in tr_regex.finditer(table)]
+
+        albuminfo = dict(genre=None, date=None, vocal=None)
+        page_structure_warned = False  # Ratelimit warnings
+
+        for tr in trcontent:
+            # Content inside of <td> elements
+            trackinfo = [m.groups()[0] for m in td_regex.finditer(tr)]
+
+            # Example of expected structure:
```
Sophist-UK, line 297: According to [PEP8](https://www.python.org/dev/peps/pep-0008/#block-comments) this is correct.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,357 @@
...
+                        self.albumpage_cache[barcode][tint][field]
+            else:
+                log.debug(
+                    "%s: No information on tango.info for barcode %s",
+                    PLUGIN_NAME, barcode)
+        else:
+            self.website_add_track(album, album._new_tracks[-1], barcode, tint)
+
+    def website_add_track(self, album, track, barcode, tint, zeros=0):
+        """
```
Sophist-UK, line 145: According to [PEP8](https://www.python.org/dev/peps/pep-0008/#documentation-strings) this is correct.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,357 @@
...
+                path,
+                partial(self.website_process, barcode, zeros),
+                priority=False,
+                important=False,
+                parse_response_type=None,
+                queryargs=None,
+            )
+
+    def website_process(self, barcode, zeros, response_bytes, reply, error):
+        # response_bytes: PyQt5.QtCore.QByteArray,
```
Sophist-UK: This looks like a doc-string to me, so according to [PEP8](https://www.python.org/dev/peps/pep-0008/#documentation-strings) this is incorrect.

plugins/tangoinfo/tangoinfo.py
Sophist-UK: Good question. AFAIK, you have to use 443 for SSL but if there is a server that requires https on a different port, there is no way with the current code or my proposed alternative to achieve that.

Sophist-UK (changes requested):
If you could tweak the comments again...

plugins/tangoinfo/tangoinfo.py
phw: See https://github.com/metabrainz/picard/pull/2057 for the changes in port handling in Picard's `build_qurl`

> One question, probably not all too important: Are there any services that will be accessed encrypted (https), but not on port 443?

As Sophist said this would be currently unsupported. This would require actually refactoring the picard.webservice API. Currently everything works by passing a hostname + port, but you cannot specify the scheme. Port 443 is assumed to be https, everything else gets http. But we should definitely refactor the API here at some point.

plugins/tangoinfo/tangoinfo.py
phw: I'd suggest using `PICARD_VERSION < Version(2, 8, 0, 'dev', 1)` . We can then bump the version once we have merged the fix

phw (changes requested):

phw:
> would you mind cherry-picking the Github CI commit disabling py3.5
@ix5 I did this, and also enabled testing with Python 3.9 + 3.10. After rebasing your branch should be mergable again.

plugins/tangoinfo/tangoinfo.py
ix5: I could only verify as far back as version `release-2.6.4`, everything else crashes on startup for me (Arch Linux, current as of 2022-01-30). So I've put API compatibilty to 2.6-2.7.

Tested with https://github.com/metabrainz/picard/pull/2057 and without (overriding version to `.dev0`), both handle fine.

Crash:
```
D: 16:50:51,470 /user/foo/picard/picard/ui/playertoolbar.__init__:90: Internal player: QtMultimedia available, initializing QMediaPlayer
D: 16:50:51,549 /user/foo/picard/picard/ui/playertoolbar.__init__:97: Internal player: available, QMediaPlayer set up
  File "/user/foo/picard/scripts/picard", line 5, in <module>
    main('/user/foo/picard/.venv-picard/share/locale', True)
  File "/user/foo/picard/picard/tagger.py", line 996, in main
    tagger = Tagger(picard_args, unparsed_args, localedir, autoupdate)
  File "/user/foo/picard/picard/tagger.py", line 285, in __init__
    self.window = MainWindow(disable_player=picard_args.no_player)
  File "/user/foo/picard/picard/ui/mainwindow.py", line 198, in __init__
    self.setupUi()
  File "/user/foo/picard/picard/ui/mainwindow.py", line 233, in setupUi
    self.cover_art_box = CoverArtBox(self)
  File "/user/foo/picard/picard/ui/coverartbox.py", line 315, in __init__
    self.cover_art = CoverArtThumbnail(False, True, self.pixmap_cache, parent)
  File "/user/foo/picard/picard/ui/coverartbox.py", line 74, in __init__
    self.shadow = self.shadow.scaled(w, h, QtCore.Qt.KeepAspectRatio, QtCore.Qt.SmoothTransformation)
TypeError: arguments did not match any overloaded call:
  scaled(self, int, int, aspectRatioMode: Qt.AspectRatioMode = Qt.IgnoreAspectRatio, transformMode: Qt.TransformationMode = Qt.FastTransformation): argument 1 has unexpected type 'float'
  scaled(self, QSize, aspectRatioMode: Qt.AspectRatioMode = Qt.IgnoreAspectRatio, transformMode: Qt.TransformationMode = Qt.FastTransformation): argument 1 has unexpected type 'float'

During handling of the above exception, another exception occurred:

  File "/user/foo/picard/.venv-picard/bin/picard", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/user/foo/picard/scripts/picard", line 9, in <module>
    from picard import crash_handler
ImportError: cannot import name 'crash_handler' from 'picard' (/user/foo/picard/picard/__init__.py)
```

plugins/tangoinfo/tangoinfo.py
ix5: Okay, using your suggestion.

plugins/tangoinfo/tangoinfo.py
ix5: Fixed

ix5:
> If you could tweak the comments again...

Done

> @ix5 I did this, and also enabled testing with Python 3.9 + 3.10. After rebasing your branch should be mergable again.

Rebased.

Also tested with https://github.com/metabrainz/picard/pull/2057 applied.

Thank you both for your review!

ix5:
Fixed a condition (introduced by my "too-clever-by-half" re-writing of the data-extract handling) where if either `genre`, `date` or `vocal` was missing there'd be a key error.

---

Just FYI, the CI check seems to still want to wait for a python 3.5 build ("some checks haven't completed yet"). I suspect that's something you'd have to turn off inside the GH UI somewhere.

Also, is there any effort to have (unit) tests for plugins I should be aware of? I've added a basic test (with a lot of assumptions) to my own plugin repo, see [tests.py](https://git.ix5.org/felix/picard-tangoinfo/src/commit/701725caa56021d5ffa57c314567047bd62913d6/tests.py)

ix5:
Hey @phw, is there anything else you'd like changed? I see https://github.com/metabrainz/picard/pull/2057 got merged.

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,360 @@
...
+                          PLUGIN_NAME, barcode, ("0" * zeros) + barcode,
+                          ("a zero" if zeros == 1 else "%d zeros" % zeros))
+
+            for track, album, tint in track_triple:
+                tm = track.metadata
+                if not self.albumpage_cache[barcode].get(tint):
+                    self.album_remove_request(album)
+                    continue
+
+                for field in ["genre", "date", "vocal"]:
```
zas: Just a minor thing, but this list could be declared as a tuple.
```suggestion
                for field in ("genre", "date", "vocal"):
```

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,360 @@
...
+                    # Write track metadata
+                    if self.albumpage_cache[barcode][tint].get(field):
+                        tm[field] = self.albumpage_cache[barcode][tint][field]
+
+                for file in track.iterfiles():
+                    fm = file.metadata
+                    for field in ["genre", "date", "vocal"]:
```
zas:
```suggestion
                    for field in ("genre", "date", "vocal"):
```

plugins/tangoinfo/tangoinfo.py
```diff
@@ -0,0 +1,360 @@
...
+
+        table = re.findall(table_regex, response)
+        if not table:
+            log.warning("%s: Could not extract table from album webpage - "
+                        "regex failed or page structure changed", PLUGIN_NAME)
+            return
+        table = table[0]  # re.findall() returns a list
+
+        # Content inside of <tr> elements
+        trcontent = [match.groups()[0] for match in tr_regex.finditer(table)]
```
zas: This could be made a generator:

```suggestion
        trcontent = (match.groups()[0] for match in tr_regex.finditer(table))
```

zas (changes requested):

plugins/tangoinfo/tangoinfo.py
ix5: Done

plugins/tangoinfo/tangoinfo.py
ix5: Done

plugins/tangoinfo/tangoinfo.py
ix5: Done

phw (approved):
@ix5 No, all good. I just did not get around to look at this again. Thanks for the reminder. I'll merge this, the plugin should than show up on the Picard website (and inside Picard) soon.

@zas Anything to add before I merge?

ix5:
Excellent! Thank you again for the review, and @zas for spotting the optimizations I overlooked.

zas (approved):

zas:
> @zas Anything to add before I merge?

Nope, LGTM.

PR #324, opened 2022-02-23
tangoinfo: Fix loading via renaming to __init__.py

ix5:
Oddly enough, `zipimport.zipimport` will succeed to import a file via `find_module()` with the following directory structure:
```
[...plugins]/tangoinfo/
└── tangoinfo.py
```
This, however, means that the generated zip archive contains more than one single file:
```
[...plugins]/tangoinfo/
├── README.md
└── tangoinfo.py
```
... and `zipimport` will fail to "find" anything.

The solution is to ~~add an intermediate `__init__` file that in turn imports the `tangoinfo.py` plugin file~~ rename the main file to  `__init__.py`.

---

Sorry that this slipped through. I wasn't aware of Picard's loading mechanism and assumed that the generated zips would work the same way as plunking the `tangoinfo.py` into the plugin file picker.

Other plugins also used e.g. `[...]/amazon/amazon.py` instead of `[...]/amazon/__init__.py`. As soon as you have more than one file though, the `zipimport` module [^1] goes haywire :-/

~~One might also rename the main file to `__init__`, but IMO that's pretty ugly and loses context when editing (try having 10 files all named `__init__.py` open in tabs and figure out which is which...)~~
**Edit:** Even having the stub `__init__.py` won't fix the issue. I hate that I'm supposed to do it this way, but I'm out of ideas.

[^1]: https://docs.python.org/3/library/zipimport.html


ix5:
For anyone wanting to verify or poke around a bit:
```sh
$ cd <picard-plugins>
python generate.py # don't forget this does a `git checkout -- plugins/` !
```

```python
import zipimport
import os

zip_importer = zipimport.zipimporter

# Compare current HEAD vs with this PR applied...
path = os.path.join(os.getcwd(), "build/tangoinfo.zip")
print(zip_importer(path).find_loader("tangoinfo"))
print(zip_importer(path).archive)

path = os.path.join(os.getcwd(), "build/amazon.zip")
print(zip_importer(path).find_loader("amazon"))
print(zip_importer(path).archive)
```

phw:
Hi, sorry for the late reply.

Yes, the current solution is the recommended one. The issue is that every plugin that consists of more than one file must be a package. The issue becomes apparent if instead of ZIP files we extract all the modules into the plugins folder. Then single file plugins can exist there as a single `{pluginname}.py` file. But if there is more than one file it needs to be a `{pluginname}` directory with an `__init__.py`  to make it a Python package.

We missed this in review because the plugin did not change. But the addition of the README was enough to trigger this change in packaging.

Thanks a lot for the fix

phw (approved):

PR #325, opened 2022-03-03
Fixes for "Generate M3U playlist" #245

### LINKED ISSUE #245: Fixes for "Generate M3U playlist"
I have noticed two things that the plugin "Generate M3U playlist" does which I would like changed. If these should be reported as separate issues, just say so and I'll be happy to do so.

1. When launched, the "Save new playlist" directory selection dialog defaults to a directory that I no longer use. I'd like for it to default to the directory where the selected album resides, if possible. Otherwise, just default to the last chosen directory, perhaps.
2. Regardless of the chosen directory for the creation of the playlist, it **always** uses absolute pathnames to the files. This needs to be a relative directory to the chosen one for the playlist. Or, at the very least, allow the user to choose somewhere in the settings. With absolute pathnames, if the album folder is ever moved to a different location, the playlist file is useless and must be recreated. Not cool.

For now, I still have to use Tag&Rename to create playlists because it always uses relative pathnames to the MP3 files. I look forward to the day when I don't have to rely upon Tag&Rename anymore and can use Picard for all of my music management work. Even if that means I learn Python and modify the existing code to make this work the way it needs to.

xjiro:
Fixes #245

The default playlist location also works for multiple album selections.

### DIFF
--- a/plugins/playlist/playlist.py
+++ b/plugins/playlist/playlist.py
@@ -16,7 +16,7 @@ PLUGIN_AUTHOR = "Francis Chin, Sambhav Kothari"
 PLUGIN_DESCRIPTION = """Generate an Extended M3U playlist (.m3u8 file, UTF8
 encoded text). Relative pathnames are used where audio files are in the same
 directory as the playlist, otherwise absolute (full) pathnames are used."""
-PLUGIN_VERSION = "1.1.1"
+PLUGIN_VERSION = "1.2"
 PLUGIN_API_VERSIONS = ["2.0"]
 PLUGIN_LICENSE = "GPL-2.0-or-later"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
@@ -133,8 +133,8 @@ class GeneratePlaylist(BaseAction):
                         if _debug_level > 1:
                             log.debug("{}: audio_filename: {}, selected dir: {}".format(
                                     PLUGIN_NAME, audio_filename, os.path.dirname(filename)))
-                        if os.path.dirname(filename) == os.path.dirname(audio_filename):
-                            audio_filename = os.path.basename(audio_filename)
+
+                        audio_filename = os.path.relpath(audio_filename, os.path.dirname(filename))
                         entry.add(str(audio_filename))
 
             playlist.write()
--- a/plugins/playlist/playlist.py
+++ b/plugins/playlist/playlist.py
@@ -12,7 +12,7 @@
 # GNU General Public License for more details.
 
 PLUGIN_NAME = "Generate M3U playlist"
-PLUGIN_AUTHOR = "Francis Chin, Sambhav Kothari"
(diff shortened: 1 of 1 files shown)


plugins/playlist/playlist.py
```diff
@@ -97,13 +103,14 @@ def callback(self, objs):
...
                     PLUGIN_NAME, default_filename))
+
         b_filename, b_selected_format = QtWidgets.QFileDialog.getSaveFileName(
             None, "Save new playlist",
             os.path.join(current_directory, default_filename),
             "Playlist (*.m3u8 *.m3u)"
         )
         if b_filename:
-            filename = string_(b_filename)
+            filename = str(b_filename)
```
phw: Not sure if that str conversion is actually needed anymore here. Is it?

plugins/playlist/playlist.py
```diff
@@ -76,9 +76,15 @@ class GeneratePlaylist(BaseAction):
...
     def callback(self, objs):
-        current_directory = (self.config.persist["current_directory"]
-                             or QtCore.QDir.homePath())
-        current_directory = find_existing_path(string_(current_directory))
+        # Find common path of all files to default where to save playlist
+        files = []
+        for album in objs:
+            for track in album.tracks:
+                if track.linked_files:
+                    files.append(track.linked_files[0].filename)
```
phw, line 84: That code here makes sense to keep backward compatibility and truly have this plugin still work in all Picard 2.x versions.

Just for completeness and if someone else reviews this, in Picard internall we would probably do something like this:

```
from picard.util import iter_files_from_objects

files = iter_files_from_objects(objs, save=True)
```

The save parameter will ensure that the files not attached to a track but to the unmatched files section of an album will not get returned.

But iter_files_from_objects only exists in newer versions (right now not sure when we added it, would need to check).

Also in general `track.linked_files` is kind of deprecated and only exists for backward compatibility. It is the same as `track.files` (for consistency with other objects having files as childs, e.g. Cluster). But again this was done in a later 2.x version.

phw, line 84: You could actually do 

```
from picard.util import iter_files_from_objects

files = iter_files_from_objects(objs, save=True)
```

The save parameter will ensure that the files not attached to a track but to the unmatched files section of an album will not get returned.

phw (approved):
Thanks for this. The changes make sense and definitely improve the usefulness of this plugin.

phw (changes requested):
@xjiro Can you please bump the plugin version to 1.2? That's needed before this PR can be 

Also about your change replacing `string_` with `str`, I think this str call is not needed at all there. This function was used to convert `QByteArray` or `bytes` to a string. But looking at other calls to e.g. `QFileDialog.getSaveFileName` in the Picard code this conversion no longer is necessary here.

phw (approved):
I rebased this PR, which essentially undid the str conversions (which got removed separately in https://github.com/metabrainz/picard-plugins/commit/239279cf1c06db4f4ee6374ddbbee92b2a0f2734 )

PR #326, opened 2022-04-02
Add multi-value variables for artist types and join phrases

rdswift:
Note:  This PR replaces #320 

As discussed in the [Community Forum](https://community.metabrainz.org/t/artist-variables-plugin/409544/13) this update provides the following new multi-value variables:

* **_artists_album_all_types** - All album artist types, as a multi-value
* **_artists_album_all_join_phrases** - All album artist join phrases, as a multi-value
* **_artists_track_all_types** - All track artist types, as a multi-value
* **_artists_track_all_join_phrases** - All track artist join phrases, as a multi-value

I had to add a default value of **'[blank]'** for join phrases consisting of an empty string because Picard removes empty strings from a multi-value in [line 436 of picard.metadata.py](https://github.com/metabrainz/picard/blob/6ac0501f55f2df6dd5e5f032dbedfb226467665f/picard/metadata.py#L432-L441)::

    def set(self, name, values):
        name = self.normalize_tag(name)
        if isinstance(values, str) or not isinstance(values, Iterable):
            values = [values]
        values = [str(value) for value in values if value or value == 0]     # This is the line #
        if values:
            self._store[name] = values
            self.deleted_tags.discard(name)
        elif name in self._store:
            del self[name]

This seems like a bit of a hack, but can work in scripts if the user does an appropriate `$replace()` for the default value.

@zas / @phw do you know off hand if changing this to:

    def set(self, name, values):
        name = self.normalize_tag(name)
        if isinstance(values, str) or not isinstance(values, Iterable):
            values = [values]
        values = [str(value) for value in values if value or value == 0 or value == '']
        if values and (len(values) > 1 or values[0]):
            self._store[name] = values
            self.deleted_tags.discard(name)
        elif name in self._store:
            del self[name]

would break things, other than the two tests that I know fail (`test_cmd_setmulti_custom_splitter_string` and `test_cmd_setmulti_will_remove_empty_items`)?  Is there another (better) way to preserve empty string elements in a multi-value variable, or is that something that we just do **not** want to do?

### DIFF
--- a/plugins/additional_artists_variables/additional_artists_variables.py
+++ b/plugins/additional_artists_variables/additional_artists_variables.py
@@ -30,7 +30,7 @@ this plugin.
 <br /><br />
 Please see the <a href="https://github.com/rdswift/picard-plugins/blob/2.0_RDS_Plugins/plugins/additional_artists_variables/docs/README.md">user guide</a> on GitHub for more information.
 '''
-PLUGIN_VERSION = '0.8'
+PLUGIN_VERSION = '0.8.1'
 PLUGIN_API_VERSIONS = ['2.0', '2.1', '2.2', '2.7']
 PLUGIN_LICENSE = 'GPL-2.0-or-later'
 PLUGIN_LICENSE_URL = 'https://www.gnu.org/licenses/gpl-2.0.html'
@@ -128,8 +128,8 @@ def process_artists(album_id, source_metadata, destination_metadata, source_type
             std_artist += temp_std_name + temp_phrase
             cred_artist += temp_cred_name + temp_phrase
             sort_artist += temp_sort_name + temp_phrase
-            artist_types.append(temp_type)
-            artist_join_phrases.append(temp_phrase)
+            artist_types.append(temp_type if temp_type else 'unknown',)
+            artist_join_phrases.append(temp_phrase if temp_phrase else '[blank]',)
             if temp_legal_name:
                 legal_artist += temp_legal_name + temp_phrase
                 legal_artist_list.append(temp_legal_name,)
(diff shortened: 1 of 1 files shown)


phw:
> would break things, other than the two tests that I know fail

Not sure, maybe people have some scripts where empty values can happen in multi value context. Like `$set(foo,)`  results in foo not being written to files I'd also expect that empty values do not get written to files.

So this probably means empty values in variables sometimes make sense (like here), but more often I'd not expect them to be part. What about maybe using a zero width space (`\u200B`) in this case?

rdswift:
> So this probably means empty values in variables sometimes make sense (like here), but more often I'd not expect them to be part. What about maybe using a zero width space (`\u200B`) in this case?

That probably works okay for display of the tags, but causes a problem if the tag is used in the file naming script, because Picard does not strip `\u200B` out of the path when saving (at least on my Windows 7 system).  Is this something that we might want to do by default (stripping `\u200B` from written paths)?

phw:
> Is this something that we might want to do by default (stripping `\u200B` from written paths)?

This might be actually useful. Not sure if someone has a use case of having a path with this character?

rdswift:
> Not sure if someone has a use case of having a path with this character?

The only thing that comes to mind is some rudimentary attempt to lock a directory or file so it is unavailable to those that don't know about it, which doesn't seem to make a lot of sense to me.

My suggestion is to go ahead with it and see if anyone complains.  In the worst case, we could always add yet another option setting in the File Naming section to keep it.  I'll have a look at it and see if I can come up with a PR.

phw (approved):

PR #327, opened 2022-04-02
Add Key Wheel Converter plugin

rdswift:
Based on the discussion in the [Community Forum](https://community.metabrainz.org/t/picard-acousticbrainz-music-key/579092) this plugin provides the ability to convert key information between 'camelot', 'open key' and 'standard' formats.
It adds three new scripting functions:

* **$key2camelot(key)** returns the key string `key` in camelot key format.
* **$key2openkey(key)** returns the key string `key` in open key format.
* **$key2standard(key\[,symbols\])** returns the key string `key` in standard key format.  If the optional argument `symbols` is set, then the '♭' and '#' symbols will be used, rather than spelling out '-Flat' and '-Sharp'.

The `key` argument can be entered in any of the supported formats, such as '2B' (camelot), '6d' (open key), 'A♭ Minor' (standard with symbols) or 'A-Flat Minor' (standard with text).  If the `key` argument is not recognized as one of the standard keys in the supported formats, then an empty string will be returned.

### DIFF
--- a/plugins/key_wheel_converter/key_wheel_converter.py
+++ b/plugins/key_wheel_converter/key_wheel_converter.py
@@ -22,11 +22,7 @@
 
 PLUGIN_NAME = 'Key Wheel Converter'
 PLUGIN_AUTHOR = 'Bob Swift'
-PLUGIN_DESCRIPTION = '''
-<p>
-Adds functions to convert between 'standard', 'camelot' and 'open key' key formats.
-</p>
-'''
+PLUGIN_DESCRIPTION = '''Adds functions to convert between 'standard', 'camelot' and 'open key' key formats.'''
 PLUGIN_VERSION = '1.0'
 PLUGIN_API_VERSIONS = ['2.3', '2.4', '2.6', '2.7']
 PLUGIN_LICENSE = "GPL-2.0"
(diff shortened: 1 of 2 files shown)


phw (approved):
This is a great plugin addition. Thanks

plugins/key_wheel_converter/key_wheel_converter.py
```diff
@@ -0,0 +1,424 @@
...
+
+# pylint: disable=C0413     (wrong-import-position)
+# pylint: disable=W0613     (unused-argument)
+
+PLUGIN_NAME = 'Key Wheel Converter'
+PLUGIN_AUTHOR = 'Bob Swift'
+PLUGIN_DESCRIPTION = '''
+<p>
+Adds functions to convert between 'standard', 'camelot' and 'open key' key formats.
+</p>
```
phw: Maybe a minor thing, but let's not add the <p> tags here. For one this does not really provide a value here with only a single paragraph. And also going forward I'd prefer to use Markdown (but that requires Picard >=2.7).

plugins/key_wheel_converter/key_wheel_converter.py
```diff
@@ -0,0 +1,424 @@
...
+            keys[item[i]] = {
+                'camelot': item[0],
+                'open': item[1],
+                'standard_s': item[2],
+                'standard_t': item[3],
+            }
+
+    def __init__(self):
+        """Class to hold the mapping dictionary.
+        """
```
phw: The constructor is not needed here, or is there some reason?

rdswift: Force of habit.  Now removed.

PR #329, opened 2022-04-24
Add persistent variables plugin

rdswift:
## Overview

This plugin provides the ability to store and retrieve script variables that persist across tracks and albums.  This allows things like finding and storing the earliest recording date of all of the tracks on an album.

There are two types of persistent variables maintained - album variables and session variables. Album variables persist across all tracks on an album.  Each album's information is stored separately, and is reset when the album is refreshed. The information is cleared when an album is removed.  Session variables persist across all albums and tracks, and are cleared when Picard is shut down or restarted.

## Scripting Functions Added

This plugin adds eight new scripting functions to allow management of persistent script variables:

### $set_a(name,value)

Sets the album persistent variable `name` to `value`.

### $unset_a(name)

Unsets the album persistent variable `name`.

### $get_a(name)

Gets the album persistent variable `name`.

### $clear_a()

Clears all album persistent variables.

### $set_s(name,value)

Sets the session persistent variable `name` to `value`.

### $unset_s(name)

Unsets the session persistent variable `name`.

### $get_s(name)

Gets the session persistent variable `name`.

### $clear_s()

Clears all session persistent variables.

## Persistent Variables Viewer

You can view the persistent variables currently set by right-clicking on a file, track or album and selecting "View persistent variables" from the Plugins section of the context menu.  This will display the variables associated with the selected album as well as all session variables.

### Example 1:

Find the disc and album playing lengths, so that they can be included in the file metadata and used in the file naming script.

Create a tagging script containing the following:

```
$if($get_a(_pt_%discnumber%_%tracknumber%),,
  $if(%_length%,
    $set(_pt_pos,$find(%_length%,:))
    $if(%_pt_pos%,
      $set(_pt_m,$add($if2($left(%_length%,%_pt_pos%),0),0))
      $set(_pt_s,$add($if2($substr(%_length%,$add(%_pt_pos%,1)),0),0))
      $set_a(pt_d%discnumber%_d,$if2($get_a(pt_d%discnumber%_d),0))
      $set_a(pt_d%discnumber%_h,$if2($get_a(pt_d%discnumber%_h),0))
      $set_a(pt_d%discnumber%_m,$add($if2($get_a(pt_d%discnumber%_m),0),%_pt_m%))
      $set_a(pt_d%discnumber%_s,$add($if2($get_a(pt_d%discnumber%_s),0),%_pt_s%))
      $set_a(pt_r_d,$if2($get_a(pt_r_d),0))
      $set_a(pt_r_h,$if2($get_a(pt_r_h),0))
      $set_a(pt_r_m,$add($if2($get_a(pt_r_m),0),%_pt_m%))
      $set_a(pt_r_s,$add($if2($get_a(pt_r_s),0),%_pt_s%))
      $if($gt($get_a(pt_d%discnumber%_s),59),
        $set_a(pt_d%discnumber%_m,$add($get_a(pt_d%discnumber%_m),1))
        $set_a(pt_d%discnumber%_s,$sub($get_a(pt_d%discnumber%_s),60))
      )
      $if($gt($get_a(pt_r_s),59),
        $set_a(pt_r_m,$add($get_a(pt_r_m),1))
        $set_a(pt_r_s,$sub($get_a(pt_r_s),60))
      )
      $if($gt($get_a(pt_d%discnumber%_m),59),
        $set_a(pt_d%discnumber%_h,$add($get_a(pt_d%discnumber%_h),1))
        $set_a(pt_d%discnumber%_m,$sub($get_a(pt_d%discnumber%_m),60))
      )
      $if($gt($get_a(pt_r_m),59),
        $set_a(pt_r_h,$add($get_a(pt_r_h),1))
        $set_a(pt_r_m,$sub($get_a(pt_r_m),60))
      )
      $if($gt($get_a(pt_d%discnumber%_h),23),
        $set_a(pt_d%discnumber%_d,$add($get_a(pt_d%discnumber%_d),1))
        $set_a(pt_d%discnumber%_h,$sub($get_a(pt_d%discnumber%_h),24))
      )
      $if($gt($get_a(pt_r_h),23),
        $set_a(pt_r_d,$add($get_a(pt_r_d),1))
        $set_a(pt_r_h,$sub($get_a(pt_r_h),24))
      )
      $set_a(_pt_%discnumber%_%tracknumber%,1)
    )
  )
)
```

Now to make the times available as tags for all tracks, you need to create another tagging script which should be run manually just before saving the files.  This script should contain the following:

```
$set(_temp,)
$if($gt($get_a(pt_d%discnumber%_d),0),$set(_temp,$get_a(pt_d%discnumber%_d)d $get_a(pt_d%discnumber%_h)h $get_a(pt_d%discnumber%_m)m $get_a(pt_d%discnumber%_s)s),
  $if($gt($get_a(pt_d%discnumber%_h),0),$set(_temp,$get_a(pt_d%discnumber%_h)h $get_a(pt_d%discnumber%_m)m $get_a(pt_d%discnumber%_s)s),
    $if($gt($get_a(pt_d%discnumber%_m),0),$set(_temp,$get_a(pt_d%discnumber%_m)m $get_a(pt_d%discnumber%_s)s),
      $set(_temp,$get_a(pt_d%discnumber%_h)h $get_a(pt_d%discnumber%_m)m $get_a(pt_d%discnumber%_s)s)
    )
  )
)
$set(Disc Playing Time,%_temp%)

$set(_temp,)
$if($gt($get_a(pt_r_d),0),$set(_temp,$get_a(pt_r_d)d $get_a(pt_r_h)h $get_a(pt_r_m)m $get_a(pt_r_s)s),
  $if($gt($get_a(pt_r_h),0),$set(_temp,$get_a(pt_r_h)h $get_a(pt_r_m)m $get_a(pt_r_s)s),
    $if($gt($get_a(pt_r_m),0),$set(_temp,$get_a(pt_r_m)m $get_a(pt_r_s)s),
      $set(_temp,$get_a(pt_r_h)h $get_a(pt_r_m)m $get_a(pt_r_s)s)
    )
  )
)
$set(Release Playing Time,%_temp%)
```

The disc and release playing times will now be stored in each of the track files under the tag names `Disc Playing Time` and `Release Playing Time`.  These tags are also available for use in the file naming script.

### Example 2:

Find the earliest recording date of any of the tracks on the album, so that it can be used in the file naming script.

Create a tagging script containing the following:

```
$set(_testdate1,$if2($get_a(earliest_date),9999))
$set_a(earliest_date,$min(%_testdate1%,$if2(%_recording_firstreleasedate%,9999)))
```

The date for the album is now available in the file naming script using `$get_a(earliest_date)`.

### Example 3:

Get a list of all tracks on the album so that it can be included as a tag in each of the tracks.

Create a tagging script containing the following:

```
$set_a(all_tracks,$if($get_a(all_tracks),; )%tracknumber%. %title%)
```

Now to make the combined list available as a tag for all tracks, you need to create another tagging script after the previous one containing the following:

```
$set(all_tracks,$get_a(all_tracks))
```

The complete track list will now be stored in each of the track files under the tag name `all_tracks`.

### DIFF
--- a/plugins/persistent_variables/persistent_variables.py
+++ b/plugins/persistent_variables/persistent_variables.py
@@ -109,6 +109,13 @@ class PersistentVariables:
         return cls.session_variables[key] if key in cls.session_variables else ""
 
 
+def _get_album_id(parser):
+    file = parser.file
+    if file and file.parent and hasattr(file.parent, 'album') and file.parent.album:
+        return str(file.parent.album.id)
+    return ""
+
+
 def func_set_s(parser, name, value):
     if value:
         PersistentVariables.set_session_var(normalize_tagname(name), value)
@@ -132,25 +139,30 @@ def func_clear_s(parser):
 
 
 def func_unset_a(parser, name):
-    album_id = parser.context['musicbrainz_albumid']
-    PersistentVariables.unset_album_var(album_id, normalize_tagname(name))
+    album_id = _get_album_id(parser)
+    if album_id:
+        PersistentVariables.unset_album_var(album_id, normalize_tagname(name))
     return ""
 
 
 def func_set_a(parser, name, value):
-    album_id = parser.context['musicbrainz_albumid']
-    PersistentVariables.set_album_var(album_id, normalize_tagname(name), value)
+    album_id = _get_album_id(parser)
+    if album_id:
+        PersistentVariables.set_album_var(album_id, normalize_tagname(name), value)
     return ""
 
 
 def func_get_a(parser, name):
-    album_id = parser.context['musicbrainz_albumid']
-    return PersistentVariables.get_album_var(album_id, normalize_tagname(name))
+    album_id = _get_album_id(parser)
+    if album_id:
+        return PersistentVariables.get_album_var(album_id, normalize_tagname(name))
+    return ""
 
 
 def func_clear_a(parser):
-    album_id = parser.context['musicbrainz_albumid']
-    PersistentVariables.clear_album_vars(album_id)
+    album_id = _get_album_id(parser)
+    if album_id:
+        PersistentVariables.clear_album_vars(album_id)
     return ""
 
 
--- a/plugins/persistent_variables/persistent_variables.py
+++ b/plugins/persistent_variables/persistent_variables.py
@@ -17,7 +17,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 # 02110-1301, USA.
 
-PLUGIN_NAME = 'Persistent Variabless'
+PLUGIN_NAME = 'Persistent Variables'
 PLUGIN_AUTHOR = 'Bob Swift (rdswift)'
 PLUGIN_DESCRIPTION = '''
 <p>
@@ -111,9 +111,13 @@ class PersistentVariables:
 
 def _get_album_id(parser):
     file = parser.file
-    if file and file.parent and hasattr(file.parent, 'album') and file.parent.album:
-        return str(file.parent.album.id)
-    return ""
+    if file:
+        if file.parent and hasattr(file.parent, 'album') and file.parent.album:
+            return str(file.parent.album.id)
+        else:
+            return ""
+    # Fall back to parser context to allow processing on albums newly retrieved from MusicBrainz
+    return parser.context['musicbrainz_albumid']
 
 
 def func_set_s(parser, name, value):
--- a/plugins/persistent_variables/persistent_variables.py
+++ b/plugins/persistent_variables/persistent_variables.py
@@ -1,0 +1,211 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 Bob Swift (rdswift)
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.
+
+PLUGIN_NAME = 'Persistent Variabless'
+PLUGIN_AUTHOR = 'Bob Swift (rdswift)'
+PLUGIN_DESCRIPTION = '''
+<p>
+This plugin provides the ability to store and retrieve script variables that persist across tracks and albums.
+This allows things like finding and storing the earliest recording date of all of the tracks on an album.
+</p><p>
+There are two types of persistent variables maintained - album variables and session variables. Album variables
+persist across all tracks on an album.  Each album's information is stored separately, and is reset when the
+album is refreshed. The information is cleared when an album is removed.  Session variables persist across all
+albums and tracks, and are cleared when Picard is shut down or restarted.
+</p><p>
+This plugin adds eight new scripting functions to allow management of persistent script variables:
+<ul>
+<li>$set_a(name,value) : Sets the album persistent variable name to value.</li>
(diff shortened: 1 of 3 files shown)


plugins/persistent_variables/persistent_variables.py
```diff
@@ -0,0 +1,211 @@
...
+    return PersistentVariables.get_session_var(normalize_tagname(name))
+
+
+def func_clear_s(parser):
+    PersistentVariables.clear_session_vars()
+    return ""
+
+
+def func_unset_a(parser, name):
+    album_id = parser.context['musicbrainz_albumid']
```
phw: I wonder if it would be better to use `parser.file.parent.album.id` instead. Needs some care checking whether it is available (see e.g. `func_matchedtracks`) but it would ensure these functions really only works in album context and that then the ID is always set. Here the user might `$unset(musicbrainz_albumid)`, or have it set on a file.

In any case, it needs to be checked whether there is an album_id. I think the `*_a` functions should do nothing in case there is no album.

phw (changes requested):
This looks good. Especially on album level I can see how this is useful, and the registered plugin hooks guarantee that you start with an empty list.

The conceptional problem I still see is that many use cases that e.g. aggregate data over all tracks will have the result only available after automatic scripts have been run. Hence the user needs an additional manual run of a script to make use of it. But as I wrote before I still think having this plugin can be useful to explore these use cases.

I could imagine that we extend Picard at some point with the ability to configure at what point scripts are being run. E.g. instead of only having it run on load or not at all, maybe there could also be kind of a "album aggregation" step. This script would run on all tracks prior to the normal loading scripts. It would not alter the tracks metadata, but values gatherd during its run are persisted and available in the loading script run later. Just a spontaneous idea, haven't fully thought this through ;)

rdswift:
> The conceptional problem I still see is that many use cases that e.g. aggregate data over all tracks will have the result only available after automatic scripts have been run. Hence the user needs an additional manual run of a script to make use of it. But as I wrote before I still think having this plugin can be useful to explore these use cases.

Is the processing order of tagging scripts not consistent and predictable as described in [the documentation](https://picard-docs.musicbrainz.org/en/extending/scripts.html)?  Of course, that doesn't help if selected scripts are executed manually.  ~~I'll give some thought to trying to create a function to execute a specified script against all tracks in an album.  That might take care of the situation.~~ **EDIT:** I don't know what I was thinking.  That functionality already exists by running a script from the context menu on the release.  As you said, it needs to be run manually, but it definitely works.

> I could imagine that we extend Picard at some point with the ability to configure at what point scripts are being run. E.g. instead of only having it run on load or not at all, maybe there could also be kind of a "album aggregation" step. This script would run on all tracks prior to the normal loading scripts. It would not alter the tracks metadata, but values gatherd during its run are persisted and available in the loading script run later. Just a spontaneous idea, haven't fully thought this through ;)

That definitely sounds interesting.  I'll have to let that percolate through my mind for a while to see how it might be implemented.

plugins/persistent_variables/persistent_variables.py
rdswift: > I wonder if it would be better to use `parser.file.parent.album.id` instead. Needs some care checking whether it is available (see e.g. `func_matchedtracks`) but it would ensure these functions really only works in album context and that then the ID is always set. Here the user might `$unset(musicbrainz_albumid)`, or have it set on a file.

Good point.  I'll make the change.

> In any case, it needs to be checked whether there is an album_id. I think the `*_a` functions should do nothing in case there is no album.

Agreed.  I was thinking that this was covered in the `if album:` tests in the class methods, but using the `id` of the `album` class in the script function definitions makes sense that there are also checks in the function definitions.

rdswift: > > I wonder if it would be better to use `parser.file.parent.album.id` instead. Needs some care checking whether it is available (see e.g. `func_matchedtracks`) but it would ensure these functions really only works in album context and that then the ID is always set. Here the user might `$unset(musicbrainz_albumid)`, or have it set on a file.
> 
> Good point. I'll make the change.

I included a fallback to using the `parser.context` to identify the album id if there was no `parser.file` object in order to allow processing of albums newly retrieved from MusicBrainz or when a file with no `parser.file.parent.album.id` was matched (dropped) to a track in the right-hand pane.

rdswift: I also added a new, more likely in the real world, example to the description (as Example 1).  I tested it here and it seems to work well.

phw (approved):
This is a very comete plugin, great addition with the dialog. lGTM

PR #330, opened 2022-04-30
Fix link to user guide

rdswift:
The link to the user guide is incorrect -- pointing to a different plugin.  The hazards of copy'n'paste.  Argh!


phw (approved):
Missed that :)

PR #331, opened 2022-05-25
Update AcousticBrainz Tags with configurable tag names

regorxxx:
Adds 2 edit boxes for mood and genre tag names. 

image

This should solve most issues with tag merging (for example with LastFM mood tag or genre tags provided by picard), since merging tags is trivial via script.

> $copymerge(mood,mood_lastfm) 
> $copymerge(mood,ab:mood) 
> $delete(ab:mood) 
> $delete(mood_lastfm) 

Also fixed author names, now point to email properly.

image

### DIFF
--- a/plugins/acousticbrainz/__init__.py
+++ b/plugins/acousticbrainz/__init__.py
@@ -22,10 +22,10 @@
 # =============================================================================
 
 PLUGIN_NAME = 'AcousticBrainz Tags'
-PLUGIN_AUTHOR = ('<a href="mailto:wargreen@lebib.org">Wargreen</a>, '
-                 '<a href="mailto:pistache@lebib.org">Hugo Geoffroy "pistache"</a>, '
-                 '<a href="mailto:ph.wolfer@gmail.com">Philipp Wolfer</a>, '
-                 '<a href="mailto:regorxxx@protonmail.com">Regorxxx</a>')
+PLUGIN_AUTHOR = ('Wargreen <wargreen@lebib.org>, '
+                 'Hugo Geoffroy "pistache" <pistache@lebib.org>, '
+                 'Philipp Wolfer <ph.wolfer@gmail.com>, '
+                 'Regorxxx <regorxxx@protonmail.com>')
 PLUGIN_DESCRIPTION = '''
 Tag files with tags from the AcousticBrainz database, all highlevel classifiers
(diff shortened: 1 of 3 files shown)


plugins/acousticbrainz/__init__.py
```diff
@@ -22,9 +22,10 @@
 # =============================================================================
 
 PLUGIN_NAME = 'AcousticBrainz Tags'
-PLUGIN_AUTHOR = ('Wargreen <wargreen@lebib.org>, '
-                 'Hugo Geoffroy "pistache" <pistache@lebib.org>, '
-                 'Philipp Wolfer <ph.wolfer@gmail.com>')
+PLUGIN_AUTHOR = ('<a href="mailto:wargreen@lebib.org">Wargreen</a>, '
+                 '<a href="mailto:pistache@lebib.org">Hugo Geoffroy "pistache"</a>, '
+                 '<a href="mailto:ph.wolfer@gmail.com">Philipp Wolfer</a>, '
+                 '<a href="mailto:regorxxx@protonmail.com">Regorxxx</a>')
```
phw: Please don't do this, this field is not supposed to hold HTML. If the e-mail does not show up in Picard right now that should be fixed there.

phw (changes requested):
General functionality looks good to me, but please change the authors back.

Overall the time for this plugin likely is counted. Depends on how long the AB API stays available. But as long as it is available we can of course provide the plugin.

plugins/acousticbrainz/__init__.py
regorxxx: Works fine within picard, the same the description field does (there are multiple plugins there with html on that field).
As it is now the emails are simply hidden and replaced with a space char which makes no sense at all. Is it intended to be fixed then?

phw: Description is ok. In current MB this can contain Markdown (which allows also some HTML). But author is meant to contain plain text, I'd consider this a bug that it tries to render this as HTML (and hides the <mail> syntax due to this by default).

I'll get this fixed. We can then also make <mail> be automatically linked then. I think most plugins just used author names and not e-mails, so this did not get noticed.

phw (approved):

phw:
Thanks for the addition, merged

PR #333, opened 2022-07-04
Add Genre Mapper plugin

rdswift:
This plugin provides the ability to standardize genres in the "genre" tag by matching the genres as found to a standard genre as defined in the genre replacement mapping configuration option. Once installed a settings page will be added to Picard's options, which is where the plugin is configured.

This plugin is set to run at low priority so that any other plugins that might affect the list of genres are processed first.

The settings panel allows the user to provide a list of the original/replacement pairs used to modify the genres provided in the "genre" tag. Each pair must be entered on a separate line in the form:

```
[genre match test string]=[replacement genre]
```

Supported wildcards in the test string part of the mapping include '*' and '?' to match any number of characters and a single character respectively.  Blank lines and lines beginning with an equals sign (=) will be ignored. If the replacement part of the pair is blank, any matching genres will be removed. Case-insensitive tests are used when matching. Replacements will be made in the order they are found in the list.

There is also a setting which allows the user to choose whether or not to apply the first matching pair only, or continue processing the remaining pairs with the updated genre.  By default, all of the pairs are processed.

image

----

~~**QUESTION:** The plugin was developed to provide a simple method for defining the match test strings.  Would it be beneficial to add an option to allow the user to enter regular expressions instead, which could provide more control for users that understand the use of regular expressions?~~


rdswift:
It was a fairly easy change, so I added the option to use regular expressions for the match strings.  This is disabled by default.

image

plugins/genre_mapper/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def track_genre_mapper(album, metadata, *args):
+    if not config.setting[OPT_MATCH_ENABLED]:
+        return
+    if 'genre' not in metadata or not metadata['genre']:
+        log.debug("%s: No genres found for: \"%s\"", PLUGIN_NAME, metadata['title'],)
+        return
+    genres = set()
+    metadata_genres = str(metadata['genre']).split(MULTI_VALUED_JOINER)
```
phw, line 152: This might actually be a bit difficult to get right all the time, because the popular genre plugins can be configured to use different joiners. Even the built-in genres might use `join_genres`  joiner to generate a single value field. If not set, the tags are stored multi-valued.

I don't have a really good idea on how to handle that. The fully proper solution would still be to refactor the genre handling inside Picard. How I imagine it there would be single genre settings inside Picard (for defining how many genres max. should be used, the joiner, genre filters and min. usage). And then plugins would just provide a list of genres + maybe count (for the community sourced genres).

Implemented that way the functionality of this plugin of genre mappings would likely also need to be part of main Picard.

Anyway, until we have this in Picard I would be happy to have this plugin work the way you did it here if it works for many and provides an improvement.

phw (approved):

plugins/genre_mapper/__init__.py
rdswift: > This might actually be a bit difficult to get right all the time, because the popular genre plugins can be configured to use different joiners. Even the built-in genres might use `join_genres` joiner to generate a single value field. If not set, the tags are stored multi-valued.

I'm not really into genres, so I didn't realize that.  You're right, this could be a problem.

> I don't have a really good idea on how to handle that.

Since I already have a UI for settings anyway, what do you think about adding a setting to (optionally) specify a joiner to use for splitting the genres?  Perhaps I could even allow multiple joiners to be specified and try splitting on each in turn.

> The fully proper solution would still be to refactor the genre handling inside Picard. How I imagine it there would be single genre settings inside Picard (for defining how many genres max. should be used, the joiner, genre filters and min. usage). And then plugins would just provide a list of genres + maybe count (for the community sourced genres).
> 
> Implemented that way the functionality of this plugin of genre mappings would likely also need to be part of main Picard.

When I get a few minutes I'll take a look at the current code in Picard and see if I can come up with something to support your vision (which makes perfect sense to me).

> Anyway, until we have this in Picard I would be happy to have this plugin work the way you did it here if it works for many and provides an improvement.

Thanks.

Issue #209, opened 2019-02-22
ReplayGain TypeError when calculating gain


djm2k:
ReplayGain Logs this error upon trying to calculate for an album:

> E: 11:46:04,011 util.thread.run:56: Traceback (most recent call last):
>   File "picard\util\thread.py", line 52, in run
>   File "...\AppData\Local\MusicBrainz\Picard\plugins\replaygain.zip\replaygain\__init__.py", line 120, in _calculate_albumgain
>     calculate_replay_gain_for_files(files, format_, self.tagger)
>   File "...\AppData\Local\MusicBrainz\Picard\plugins\replaygain.zip\replaygain\__init__.py", line 43, in calculate_replay_gain_for_files
>     decode_filename(b' '.join(file_list)))
> TypeError: sequence item 0: expected a bytes-like object, str found

harvest316:
I have this issue as well.  Even occurs when calculating for a single MP3 track.

phw:
This was already fixed with #290 a while back.

Also there is a new replaygain2 plugin (which will be available for download in a few hours) and the old replaygain plugin is considered deprecated and will be removed in the future. See #336 for details.

.

Issue #121, opened 2017-11-01
support mp3gain alternative


bgemmill:
Since mp3gain was dropped from debian due to being unmaintained, from Ubuntu 16.04 onwards it's only available from unofficial sources:
https://launchpad.net/~flexiondotorg/+archive/ubuntu/audio

Of the alternatives, it looks like bs1770gain is the most recently maintained as of 2017:
http://bs1770gain.sourceforge.net/

I would have suggested rgain as well, but that also looks abandoned as of mid 2016:
https://pypi.python.org/pypi/rgain

phw:
The new replaygain2 plugin (which will be available for download in a few hours) uses the more modern [rsgain](https://github.com/complexlogic/rsgain) utility. See #336 for details.

 The old replaygain plugin is considered deprecated and will be removed in the future.

Issue #89, opened 2017-03-24
Calculated ReplayGain values do not show up in Tag Editor


Holzhaus:
When using the ReplayGain plugin, the calculated values are not listed in the Tag Editor area at the bottom of the Picard window. A user would expect that the new values show up in green color on the right side.

mineo:
This is because the tags from the files are not immediately reloaded after the replaygain calculation utility has finished.

Holzhaus:
IMHO it would be better if ReplayGain tags are not saved until the user hits the "Save" button.

When using MP3 files, this is possible by invoking `mp3gain` with the `-s s` argument, so that tags are not changed:

```
$ mp3gain -s s -o -q *.mp3 
File	MP3 gain	dB gain	Max Amplitude	Max global_gain	Min global_gain
somefile1.mp3	-5	-7.290000	34802.173241	255	142
somefile2.mp3	-5	-7.250000	36258.364289	255	147
somefile3.mp3	-5	-8.180000	37765.563247	255	146
somefile4.mp3	-6	-9.350000	36827.307860	255	141
"Album"	-5	-7.910000	37765.563247	255	141
```

The plugin can parse the output and update the tag editor.

Likewise, we could use `metaflac --scan-replay-gain`, `wvgain -a -d` and `vorbisgain -a -d` to display the calculated values instead of writing them directly.

charlydelta:
I second that, @Holzhaus. Are there any plans to address this? It seems wrong to me that the plugin just bypasses picard and writes the files on its own.

My whole music library is synchronized across several devices using nextcloud, I carefully review changes made to the files, because just touching a file (i.e., writing it without semantically changing any tag) forces a re-sync with nextcloud. I cannot do this with the replaygain plugin because it will just brute-force changes instead of allowing me to review if anything even changed to given files.

phw:
@charlydelta There have been discussions about this, see e.g. https://community.metabrainz.org/t/help-with-replaygain-plugin/446469/5 . I hope @Sophist-UK finds the time for it.

charlydelta:
Great, thanks for the link! Yes would be really awesome, but I understand that the current workflow is convenient for many.

Sophist-UK:
As and when I have time, my plan is to plagiarise (with permission) the [regainer](https://github.com/kepstin/regainer) code and use @phw 's ffmpeg image in order to calculate the replay gain values and put them into the correct Picard variables so that they are only saved when you click Save in Picard.

[...]

PR #336, opened 2022-09-25
New ReplayGain Plugin

### LINKED ISSUE #335: [Proposal] New Backend for ReplayGain Plugin
I am the developer of [rsgain](https://github.com/complexlogic/rsgain), a cross-platform Windows/Mac/Linux ReplayGain tagging utility. I propose to make the Picard ReplayGain plugin use rsgain as the sole backend instead of the existing list of backends. This would have several benefits:
- Users don't have to download and manage multiple programs for each file format they have in their library
- Having a single backend results in metadata that is consistent across all file formats
- rsgain supports several file formats that aren't currently supported by the plugin. Notably, M4A support is missing, which is a popular file format
- For Windows, it can properly handle Unicode paths when run on Windows 10 or later

If this proposal is acceptable, I will work on a PR to update the plugin. Please let me know what you think.

complexlogic:
Following up on the discussion in #335, I created a ReplayGain plugin that uses [rsgain](https://github.com/complexlogic/rsgain) as its sole backend. All of the noted issues with the current ReplayGain plugin have been resolved:
- The backend only calculates the gain/peak values. It does not actually tag the files. The plugin parses the output from the backend and stores the values in the track's `Metadata` object, so the user can view the result in the metadata window before saving. Picard is responsible for tagging the files
- On Windows, the plugin no longer opens an empty console window
- File paths containing Unicode characters work on Linux/Mac, and on Windows 10 and later

I also added a basic progress update to the status message when a user is scanning multiple albums.

The new plugin extends support to the following file formats (in addition to what the current plugin supports):
- M4A
- Opus
- WMA
- WAV
- Speex
- MP2
- Monkey's Audio
- AIFF

## Settings
The new plugin contains a settings menu which allows advanced users a high level of control over the scan settings.

For example, a user could choose a profile that targets the EBU R128 standard by setting the target loudness to -23 LUFS and true peak calculations. The settings also allow users to chose between ReplayGain tags and the RFC 7845-complaint `R128_*_GAIN` tags for Opus file.

screen32

## New Plugin or Replace Existing?
In #335, there was some discussion whether this should replace the existing ReplayGain plugin, or be a new plugin. After some thought, I think this should be a new plugin because this change is too disruptive for users of the existing plugin - even for a major version update. I have decided to call the new plugin "ReplayGain 2.0" because it implements the [ReplayGain 2.0 specification](https://wiki.hydrogenaud.io/index.php?title=ReplayGain_2.0_specification). This name makes it distinct from the current plugin while still clear that ReplayGain is the subject.

## To Do
- I still need to write the description of the plugin, but I would like to get some feedback on the code before working on that.
- Should we add a deprecation notice to the description of the current ReplayGain plugin?

Fixes #335


Sophist-UK:
I haven't tested this myself, but jolly well done for creating this.

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,336 @@
+# -*- coding: utf-8 -*-
+PLUGIN_NAME = "ReplayGain 2.0"
+PLUGIN_AUTHOR = "Philipp Wolfer, Sophist, complexlogic"
```
phw: I don't think I contributed much to this plugin really :)

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
+
+from functools import partial
+import subprocess
+import shutil
+import os
+
+from picard.album import Album
+from picard.track import Track
```
phw: `Album`  and `Track` are unused

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+def parse_result(line, tagger):
+    result = dict()
+    columns = line.strip("\t").split("\t")
+    tagger.log.debug(columns)
+
+    if len(columns) != len(TABLE_HEADER):
+        return None
+    for i, column in enumerate(columns):
+        result[TABLE_HEADER[i]] = column
+    return result
```
phw, line 132: Just an idea maybe for rsgain itself: What about providing output in e.g. JSON format? That would make it easily parseable. fpcalc provides an option for generating JSON output, and it makes it easier for Picard to handle the result.

But just a note here, parsing the text output is straight forward enough in this case.

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+
+SUPPORTED_EXTS = (
+    ".mp2",
+    ".mp3",
+    ".flac",
+    ".ogg",
+    ".oga",
+    ".spx",
+    ".opus",
+    ".m4a",
```
phw, line 74: I find it a bit odd other MP4 extensions (such as .mp4) are not in the list. Is this a fixed limitation currently of rsgain and would it reject other extensions? If yes, could this be extended in rsgain?

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+            raise Exception("ReplayGain 2.0: Failed to parse result")
+        results.append(result)
+
+    # Update track metadata with results
+    for i, track in enumerate(valid_tracks):
+        for tag in TAGS:
+            track.metadata.delete(tag)
+
+        # Opus R128 tags
+        if track.files[0].EXTENSIONS[0] == ".opus" and tagger.config.setting["opus_mode"] == OPUS_MODE_R128:
```
phw: It could check here if the file is an instance of `picard.formats.OggOpusFile`. That would make it less dependent on the hard coded file extension.

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+        else:
+            self.current += 1
+            return f" ({self.current}/{self.num_albums})"
+
+    def _albumgain_callback(self, album, result=None, error=None):
+        progress = self._format_progress()
+        if error is None:
+            for track in album.tracks:
+                for file in track.files:
+                    track.update_file_metadata(file)
```
phw: I think this is not the correct method to call. `Track.update_file_metadata`  is very specific and was created for the use case of needing to re-run the entire scripting on the file. I think for triggering the update of the metadata display `file.update()`  should be sufficient. Eventually you need to call `track.update()` after looping through the files as well.

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+        self.config.setting["album_tags"] = self.ui.album_tags.isChecked()
+        self.config.setting["true_peak"] = self.ui.true_peak.isChecked()
+        self.config.setting["target_loudness"] = self.ui.target_loudness.value()
+        self.config.setting["clip_mode"] = self.ui.clip_mode.currentIndex()
+        self.config.setting["max_peak"] = self.ui.max_peak.value()
+        self.config.setting["opus_mode"] = self.ui.opus_mode.currentIndex()
+        self.config.setting["opus_m23"] = self.ui.opus_m23.isChecked()
+
+register_track_action(ScanTracks())
+register_album_action(ScanAlbums())
```
phw, line 385: The old plugin also registered `register_file_action` , which would allow the plugin to run on single files even if they are not linked to a track, e.g. untagged files on the left. What do you think, does this make sense to add here?

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+
+def calculate_replaygain(tracks, options, tagger):
+
+    # Make sure files are of supported type, build file list
+    files = list()
+    valid_tracks = list()
+    for track in tracks:
+        if not track.files:
+            continue
+        file = track.files[0]
```
phw, line 174: There can be potentially multiple files per track (e.g. if the user has the file in both FLAC and MP3 format and tags both in one go). Would it break something if both files would be handled here?

phw (changes requested):
This is seriously awesome. Thank you so much for creating this plugin. Naming this replaygain2 is a good choice. And I like the options page, this is a sensible and useful set of options. But I'd suggest to add a link or button "Get rsgain" opening https://github.com/complexlogic/rsgain in a browser. That way the users get some guidance to obtaining rsgain if they don't have it.

Apart from that I have only a few comments on the code for discussion, and Codacy has some minor complaints.

> Should we add a deprecation notice to the description of the current ReplayGain plugin?

Yes

phw:
Two more things I noticed during testing:

1. Would you consider writing `replaygain_reference_loudness` as well? Maybe make it an option to write this tag. This is something the author of loudgain came up with, and I think it can be useful to remember the reference loudness, especially when it was set to something else then the RG2 default. Maybe this again would be something worse considering in rsgain itself, but it can also be done in the plugin.

2. While using the plugin I got this exception:

```
E: 23:12:21,695 /home/phw/devel/musicbrainz/picard/picard/util/thread.future_callback:59: Traceback (most recent call last):
  File "/home/phw/devel/musicbrainz/picard/./picard/util/thread.py", line 55, in future_callback
    result = future.result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/phw/.config/MusicBrainz/Picard/plugins/replaygain2/__init__.py", line 121, in calculate_replaygain
    if not track.files:
AttributeError: 'Album' object has no attribute 'files'
```

I think it happened when I had selected both an album and tracks. You need to filter the `objs`  passed to the plugin callback functions by proper type, see e.g. what the old plugin is doing at https://github.com/metabrainz/picard-plugins/blob/b3e37ce846ca1bdf1a7160e5d27b90de9edee4f8/plugins/replaygain/__init__.py#L144-L147

This also shows some special handling for the "standalone recrdongs" aka "non-album tracks" special album entry. It does not make sense to calculate album gain for this, because it is just a list of random tracks not attached to an album. 

As a side note: rsgain easy mode is great! That really makes things easy, and really should work for most. I just had it run over some 4k files, and it worked flawlessly. Looks like I found my new favorite replaygain utility :)

plugins/replaygain2/__init__.py
complexlogic: It would bias the album gain if one track has multiple files, but not others. That's why I calculate the gain for only the first file, and assume all others are the same

plugins/replaygain2/__init__.py
complexlogic: Most people scan ReplayGain on a per-album basis, so I'm not sure how useful the file mode is for this plugin. For example, if you have multiple unmatched albums of files on the left, you typically wouldn't want to scan that all at once, you'd want the files to be placed in an album first.

The `calculate_replaygain()` function in the plugin currently is `Track`-based, so it would require some refactoring to get it to work with `File` instead, but I could do it if it's really necessary.

plugins/replaygain2/__init__.py
complexlogic: I tried this, but it didn't work. Perhaps it's because the plugin writes to the `Track` metadata object instead of the `File` metadata object? My understanding is that the `Track.update_file_metadata()` copies the `Track.metadata` to the `File.metadata` for each file, and the `File.metadata` is what actually gets written when you click save. Is that correct?

plugins/replaygain2/__init__.py
complexlogic: Yes, rsgain will reject the other MP4 extensions. Yes, I am willing to extend it. The main reason I limited rsgain to .m4a only was because I was unsure if I could properly test all of the edge cases from the other extensions since they are less common. Let me know how you would like to proceed here.

plugins/replaygain2/__init__.py
complexlogic: That is an interesting idea. However, I think the existing parsing is simple enough and I would really prefer not to pull in another dependency for rsgain

plugins/replaygain2/__init__.py
complexlogic: I consider this a fork since I used your plugin as a base, so I'm going to leave both you and Sophist as authors

complexlogic:
> Would you consider writing replaygain_reference_loudness as well? Maybe make it an option to write this tag. This is something the author of loudgain came up with, and I think it can be useful to remember the reference loudness, especially when it was set to something else then the RG2 default. Maybe this again would be something worse considering in rsgain itself, but it can also be done in the plugin.

The reference loudness tags aren't part of the ReplayGain standard, and they aren't read by players, so I declined to support them in rsgain. But I can support them in this plugin instead, it's simply the target loudness setting formatted. I just added a checkbox in the plugin settings for it, disabled by default. The formatting is the same used by bs1770gain and loudgain, e.g. "-18.00 LUFS". Let me know if this works for you.

>While using the plugin I got this exception:

Fixed.

> It does not make sense to calculate album gain for this, because it is just a list of random tracks not attached to an album.

Perhaps we can introduce a new setting such as "Use track values for album tags on non-album tracks"? I think we should always have album tags if the album tags setting is checked, and this could be a good way of doing it.

plugins/replaygain2/__init__.py
phw: Fair enough :)

plugins/replaygain2/__init__.py
phw: Yes, you are right, the `file.metadata`  would need to be updated as well separately. That's how some other plugins did it, but others are using `Track.update_file_metadata()`. 

Let's keep it that way. It also has the advantage that it cleverly applies scripts again, so users can actually use the replaygain tags in their scripts if they want to.

plugins/replaygain2/__init__.py
phw: The edge cases is a good point. That's much easier for Picard as it only handles the metadata part, but I see how it makes sense for `rsgain`  to limit this to `.m4a`. E.g. `m4p` is with DRM and can't be decoded anyway. I think I just stumbled across this because one of my test files had the generic `.mp4` extension, but for most users `.m4a`  should be what they deal with, and other cases can be explained.

plugins/replaygain2/__init__.py
phw: Ok for me. Let's keep it the way it is and see for user feedback. If there is demand a file mode could be added.

I also agree on the album mode. For untagged files this could be realized by also allow the action for clusters.

plugins/replaygain2/__init__.py
phw: Now I see it, that makes sense. That's something I think the old plugin did not handle well, it just took all the files attached to an album.

phw:
> I just added a checkbox in the plugin settings for it, disabled by default. The formatting is the same used by bs1770gain and loudgain, e.g. "-18.00 LUFS". Let me know if this works for you.

That's perfect, thanks.

> Perhaps we can introduce a new setting such as "Use track values for album tags on non-album tracks"? I think we should always have album tags if the album tags setting is checked, and this could be a good way of doing it.

Agreed on always setting the album tracks. Not sure about the setting. IMHO it's not needed, it should always set the per-track values for NAT. But if there is a usecase for people wanting to tag all NAT tracks as if they were one album I'm not opposed to having such an option, it just should then be enabled by default.

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,345 @@
...
+                track.metadata.set("replaygain_album_peak", album_result["peak"])
+
+
+class ScanTracks(BaseAction):
+    NAME = N_("Calculate Replay&Gain...")
+
+    def callback(self, objs):
+        if not rsgain_found(self.tagger.config.setting["rsgain_command"], self.tagger.window):
+            return
+        tracks = list(filter(lambda o: isinstance(o, Track), objs))
```
phw, line 250: Should tracks here get grouped by album and then call `calculate_replaygain` for each group of album tracks instead?

plugins/replaygain2/__init__.py
Sophist-UK: Actually, I believe that the code of this is so new that I don't consider myself to be an author, and that you should get the author credit.

I do think that Phillip and I have contributed to the thinking behind this, and personally I feel a credit for previous versions and input to this one might be better served by a comment at the end of the description, with you taking sole author credit for this new plugin (and of course for the rsgain utility!!).

plugins/replaygain2/__init__.py
complexlogic: Picard's right window already groups tracks by album. I think it would be very unusual for a user to have multiple tracks selected across different albums simultaneously.

complexlogic:
> Agreed on always setting the album tracks. Not sure about the setting. IMHO it's not needed, it should always set the per-track values for NAT. But if there is a usecase for people wanting to tag all NAT tracks as if they were one album I'm not opposed to having such an option, it just should then be enabled by default.

Ok, that makes sense. I changed it permanently without setting.

complexlogic:
@phw This is ready for final review for merging. I added a description, and put a deprecation notice on the old plugin. 

There were some minor formatting inconsistencies in rsgain that I had to work around in the plugin (e.g. having an extra `\n` for album result but not track result). I decided to fix it in rsgain to make the code cleaner in the plugin. The new version of rsgain is 3.0.1, and you will need to upgrade to be compatible with the latest plugin.

Please let me know if you have any other feedback.

plugins/replaygain/__init__.py
```diff
@@ -2,6 +2,7 @@
 PLUGIN_NAME = "ReplayGain"
 PLUGIN_AUTHOR = "Philipp Wolfer, Sophist"
 PLUGIN_DESCRIPTION = """<h2>Calculate ReplayGain for selected files and albums.</h2>
+<i>This plugin has been deprecated in favor of the ReplayGain 2.0 plugin</i>
```
phw, line 5: There is no functional change, but as the content has been changed I bumped the version of the plugin to 0.3.2

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+
+This plugin is based on the original ReplayGain plugin by Philipp Wolfer and Sophist.
+'''
+PLUGIN_VERSION = "1.0"
+PLUGIN_API_VERSIONS = ["2.0"]
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
+
+from functools import partial
+import subprocess
```
phw: I think we understand the risks of subprocess here but we need to run this external command. I made codacy ignore this warning by adding ` # nosec: B404`  to this line.

phw (approved):
This is working great for me, thanks a lot.

I took the freedom to bump the version of the old replaygain plugin and also made codacy ignore some warnings. We can squash this when merging.

regarding codacy I'm not quite sure why it is complaining about the ` N_`  function calls. It's supposed to know about this as an built-in due to the pylint config in [.pylintrc](https://github.com/metabrainz/picard-plugins/blob/2.0/.pylintrc). But that's a different thing to investigate, does not prevent us to merge this PR.

I'll give this some more testing locally and the merge it.

phw (changes requested):
Actually I found an issue: If you calculate the replay gain the calculated tags get added. If you now remove those tags in the UI and use "Calculate ReplayGain" again it does not re-add those tags.

The reason is that when your remove those tags they get marked as deleted in the metadata. and the default behavior of `File.copy_metadata` , as called in https://github.com/metabrainz/picard/blob/master/picard/track.py#L187 , is to keep all the deleted tags on the file.

This is in general intentional. The idea is that if a user marked a tag for deletion on a file it should not undo this intentional deletion when the file got matched to a release.

But it's not wanted here in this case. I guess the solution would be to not use `Track.update_file_metadata`  but instead to explicitly set the tag changes on the file's metadata.

I'd suggest factoring out the code that sets the tags to the track metadata in https://github.com/metabrainz/picard-plugins/pull/336/files#diff-e0c03c9ec28ab3972fefedd02d351293936964857e77ea1eb1a530bde5ad5e36R208-R229 into a separate function, accepting a metadata object to update and a boolean flag indicating whether Opus mode should be used.

complexlogic:
@phw Good find. I refactored the tagging code into a function that accepts a `Metadata` object, plus a few other parameters. I pass the `File.metadata` into into the new function for each `File` in `Track` and later call `File.update`, so `Track.update_file_metadata` is no longer used. I tested deleting the tags and re-scanning and it's working now.

Let me know what you think of the updated code, and if you find any other issues.

plugins/replaygain2/__init__.py
```diff
@@ -0,0 +1,383 @@
...
+
+    # Update track metadata with results
+    for i, track in enumerate(valid_tracks):
+        for file in track.files:
+            update_metadata(
+                file.metadata,
+                results[i],
+                album_result,
+                isinstance(track, NonAlbumTrack),
+                True if isinstance(file, OggOpusFile) and config.setting["opus_mode"] == OPUS_MODE_R128 else False,
```
phw: Maybe set a variable `opus_mode = config.setting["opus_mode"] == OPUS_MODE_R128`  before the outer for loop and then here just check for:

```python
opus_mode and isinstance(file, OggOpusFile),
```

phw (commented):
Looks and works good, thanks a lot. Minor suggestion for code improvement. 

Also pushed a commit myself that replaces access to `tagger.config`  and ` tagger.log` to use the imports directly. That's the current preferred way, the old replaygain plugin set a bad example here :)

plugins/replaygain2/__init__.py
complexlogic: Done.

phw (approved):
Ok, let's merge this. I'll squash and merge this shortly

phw:
Merged, thanks a lot again. The new plugin will show up for download from within Picard sometime in the next 12 hours. @complexlogic would you like to make a post on https://community.metabrainz.org/ to introduce this new plugin and rsgain? I think this would be the best way to get people know about this new plugin.

On a side note I also updated the linter configuration needed for codacy to follow the Picard settings, so the code no longer fails the codacy check.

complexlogic:
@phw Yes, I will make a post on the forum.

To clean up this repo: I recommend closing #89 and #121.

#209 was fixed by Sophist in #290 so that one should definitely be closed.

PR #338, opened 2022-10-16
Remove obsolete string  calls

phw:
The global `string_`  function was deprecated in Picard a long time ago. As far as I can tell from other uses it is no longer required at the places were it was in use. Completely remove it from all plugins.


zas (approved):
Good move, LGTM

PR #339, opened 2022-10-23
add critiquebrainz review comments

tobiasSarner:
add review comments for release-group and recording using critiquebrainz

PR #340, opened 2022-10-26
Patch 2 relations

tobiasSarner:
add website tags to store references from release, release-group, artist


phw:
Thanks, but now I'm confused :) The Critiquebrainz plugin is already part of #339 . Can you remove this from the pull request and if there are any changes you'd like to inlcude add those to the original pull request? Then we can better separate them.

Not so sure about the relations plugin. Maybe this should be better called `website` or `artist_websites`? Also I think the official homepage should be set to the `website` tag. Then it would end up in the proper format specific tags, such as `WOAR` in ID3, see https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#website-official-artist-website

tobiasSarner:
yes, i remove this one and will a new one base on main branch

PR #341, opened 2022-11-03
add plugin to set website from relations

tobiasSarner:
this do not attache existing artist website plugin and can used separate to it
it is add currently the url from album-artist, release and release-group

### DIFF
--- a/plugins/relations_website/relations_website.py
+++ b/plugins/relations_website/relations_website.py
@@ -29,7 +29,7 @@ Example: Taylor Swift
  website:release-group_1_https://www.discogs.com/master/2831825'''
 PLUGIN_LICENSE = "GPL-2.0"
 PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
-PLUGIN_VERSION = "1.0.0"
+PLUGIN_VERSION = "1.0.1"
 PLUGIN_API_VERSIONS = ["2.0"]
 
 from functools import partial
@@ -38,8 +38,8 @@ from picard.metadata import register_album_metadata_processor
 from picard.webservice import ratecontrol
 from picard.util import load_json
 
-MUSICBRAINZ_HOST = "musicbrainz.org"
-MUSICBRAINZ_PORT = 80
+MUSICBRAINZ_HOST = config.setting["server_host"]
+MUSICBRAINZ_PORT = config.setting["server_port"]
 
 ratecontrol.set_minimum_delay((MUSICBRAINZ_HOST, MUSICBRAINZ_PORT), 50)
 
@@ -59,7 +59,7 @@ def result_albumartist(album, metadata, data, reply, error):
             for relations in data:
                 if "url" in relations and "resource" in relations["url"]:
                     idx += 1
-                    ident = "website:" + prefix + "_" + str(idx) + "_" + relations["type"]
+                    ident = "website:" + prefix + "_" + str(idx) + "_" + relations["type"].lower()
                     if ident in metadata:
                         metadata[ident] = review_text
                     else:
@@ -71,35 +71,6 @@ def result_albumartist(album, metadata, data, reply, error):
     finally:
         album._requests -= 1
         album._finalize_loading(None)
-
-
-def result_release(album, metadata, data, reply, error):
-    prefix = "release"
-    musicbrainz_id = metadata["musicbrainz_albumid"]
-    if error:
-        album._requests -= 1
-        album._finalize_loading(None)
-        return
-
-    try:
-        data = load_json(data).get("relations")
-        if data:
-            idx = 0
-            for relations in data:
-                if "url" in relations and "resource" in relations["url"]:
-                    idx += 1
-                    key = "website:" + prefix + "_" + str(idx) + "_" + relations["type"]
-                    if key in metadata:
-                        metadata.delete(key)
-                    metadata.add(key, relations["url"]["resource"])
-
-            log.debug("%s: %s %s (%s) Parsed response", PLUGIN_NAME, prefix, musicbrainz_id, metadata["albumtitle"])
-    except Exception as e:
-        log.error("%s: %s %s (%s) Error parsing response: %s", PLUGIN_NAME, prefix, musicbrainz_id, metadata["albumtitle"], str(e))
-    finally:
-        album._requests -= 1
-        album._finalize_loading(None)
-
 
 def result_releasegroup(album, metadata, data, reply, error):
     prefix = "release-group"
@@ -116,10 +87,11 @@ def result_releasegroup(album, metadata, data, reply, error):
             for relations in data:
                 if "url" in relations and "resource" in relations["url"]:
                     idx += 1
-                    key = "website:" + prefix + "_" + str(idx) + "_" + relations["type"]
-                    if key in metadata:
-                        metadata.delete(key)
-                    metadata.add(key, relations["url"]["resource"])
+                    ident = "website:" + prefix + "_" + str(idx) + "_" + relations["type"].lower()
+                    if ident in metadata:
+                        metadata[ident] = review_text
+                    else:
+                        metadata.add(ident, relations["url"]["resource"])
 
             log.debug("%s: %s %s (%s) Parsed response", PLUGIN_NAME, prefix, musicbrainz_id, metadata["albumtitle"])
     except Exception as e:
@@ -128,15 +100,14 @@ def result_releasegroup(album, metadata, data, reply, error):
         album._requests -= 1
         album._finalize_loading(None)
 
-
 def process_relations(album, metadata, release):
     queryargs = {
         "fmt": "json",
         "inc": "url-rels"
     }
     album.tagger.webservice.get(
-        config.setting["server_host"],
-        config.setting["server_port"],
+        MUSICBRAINZ_HOST,
+        MUSICBRAINZ_PORT,
         "/ws/2/artist/" + metadata["musicbrainz_albumartistid"],
         partial(result_albumartist, album, metadata),
         priority=True,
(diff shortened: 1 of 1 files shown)


plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+PLUGIN_NAME = 'Musicbrainz Relations Website'
+PLUGIN_AUTHOR = 'Tobias Sarner'
+PLUGIN_DESCRIPTION = '''Uses Musicbrainz relations to set website advanced.
+ -Album Artist Website- plugin, can used separate to set album artist(s) Official Homepage(s) and will not attached by this plugin
```
Sophist-UK, line 22: I am entirely unclear from this description what the relationship of this plugin is to my Album Artist Website plugin.  But I wonder whether this functionality would be better served by adding it to the Album Artist Website plugin (renaming it Websites Metadata perhaps) rather than creating a new one.

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+
+WARNING: Experimental plugin. All guarantees voided by use. 
+
+Example: Taylor Swift
+ website:artist_1_official homepage https://www.taylorswift.com/
+ website:release_1_purchase https://taylorswift.universal-music.de/p51-i0602445790098/taylor-swift/midnights/index.html
+ website:release-group_1_https://www.discogs.com/master/2831825'''
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
+PLUGIN_VERSION = "1.0.0"
```
Sophist-UK: If this is experimental, the version number should be 0.0.1.

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+
+PLUGIN_NAME = 'Musicbrainz Relations Website'
+PLUGIN_AUTHOR = 'Tobias Sarner'
+PLUGIN_DESCRIPTION = '''Uses Musicbrainz relations to set website advanced.
+ -Album Artist Website- plugin, can used separate to set album artist(s) Official Homepage(s) and will not attached by this plugin
+
+WARNING: Experimental plugin. All guarantees voided by use. 
+
+Example: Taylor Swift
+ website:artist_1_official homepage https://www.taylorswift.com/
```
Sophist-UK, line 27: 1. I am not sure that you can just suffix the existing `website` tag with `:subtag` and not screw up the existing usage of `website` without a suffix. From memory (which is old and may be wrong) Picard knows which tags are multi-value, and AFAIK `website` is not one of them.
2. `website` is explicitly mapped to e.g. the ID3v2 WOAR tag which is explicitly for the Album Artist official webpage. The WOAS tag is defined in ID3v2 as holding the official web page for the Release, and Picard needs a definition of a metadata name for this and a mapping to ID3 and other tagging formats to defined how it is loaded and saved.
3. Similarly there is a WXXX:subtag tag in ID3v2 to hold miscellaneous URLs that do not have explicit tags - so e.g. the Discord URL should be saved in WXXX:discord, and again a mapping from metadata to WXXX would be needed in Picard to support this. I don't have any issues with the idea of storing the Discord URL for use in players, but if we are going to support Discord we should probably also have a Discord-Release metadata variable which holds the discord release number and Discord-Release Group to store the Discord master release number. Discord URLs can then be generated from these if you double click on them in Picard.

So, what I am saying is that ideally there would need to be a Picard enhancement to support the metadata you want to hold and make sure it gets saved in the write tags for the various formats. If someone is going to do this, then a comprehensive review of ID3 tags and other formats and a full mapping of these to metadata names would be a great enhancement IMO.

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+from functools import partial
+from picard import config, log
+from picard.metadata import register_album_metadata_processor
+from picard.webservice import ratecontrol
+from picard.util import load_json
+
+MUSICBRAINZ_HOST = "musicbrainz.org"
+MUSICBRAINZ_PORT = 80
+
+ratecontrol.set_minimum_delay((MUSICBRAINZ_HOST, MUSICBRAINZ_PORT), 50)
```
Sophist-UK, line 44: IMO you absolutely should **not** override the ratecontrol in a plugin like this.

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+        album._requests -= 1
+        album._finalize_loading(None)
+
+
+def process_relations(album, metadata, release):
+    queryargs = {
+        "fmt": "json",
+        "inc": "url-rels"
+    }
+    album.tagger.webservice.get(
```
Sophist-UK, line 108: When I wrote the Artist website plugin, the artist official website was not available in the standard release metadata (even with the extra relations enabled) so I wrote the plugin to make an additional webservice call. BUT, since it was likely that a user might be tagging multiple albums for the same artist, I spent a lot of time crafting a caching mechanism, so that you only made 1 call per artist for the lifetime of the Picard execution process. If you are going to duplicate the functionality to make such a webservice call, you probably ought to reuse the caching code I wrote.

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+        config.setting["server_host"],
+        config.setting["server_port"],
+        "/ws/2/artist/" + metadata["musicbrainz_albumartistid"],
+        partial(result_albumartist, album, metadata),
+        priority=True,
+        parse_response_type=None,
+        queryargs=queryargs
+    )
+    album._requests += 1
+    album.tagger.webservice.get(
```
Sophist-UK, line 147: Picard has already downloaded this information - why are you repeating the call?

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+        config.setting["server_host"],
+        config.setting["server_port"],
+        "/ws/2/release/" + metadata["musicbrainz_albumid"],
+        partial(result_release, album, metadata),
+        priority=True,
+        parse_response_type=None,
+        queryargs=queryargs
+    )
+    album._requests += 1
```
Sophist-UK, line 156: Picard has already downloaded this information - why are you repeating the call?

Sophist-UK (changes requested):

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+        partial(result_albumartist, album, metadata),
+        priority=True,
+        parse_response_type=None,
+        queryargs=queryargs
+    )
+    album._requests += 1
+    album.tagger.webservice.get(
+        config.setting["server_host"],
+        config.setting["server_port"],
+        "/ws/2/release/" + metadata["musicbrainz_albumid"],
```
Sophist-UK: Picard has already downloaded this information - why are you repeating the call?

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,169 @@
...
+        partial(result_release, album, metadata),
+        priority=True,
+        parse_response_type=None,
+        queryargs=queryargs
+    )
+    album._requests += 1
+    album.tagger.webservice.get(
+        config.setting["server_host"],
+        config.setting["server_port"],
+        "/ws/2/release-group/" + metadata["musicbrainz_releasegroupid"],
```
Sophist-UK, line 121: Picard has also already downloaded this information - why are you repeating the call?

Sophist-UK (changes requested):

Sophist-UK:
P.S. I encourage you to continue to work on this plugin - community contributions are always welcome IMO. But writing plugins does have a learning curve - sorry.

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,140 @@
...
+
+WARNING: Experimental plugin. All guarantees voided by use. 
+
+Example: Taylor Swift
+ website:artist_1_official homepage https://www.taylorswift.com/
+ website:release_1_purchase https://taylorswift.universal-music.de/p51-i0602445790098/taylor-swift/midnights/index.html
+ website:release-group_1_https://www.discogs.com/master/2831825'''
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.txt"
+PLUGIN_VERSION = "1.0.1"
```
Sophist-UK: You don't need to increment the version except from one **published** version to another. This plugin has not yet been published, so you don't need to increment the version. But as previously stated, since this experimental it should probably be version 0.0.1 at the moment.

plugins/relations_website/relations_website.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
+# Critiquebrainz plugin for Picard
```
Sophist-UK: Minor point, but this is NOT the Critiquebrainz plugin.

Sophist-UK (changes requested):

tobiasSarner:
To that 3 points of ID3 tags.
I am working with vorbis flags on accurat encoded flac files, this compatiblity of website tag in IDv3 I will have a look to you plugin and will test it.
It should by possible to set, but the interpretation of this tag-type only on artist website is low and would be nice do differ that.
It could by like cover images by creation of tag by type and text/link.
I would like to see that information on playing track by do not searching it again in browser.

Sophist-UK:
1. It doesn't matter what format you are talking about, if you want to load / save metadata in a way that it is useful e.g. to a player app, you need to ensure that the data will be saved in the correct way into the tags. Some formats (like ID3) get unknown metadata names mapped to e.g. a TXXX field, other formats just don't save it.

2. Picard does not have any functionality designed to allow plugins to add new mappings. You might be able to monkey-patch it in (which might make the plugin much more specific to a Picard Release and more ,likely to break due to a code change in Picard, but it would IMO be easier (and by sticking to the API lower risk) to avoid monkey-patching and add a more comprehensive set of mappings to the next release of Picard.

plugins/relations_website/relations_website.py
phw: Without further knowledge Picard would just save a `TXXX:website:subtag`, so that would not be a problem in general. Even though some tags use colons for special behavior (e.g. `comment:description`), there is nothing special per se for this notation.

But anyway, I totally agree with the points you make above. It would be better if support for `WXXX` (and in general more of the `W***` webpage tags) would exist. As a first step having `website:subtag` mapped to `WXXX:subtag` sounds sensible. But we could also directly support let's say `website:pay` for `WPAY` (PICARD-1394) or so.

phw: I added PICARD-2587 as a start. Not yet sure if my suggestion is the best idea.

PR #349, opened 2023-04-05
Plugin addition - Submit Folksonomy Tags

ghost:
You may know me as Victini on MusicBrainz or Flaky on Discord.

Put briefly, this plugin lets users submit tags from their files as MusicBrainz' folksonomy tags. I've gone into more detail here: https://community.metabrainz.org/t/submit-folksonomy-tags-plugin-for-submitting-tags-to-musicbrainz-via-picard/626105

The only major bug I have with the plugin is that submitting after logging in from Picard's "to do this you must login" popup gives me an Unauthorised error (any subsequent submits are fine and of course, logging on before submitting), but Aerozol suggested on Discord that I should submit this anyhow and fix it later.

### DIFF
--- a/plugins/submit_folksonomy_tags/__init__.py
+++ b/plugins/submit_folksonomy_tags/__init__.py
@@ -5,7 +5,7 @@ A plugin allowing submission of specific tags on tracks you own (defaults to <i>
 
 A MusicBrainz login is required to use this plugin. Log in first by going to the General options. Then, to use, right click on a track or release then go to <i>Plugins</i> and depending on what you want to submit, choose the option you want.
 
-Uses code from rswift's "Submit ISRC" plugin (specifically, the handling of the network response)
+Uses code from rdswift's "Submit ISRC" plugin (specifically, the handling of the network response)
 """
 PLUGIN_VERSION = '0.2.4'
 PLUGIN_API_VERSIONS = ['2.2']
@@ -28,7 +28,7 @@ from PyQt5 import QtCore
 from PyQt5.QtWidgets import QMessageBox
 
 # List of Qt network error codes.
-# From "Submit ISRC" plugin - credit to rswift.
+# From "Submit ISRC" plugin - credit to rdswift.
 q_error_codes = {
     0: 'No error',
     1: "The remote server refused the connection (the server is not accepting requests).",
@@ -86,11 +86,11 @@ def tag_submit_handler(document, reply, error, tagger):
     The function handling the network response from MusicBrainz
     or QtNetwork, showing a message box if an error had occurred.
 
-    Uses the network response handler code from rswift's "Submit ISRC"
+    Uses the network response handler code from rdswift's "Submit ISRC"
     plugin.
     """
     if error:
-        # Error handling from rswift's Submit ISRC plugin
+        # Error handling from rdswift's Submit ISRC plugin
         xml_text = str(document, 'UTF-8') if isinstance(document, (bytes, bytearray, QtCore.QByteArray)) else str(document)
 
         # Build error text message from returned xml payload
--- a/plugins/submit_folksonomy_tags/__init__.py
+++ b/plugins/submit_folksonomy_tags/__init__.py
@@ -22,8 +22,8 @@ from picard.ui.options import OptionsPage, register_options_page
 from picard.webservice.api_helpers import MBAPIHelper, _wrap_xml_metadata
 from .ui_config import TagSubmitPluginOptionsUI
 import re
-import html
 import functools
+from xml.sax.saxutils import escape
 from PyQt5 import QtCore
 from PyQt5.QtWidgets import QMessageBox
 
@@ -197,8 +197,7 @@ def handle_submit_process(tagger, track_list, target_tag):
     for track in track_list:
         if track.files:
             for file in track.files:
-                mbid_list = [html.escape(tag.strip().lower()) for tag
-                             in re.split(";|/|,", file.metadata[target_tag])]
+                mbid_list = file.metadata.getall(target_tag)
                 if len(mbid_list) > 1:
                     alert_multiple_mbids = True
                 for mbid in mbid_list:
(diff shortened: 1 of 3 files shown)


plugins/submit_folksonomy_tags/__init__.py
```diff
@@ -0,0 +1,440 @@
...
+                        else:
+                            data[dict_key][mbid] = processed_tags
+                    else:
+                        log.info(f"Not submitting MBID {track.metadata[target_tag]} as it was found on 'do not submit' MBID set.")
+    
+    # Send an alert when, at the end of it all, inconsistent tagging was detected.
+    if inconsistent_detected or alert_multiple_mbids:
+        warning = QMessageBox()
+        warning.setStandardButtons(QMessageBox.Ok|QMessageBox.Cancel)
+        warning.setDefaultButton(QMessageBox.Cancel)
```
rdswift, line 232: Since you use the same icon in all cases, you might consider setting it here rather than in each conditional block.

ghost: Alright, thanks!

ghost:
Only major issue is the complexity now. I should be able to work on that next week, if that's alright?

rdswift:
> Only major issue is the complexity now. I should be able to work on that next week, if that's alright?

That should be fine.  I don't think there's any need to rush.  The current working version is available from your repository if anybody wants to try it out, so I suggest that you take whatever time you need.  For what it's worth, I still have a tendency to include everything in the same function / module / method myself.  I think it's because I've done way too much "linear" programming in the past, and haven't totally embraced the concepts of OOP.  :grin: 

One final thing, and this is really minor...  I appreciate the credit that you've provided in the comments in your code, but I noticed that you referred to 'rswift' rather than 'rdswift'.  Like I said, really minor.  :wink:

ghost:
Alright, thank you! :)

I've tested the plugin on Arch Linux (Python 3.10.10) / Picard 2.8.5 and all seems to be working just fine now. The complexity of the main function should still be sorted in the near future, but if there's no other problems I think it can be merged.

plugins/submit_folksonomy_tags/__init__.py
```diff
@@ -0,0 +1,438 @@
...
+        "a0ef7e1d-44ff-4039-9435-7d5fefdeecc9", # artist: [theatre]
+        "80a8851f-444c-4539-892b-ad2a49292aa9", # artist: [language instruction]
+        "",                                     # blank
+    }
+
+    for track in track_list:
+        if track.files:
+            for file in track.files:
+                mbid_list = [html.escape(tag.strip().lower()) for tag
+                             in re.split(";|/|,", file.metadata[target_tag])]
```
phw: Instead of the splitting `file.metadata.getall(target_tag)`  should be used to obtain the list of individual values.

plugins/submit_folksonomy_tags/__init__.py
```diff
@@ -0,0 +1,438 @@
...
+    dict_key = ""
+    tags_to_search = config.setting['tag_submit_plugin_tags_to_submit']
+
+    # Variable to enable when inconsistent tagging is detected, which can be problematic for anything other than recordings.
+    alert_inconsistent = True
+    inconsistent_detected = False
+    # Variable to enable alert if multiple MBIDs are associated, must be toggled.
+    alert_multiple_mbids = False
+
+    # TODO when Windows Picard updates with Python 3.10, use case/switch.
```
phw, line 167: Just a side note: It will take quite a while until we consider 3.10 the minimum supported version. Essentially this will happen when either 3.9 is no longer supported upstream or when we have some really urgent need to use something not available in earlier Python versions.

plugins/submit_folksonomy_tags/__init__.py
```diff
@@ -0,0 +1,438 @@
...
+        # start the list
+        xml_data.append(f"<{key}-list>")
+        for mbid in data[key]:
+            # deduplicate the list of genres so no redundant tags are sent.
+            data[key][mbid] = list(set(data[key][mbid]))
+            # start the user tag list
+            xml_data.extend([f'<{key} id="{mbid}">', "<user-tag-list>"])
+            # add the tags
+            for tag in data[key][mbid]:
+                xml_data.append(f'<user-tag{upvote_tag_fill}><name>{process_tag_aliases(tag.lower())}</name></user-tag>')
```
phw: Are the tags already XML escaped at this point?

phw (changes requested):
Thanks a lot for submitting this. Overall looks good to me and I had already tested your plugin a while back. Only one suggestion to use `metadata.getall`  instead of splitting tag values.

> The complexity of the main function should still be sorted in the near future, but if there's no other problems I think it can be merged.

Agreed. Factoring out some code into separate functions should do the trick, but I'm fine with merging this as is for now.

plugins/submit_folksonomy_tags/__init__.py
ghost: I don't think so, but I can get that sorted. Shouldn't be a problem.

phw: From reading the source it is probably fine because of https://github.com/metabrainz/picard-plugins/pull/349/files#diff-9cccf3c2c55a90e5485eb300d6ad5dba86a9cb517fbd46f45f2573dce5ac62b3R200 . But please double check.

Maybe it is even better to move the escaping to here where the XML is actually build. That makes it less likely this breaks somewhere in between after some refactoring.

ghost: Before I go and change it, should I keep to the current library or move to `xml.sax.saxutils` instead?

And yeah, once the plugin's in, I'm def gonna refactor. :sweat_smile:

phw: General rule to follow is to stick to those modules Picard already uses, as only those are available in the binary builds for Windows and macOS.

But having said this `xml.sax.saxutils` should be fine and is probably the best choice here. Picard already uses this internally at https://github.com/metabrainz/picard/blob/master/picard/webservice/api_helpers.py

Just as a side note: There is not much XML functionality left in Picard as all the core services moved to JSON. The ratings submission is the only built-in place left where XML is actively used. Apart from that only https://github.com/metabrainz/picard/blob/master/picard/util/xml.py exists for parsing XML (based on Qt functionality), but it's only there for use by plugins that access XML web services.

phw (approved):

ghost:
I've started work on refactoring, and thus far I've managed to split the submission handler into different functions in a class. I'm using Radon as a way to check if I'm on the right track, will that pose an issue with Codacy?

FWIW, the initial functions I have have a complexity of 13 at the highest (Radon also says B-grade) right now.

PR #353, opened 2023-05-11
Replaygain type detection

phw:
The replaygain2 plugin does not detect Ogg Vorbis files as supported because it checks whether the first extension defined in `File.EXTENSIONS` is contained in a whitelist of supported file types.

But in Picard the `OggVorbisFile` file class as an empty `EXTENSIONS`, as ".ogg" is a generic extension used for various formats using the Ogg container. `EXTENSIONS` is meant for Picard's format detection logic.

Hence replace the extension based detection with a proper type check checking for the Picard file type being used.


zas (approved):
LGTM

PR #355, opened 2023-07-19
replaygain2: fix several issues with with call to window.set_statusbar_message

phw:
Calls to `window.set_statusbar_message` had several issues:

- `N_()` was used, even though this neither causes translation of strings nor are translations in plugins currently supported. `N_()` just prepares a string for extraction. Without the use of `_()` somewhere this has no effect.
- Use of f-strings should be avoided here, as the function expects percent formatting. Passing in variables via f-string can lead to errors if such variables contain percentage signs (PICARD-2683)
- f-strings must also never be used in combination with translations, as this would look for a translation string matching exactly the substituted string. Instead translations must always use some kind of template string (e.g. percentage formatting or  `String.format()`)

### DIFF
--- a/plugins/replaygain2/__init__.py
+++ b/plugins/replaygain2/__init__.py
@@ -29,7 +29,7 @@ The following file formats are supported:
 
 This plugin is based on the original ReplayGain plugin by Philipp Wolfer and Sophist.
 '''
-PLUGIN_VERSION = "1.2"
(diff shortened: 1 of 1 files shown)


phw:
@complexlogic FYI

complexlogic:
Thanks for the information, I'll keep that in mind for future plugin development

zas (approved):
LGTM

PR #356, opened 2023-07-28
Update Submit ISRC plugin to match the revised format of call to MBAPIHelper class in Picard

rdswift:
Changes in Picard 2.9 to the way the path is passed to the `MBAPIHelper` class `post` method (https://github.com/metabrainz/picard/pull/2225) broke the **Submit ISRC** plugin.  This PR revises the plugin to use the new format for the path argument to the method.

**Question:** Should the `PLUGIN_API_VERSIONS` list be reduced to only '2.9' because I'm not sure that it will work with prior versions of Picard?

### DIFF
--- a/plugins/submit_isrc/__init__.py
+++ b/plugins/submit_isrc/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020-2021 Bob Swift (rdswift)
+# Copyright (C) 2020-2021, 2023 Bob Swift (rdswift)
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


phw:
We missed that use, really thought we would not break any of the existing plugins.

> Should the PLUGIN_API_VERSIONS list be reduced to only '2.9' because I'm not sure that it will work with prior versions of Picard?

Yes, I would suggest to do that. The alternative would be to check the Picard version and use the old format for the path for older versions.

rdswift:
> We missed that use, really thought we would not break any of the existing plugins.

It looks like it also breaks the **Submit Folksonomy Tags** plugin.  I can open a PR for that one as well.

> > Should the PLUGIN_API_VERSIONS list be reduced to only '2.9' because I'm not sure that it will work with prior versions of Picard?
> 
> Yes, I would suggest to do that. The alternative would be to check the Picard version and use the old format for the path for older versions.

I'll maybe check the Picard version as you suggested.  That way it will still work on older installations.

zas (approved):

phw (approved):

PR #357, opened 2023-07-28
Update Submit Folksonomy Tags plugin to match the revised format of call to MBAPIHelper class in Picard

rdswift:
Changes in Picard 2.9 to the way the path is passed to the MBAPIHelper class post method (https://github.com/metabrainz/picard/pull/2225) broke the Submit Folksonomy Tags plugin by @Viktini. This PR revises the plugin to use the new format for the path argument to the method.

### DIFF
--- a/plugins/submit_folksonomy_tags/__init__.py
+++ b/plugins/submit_folksonomy_tags/__init__.py
@@ -1,3 +1,23 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 Flaky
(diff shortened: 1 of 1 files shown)


phw (approved):
Looks good. Thanks for taking care

zas (approved):

ghost:
Hey, thank you for sorting this! Good timing too, for multiple reasons (including illness) I was unable to get things done here for 2.9, hence my late reply.

PR #358, opened 2023-08-09
Submit Folksonomy Tags: Refactor code + clean up imports

ghost:
First off, thanks to @rdswift for updating SFT to work with version 2.9's API. I was recovering from illness when that PR was merged, so I wouldn't have been able to deal with it.

Now, a while back I said I would refactor the code, and this is essentially part of that. There might be some more refactoring required, but I think this should be good enough right now to make the code easier to deal with. I've also split the network function and submit handling off to separate files.

LMK if there are any problems!

### DIFF
--- a/plugins/submit_folksonomy_tags/submit_handler.py
+++ b/plugins/submit_folksonomy_tags/submit_handler.py
@@ -83,7 +83,7 @@ def process_tracks_to_file_list(track_list, target_tag):
     the MBID is in the banned list and if alert_multiple_mbids
     needs to be checked.
     """
-    global inconsistent_detected, alert_multiple_mbids
+    global alert_inconsistent, alert_multiple_mbids
 
     file_list = []
     for track in track_list:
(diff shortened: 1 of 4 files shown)


plugins/submit_folksonomy_tags/submit_handler.py
```diff
@@ -78,12 +78,13 @@ def handle_submit_process(tagger, track_list, target_tag, client_params):
 
 
 def process_tracks_to_file_list(track_list, target_tag):
-    global alert_inconsistent, inconsistent_detected, alert_multiple_mbids
     """
     Processes a track list to just a file list, checking if
     the MBID is in the banned list and if alert_multiple_mbids
     needs to be checked.
     """
+    global inconsistent_detected, alert_multiple_mbids
```
ghost: whoops, removed the wrong one. :neutral_face: fixed it though

ghost:
Closing as I'm not 100% satisfied with how it was done, I'll keep the old code for now.

PR #360, opened 2023-09-15
Deezerart: Use astrcmp and make matching threshold configurable

phw:
The deezerart plugin used Python's difflib for string similarity comparison. This has some problems:

1. difflib ist not part of the Picard binary packages for macOS and Windows, as the bundled Python only includes those modules actually used
2. The `SequenceMatcher.quick_ratio` only compares length and characters, not sequences (which means order of characters does not matter). See for example:

    ```
    >>> from difflib import SequenceMatcher
    >>> SequenceMatcher(None, "michael jackson", "joachim slacken").quick_ratio()
    1.0
    >>> from picard.util.astrcmp import astrcmp
    >>> astrcmp("michael jackson", "joachim slacken")
    0.3333333134651184
    ```

Use Picard's `astrcmp` function instead. A value of 0.6 seemed sensible in my tests. But a configuration option was added so users can tweak the threshold if they get bad results.


zas (approved):
LGTM

PR #361, opened 2023-10-21
Use configured 'join_genres' setting to split genres.

rdswift:
If the user had set the "**Join multiple genres with**" setting to something other than the default (blank), the plugin did not properly split the genres for processing.  See the discussion at https://community.metabrainz.org/t/proposed-genre-mapping-plugin/528398/14 for more information.

Solution is to use the "**Join multiple genres with**" setting if it has been set, otherwise fall back to the standard **MULTI_VALUED_JOINER** constant.


phw (approved):
Great fix

PR #362, opened 2023-10-24
Update acousticbrainz plugins to correctly use a 1s rate limit.

Sophist-UK:
Actual acousticbrainz rate limit is [10 requests every 10s](https://acousticbrainz.readthedocs.io/api.html#rate-limiting)


phw (approved):
Thanks. The history is interesting. Originally there was a 50ms delay. Got increased in both plugins, one to 100, one to 250ms.

I don't know the AB side, but maybe the rate limiting was increased after deprecation of the service?

Anyway, let's merge this

PR #363, opened 2023-10-24
Add new "Script Logger" plugin

rdswift:
This plugin provides a new script function `$logline()` to write entries to Picard's system log.  By default, the log level is set at `Info`, but any level can be used by providing the level as an optional second parameter to the function.

The function is used as:

`$logline(text[,level])`

where `text` is the text to write to the log.  The entry will be written at log level `Info` by default, but this can be changed by specifying a different level as an optional second parameter.  Allowable log levels are:

- E (Error)
- W (Warning)
- I (Info)
- D (Debug)

If an unknown level is entered, the function will use the default `Info` level.

### DIFF
--- a/plugins/script_logger/__init__.py
+++ b/plugins/script_logger/__init__.py
@@ -1,0 +1,96 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 Bob Swift (rdswift)
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


Sophist-UK:
Nice!!!!

phw (approved):
Code looks good. I guess the plugin also can be useful for debugging. It's clearly something advanced, so I'd be happy to keep this as a plugin only.

I wonder if it would make sense to write all the output lines with a fixed, common prefix to easily identify the log lines created by this plugin?

rdswift:
> I wonder if it would make sense to write all the output lines with a fixed, common prefix to easily identify the log lines created by this plugin?

I actually had this originally set up to preface each line with "Script Logger:", but when added to the plugin path that Picard includes automatically, it became a bit lengthy (and I couldn't come up with a way to prevent Picard from including the full path to the plugin).  It would actually be nice to preface the text with the script name and line number (perhaps as options), but I couldn't find a way to retrieve the name of the currently running script.

Any suggestions on how to overcome these?

PR #364, opened 2023-12-16
Add "as credited" sort variables to "Additional Artists Variables"

rdswift:
Add the following new variables to the "Additional Artists Variables" plugin in response to the "[Is there a way to get a renamed Artist’s alias’s sort order in Picard?](https://community.metabrainz.org/t/is-there-a-way-to-get-a-renamed-artists-aliass-sort-order-in-picard/670245)" discussion on the community forum.

### Album Variables

* **_artists_album_primary_cred_sort** - The primary / first album artist listed (sort name as credited)
* **_artists_album_additional_cred_sort** - All album artists listed (sort names as credited) except for the primary / first artist, separated by strings provided from the release entry
* **_artists_album_additional_cred_sort_multi** - All album artists listed (sort names as credited) except for the primary / first artist, as a multi-value
* **_artists_album_all_cred_sort** - All album artists listed (sort names as credited), separated by strings provided from the release entry
* **_artists_album_all_cred_sort_multi** - All album artists listed (sort names as credited), as a multi-value

### Track Variables

* **_artists_track_primary_cred_sort** - The primary / first track artist listed (sort as credited name)
* **_artists_track_additional_cred_sort** - All track artists listed (sort names as credited) except for the primary / first artist, separated by strings provided from the track entry
* **_artists_track_additional_cred_sort_multi** - All track artists listed (sort names as credited) except for the primary / first artist, as a multi-value
* **_artists_track_all_cred_sort** - All track artists listed (sort names as credited), separated by strings provided from the track entry
* **_artists_track_all_cred_sort_multi** - All track artists listed (sort names as credited), as a multi-value

The cred sort values are taken as the sort name from the alias that matches the "as credited" name if it exists, otherwise the sort name from the artist (standardized name).

### DIFF
--- a/plugins/additional_artists_variables/additional_artists_variables.py
+++ b/plugins/additional_artists_variables/additional_artists_variables.py
@@ -46,6 +46,10 @@ from picard.metadata import (register_album_metadata_processor,
 from picard.plugin import PluginPriority
 
 
+ID_ALIASES_ARTIST_NAME = '894afba6-2816-3c24-8072-eadb66bd04bc'
+ID_ALIASES_LEGAL_NAME = 'd4dcd0c0-b341-3612-a332-c0ce797b25cf'
+
+
 def process_artists(album_id, source_metadata, destination_metadata, source_type):
     # Test for valid metadata node.
     # The 'artist-credit' key should always be there.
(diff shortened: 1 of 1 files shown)


plugins/additional_artists_variables/additional_artists_variables.py
```diff
@@ -115,6 +121,9 @@ def process_artists(album_id, source_metadata, destination_metadata, source_type
                                     temp_legal_name = item['name']
                                 if 'sort-name' in item:
                                     temp_legal_sort_name = item['sort-name']
+                        if 'type-id' in item and item['type-id'] == '894afba6-2816-3c24-8072-eadb66bd04bc':
```
phw: Let's use a variable here for the type ID, so it's clear what it is. I guess it's the primary alias, right?

phw (changes requested):
@rdswift This looks good, thanks. Only change would be a variable for the type ID.

plugins/additional_artists_variables/additional_artists_variables.py
rdswift: Great idea!  Constant variables now set for both ID codes used.  This helps clarify the process significantly.

phw (approved):
Thanks. Let's merge this

PR #366, opened 2024-01-02
Fixes for "Generate M3U playlist" #245 , along with all reviews required 

### LINKED ISSUE #245: Fixes for "Generate M3U playlist"

sambhavnoobcoder:
fixes #245

made changes in the playlist.py to accommodate default playlist location to also work for multiple album selections. also takes into account the review changes from the past pr and makes sure that all necessary reviews from the last pr are accommodated in this pr .

### DIFF
--- a/plugins/abbreviate_artistsort/abbreviate_artistsort.py
+++ b/plugins/abbreviate_artistsort/abbreviate_artistsort.py
@@ -14,7 +14,7 @@
 # GNU General Public License for more details.
 
 PLUGIN_NAME = "Abbreviate artist-sort"
-PLUGIN_AUTHOR = "Sophist"
+PLUGIN_AUTHOR = "Sophist, Sambhav Dixit"
 PLUGIN_DESCRIPTION = '''Abbreviate Artist-Sort and Album-Artist-Sort Tags.
 e.g. "Vivaldi, Antonio" becomes "Vivaldi, A."
 This is particularly useful for classical albums that can have a long list of artists.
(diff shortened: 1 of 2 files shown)


sambhavnoobcoder:
@phw  kindly look into this and if any more changes are required , inform me of the same , i'll make them as well , and If it meets requirements , kindly merge the same .

phw:
@sambhavnoobcoder Thanks a lot for this. I'll look over this in detail later today.

sambhavnoobcoder:
this pr seems to have an issue ,as there are some other commits with it as well, ill close this pr and raise another pr with only the commit for this issue

PR #367, opened 2024-01-02
pr fixing issue exception when artist sorted name has no last name pa…

### LINKED ISSUE #350: [abbreviate_artistsort] exception when artist sorted name has no last name part
When an artist sorted name only consists of 1 word (e.g., "Josquin" (for Josquin des Prez"), the **abbreviate_artistsort** plugin throws an exception since `surname` is not defined in this case.

Here's an example:
```
Traceback (most recent call last):
  File "coverart\__init__.py", line 144, in next_in_queue
StopIteration

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "album.py", line 407, in _finalize_loading_track
  File "metadata.py", line 655, in run_track_metadata_processors
  File "plugin.py", line 257, in run
  File "C:\Users\Xxx\AppData\Local\MusicBrainz\Picard\plugins\abbreviate_artistsort.zip\abbreviate_artistsort.py", line 138, in abbreviate_artistsort
    while surname.split(None, 1)[0] == unsort.split(None, 1)[0]:
IndexError: list index out of range
Traceback (most recent call last):
  File "coverart\__init__.py", line 144, in next_in_queue
StopIteration
```

Consider checking for this edge case everywhere in the plugin.

sambhavnoobcoder:
…rt issue#350

fixes #350 

added conditional , that recognises a single word artist name different from one with a last name .In such cases, the code generates initials for this single word to serve as an abbreviated representation.The initials are formed by taking the first letter of each part of the single word and appending a dot ('.') after each letter.

moreover the code also handles the whitespaces in sorted and unsorted names .

### DIFF
--- a/plugins/abbreviate_artistsort/abbreviate_artistsort.py
+++ b/plugins/abbreviate_artistsort/abbreviate_artistsort.py
@@ -14,7 +14,7 @@
 # GNU General Public License for more details.
 
 PLUGIN_NAME = "Abbreviate artist-sort"
-PLUGIN_AUTHOR = "Sophist"
+PLUGIN_AUTHOR = "Sophist, Sambhav Dixit"
 PLUGIN_DESCRIPTION = '''Abbreviate Artist-Sort and Album-Artist-Sort Tags.
 e.g. "Vivaldi, Antonio" becomes "Vivaldi, A."
 This is particularly useful for classical albums that can have a long list of artists.
@@ -80,7 +80,6 @@ _abbreviate_cache = {}
 
 
 def abbreviate_artistsort(tagger, metadata, track, release):
-
     for sortTag, unsortTag, sortTagNew in _abbreviate_tags:
         if not (sortTag in metadata and unsortTag in metadata):
             continue
@@ -99,7 +98,6 @@ def abbreviate_artistsort(tagger, metadata, track, release):
             new_unsort = ""
 
             while len(sort) > 0 and len(unsort) > 0:
-
                 if not _split in sort:
                     log.debug("  Ending without separator '%s' - moving '%s'." % (_split, sort))
                     new_sort += sort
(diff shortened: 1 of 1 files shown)


plugins/abbreviate_artistsort/abbreviate_artistsort.py
```diff
@@ -146,88 +129,51 @@ def abbreviate_artistsort(tagger, metadata, track, release):
...
-                        _("%s: Abbreviated surname (%s, %s) to (%s, %s) in '%s'."),
-                        PLUGIN_NAME,
-                        surname,
-                        forename,
-                        surname,
-                        inits,
-                        sortTag,
-                    )
-                    log.debug("Abbreviated (%s, %s) to (%s, %s)." % (surname, forename, surname, inits))
-            else:  # while loop ended without a break i.e. no errors
```
Sophist-UK, line 211: Why did you remove this else statement? Removing this else statement changes the whole meaning of the statements that follow it.

Sophist-UK (commented):
Sorry - but I am unable to review this because it contains multiple changes in a single commit and it is impossible to relate changes made to the problem they are fixing:

a. To fix the exception reported in #350;
b. To do other stuff

There are no examples that have been provided to test the use cases that have changed both before and after changes, nor examples that can be used to test that functionality has not changed for existing releases that work just fine.

In addition, there seems to be a `for ... else` statement that has been removed but I cannot determine why this was removed and what the impact would be.

sambhavnoobcoder:
okay , I'll take this into account and redraft smaller pr's in a systematic manner to make it easier to review and more step by step to resolve any issues in the code as well . in doing so , I'll also provide with the adequate testing use cases for both before and after the changes to assist with testing the changes in the code . thanks .

phw:
@sambhavnoobcoder Yes, that would be great. I find this a bit difficult to review as it is, but you already had to wrap you head around all the functionality here. Especially some real test cases, if you have some, would be useful.

eharris:
I would love to see this get fixed, having quite a few errors with this plugin.

PR #369, opened 2024-01-18
PICARD-2032: Set albumsort/titlesort tags

twodoorcoupe:
This plugin attempts to fix [[PICARD-2032]](https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-2032). For each track or album it first checks if there are any sort names already available in the aliases, otherwise it swaps the title's prefix.

Then it also partly fixes [[PICARD-2779]](https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-2779) by providing script functions: 
- $swapprefix_lang 
- $delprefix_lang
- $title_lang

So far, the languages available are English, Spanish, Italian, French, German, and Portuguese. For each language there's a list of articles that are used as prefixes and a list of short prepositions and conjunctions that are used in $title_lang. These I wrote myself by looking online, so they are prone to errors.

$swapprefix_lang and $delprefix_lang do the same thing as their original counterparts but take languages as inputs instead of prefixes. $title_lang returns the text in title case by keeping the small words of each language in lower case.
During tagging, if the title's prefix is swapped, only the track's or album's language are considered. If none are found then all of the available ones are considered.

Checking the aliases for each track and album can be turned off in the options, as well as ignoring titles in all caps for $title_lang.
It's my first time contributing so I apologize if I missed anything important.

### DIFF
--- a/plugins/enhanced_titles/__init__.py
+++ b/plugins/enhanced_titles/__init__.py
@@ -1,0 +1,331 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024 Giorgio Fontanive (twodoorcoupe)
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+PLUGIN_NAME = "Enhanced Titles"
+PLUGIN_AUTHOR = "Giorgio Fontanive"
+PLUGIN_DESCRIPTION = """
+This plugin sets the albumsort and titlesort tags. It also provides the script
+functions $swapprefix_lang, $delprefix_lang and $title_lang. The languages
+included at the moment are English, Spanish, Italian, German, French and Portuguese.
+
+The functions do the same thing as their original counterparts, but take
+multiple languages as parameters. If no languages are provided, all the available ones are
+included. Languages are provided with ISO 639-3 codes: eng, spa, ita, fra, deu, por.
+
+Tagging and checking aliases can be disabled in the plugin's options page, found 
+under "plugins". Checking aliases will slow down processing.
+"""
+PLUGIN_VERSION = "0.1"
+PLUGIN_API_VERSIONS = ["2.10"]
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
+
+from picard import log, config
+from picard.plugin import PluginPriority
+from picard.metadata import register_album_metadata_processor, register_track_metadata_processor
+from picard.script import register_script_function
+from picard.script.functions import func_swapprefix, func_delprefix
+from picard.webservice.api_helpers import MBAPIHelper
+
+from picard.ui.options import OptionsPage, register_options_page
+from .ui_options_enhanced_titles import Ui_EnhancedTitlesOptions
+
+from functools import partial
+import re
+
+# Options.
+KEEP_ALLCAPS = "et_keep_allcaps"
+ENABLE_TAGGING = "et_enable_tagging"
+CHECK_ALBUM = "et_check_album_aliases"
+CHECK_TRACK = "et_check_track_aliases"
+
+# ISO 639-3 language codes.
+ADDED_LANGUAGES = {"eng", "spa", "ita", "fra", "deu", "por"}
+
+_articles = {
+    "eng" : ["the", "a", "an"],
+    "spa" : ["el", "los", "la", "las", "lo", "un", "unos", "una", "unas"],
+    "ita" : ["il", "l'", "la", "i", "gli", "le", "un", "uno", "una", "un'"],
+    "fra" : ["le", "la", "les", "un", "une", "des", "l'"],
+    "deu" : ["der", "den", "die", "das", "dem", "des", "den"],
+    "por" : ["o", "os", "a", "as", "um", "uns", "uma", "umas"]
+}
+
+# Prepositions and conjunctions with 3 letters or less.
+_other_minor_words = {
+    "eng" : ["so", "yet", "as", "at", "by", "for", "in", "of", "off", "on", "per",
+            "to", "up", "via", "and", "as", "but", "for", "if", "nor", "or"],
+    "spa" : ["mas", "que", "que", "en", "con", "por", "de", "y", "e", "o", "u", "si", "ni"],
+    "ita" : ["di", "a", "da", "in", "con", "su", "per", "tra", "fra", "e", "o", "ma", "se"],
+    "fra" : ["à", "de", "en", "par", "sur", "et", "ou", "que", "si"],
+    "deu" : ["bis", "für", "um", "an", "auf", "in", "vor", "aus", "bei", "mit", 
+            "von", "zu", "la", "so", "daß", "als", "ob", "ehe"],
+    "por" : ["dem", "em", "por", "ao", "à", "aos", "às", "do", "da", "dos", "das",
+            "no", "na", "nos", "nas", "num", "dum", "e", "mas", "até", "em", "ou",
+            "que", "se", "por"]
+}
+
+
+class ReleaseGroupHelper(MBAPIHelper):
+    """API Helper to retreive release group information.
+    """
+    
+    def get_release_group_by_id(self, release_id, handler, inc = None):
+        """Gets the information for a release group.
+        """
+        return self._get_by_id("release-group", release_id, handler, inc)
+    
+
+class SortTagger:
+    """Sets the titlesort and albumsort tags.
+        
+    First, it checks if there is already a sort name available in one of the
+    aliases. If it does not find any, it swaps the prefix if there is one in
+    the title.
+    """
+    
+    def _select_alias(self, aliases, name):
+        """Selects the first alias where the names match and the sort name is
+        different.
+        
+        Args:
+            aliases (list): One dictionary for each alias available.
+            name (str): Title of the album/track.
+            
+        Returns:
+            (str): The sort name of the first useful alias it finds. None if
+                   none are found.
+            
+        For example, "The Beatles" has alias "Le Double Blanc" with sort name 
+        "Double Blanc, Le", so it's not considered. But it also has alias
+        "The Beatles" with sort name "Beatles, The", so this one is chosen.
+        Another example, "The Continuing Story of Bungalow Bill" has an alias
+        with the sort name equal to the title, this is not considered because
+        it makes more sense to swap the prefix.
+        """
+        for alias in aliases:
+            sortname = alias["sort-name"]
+            if (alias["name"].casefold() == name.casefold() and 
+                    not sortname.casefold() == name.casefold()):
+                log.info("Enhanced Titles: sort name found for \"" + name + "\", \"" + sortname + "\".")
+                return sortname
+        log.info("Enhanced Titles: no proper sort name found for \"" + name + "\".")
+        return None
+            
+    def _response_handler(self, document, reply, error, metadata = None, field = None):
+        """Handles the response from MusicBrainz.
+        
+        Args:
+            metadata (MetaData): The object that needs to be updated.
+            field (str): Either "title" or "album", depending on what is being
+                         updated.
+        """
+        sortname = ""
+        try:
+            if document:
+                if error:
+                    log.error("Enhanced Titles: information retrieval error.")
+                if document["aliases"]:
+                    sortname = self._select_alias(document["aliases"], metadata[field])
+                else:
+                    log.info("Enhanced Titles: no aliases found for \"" + metadata[field] + "\".")
+        finally:
+            if sortname:
+                sortfield = field + "sort"
+                metadata[sortfield] = sortname
+            else:
+                self._swapprefix(metadata, field)
+    
+    def _swapprefix(self, metadata, field):
+        """Swaps the prefix of the title based on the album/track language."
+        
+        If no language information is found, then it uses all languages available.
+        Otherwise, if none of the languages are available, it just copies the title.
+        Otherwise it uses exclusively the languages that are also available.
+        """
+        sortfield = field + "sort"
+        languages = [metadata["language"], metadata["_releaselanguage"]]
+        languages = [language for language in languages if language]
+        if not languages:
+            metadata[sortfield] = func_swapprefix(None, metadata[field], *_create_prefixes_list())
+        else:
+            languages = [language for language in languages if language in ADDED_LANGUAGES]
+            if not languages:
+                metadata[sortfield] = metadata[field]
+            else:
+                metadata[sortfield] = func_swapprefix(None, metadata[field], *_create_prefixes_list(languages))
+
+    def set_track_titlesort(self, album, metadata, track, release):
+        """Sets the track's titlesort field.
+        """
+        if config.setting[ENABLE_TAGGING]:
+            handler = partial(self._response_handler, metadata = metadata, field = "title")
+            if config.setting[CHECK_TRACK]:
+                MBAPIHelper(album.tagger.webservice).get_track_by_id(
+                    metadata["musicbrainz_recordingid"], 
+                    handler, 
+                    inc = ["aliases"]
+                )
+            else:
+                self._swapprefix(metadata, "title")
+
+    def set_album_titlesort(self, album, metadata, release):
+        """Sets the album's albumsort field.
+        """
+        if config.setting[ENABLE_TAGGING]:
+            handler = partial(self._response_handler, metadata = metadata, field = "album")
+            if config.setting[CHECK_ALBUM]:
+                ReleaseGroupHelper(album.tagger.webservice).get_release_group_by_id(
+                    metadata["musicbrainz_releasegroupid"], 
+                    handler, 
+                    inc = ["aliases"]
+                )
+            else:
+                self._swapprefix(metadata, "album")
+
+
+def _create_prefixes_list(languages = None, is_title = False):
+    """Creates a list of all the prefixes or minor words for all the given languages.
+    
+    Args:
+        languages (list): All the languages to be considered.
+        is_title (bool): If true, only articles are included, with a lower case
+                         and a capitalized copy for each. This is because the
+                         swapprefix and delprefix functions are case sensitive.
+                         If false, also prepositions and conjunctions are included,
+                         all in lowercase.
+                         
+    Returns:
+        (set): The set of prefixes or minor words.
+        
+    The available languages are saved with ISO 639-3 codes.
+    """
+    prefixes = set()
+    if not languages:
+        languages = ADDED_LANGUAGES
+    else:
+        languages = [lang.lower()[:3] for lang in languages]
+        languages = [lang for lang in languages if lang in ADDED_LANGUAGES]
+    for language in languages:
+        prefixes.update(_articles[language])
+        if is_title:
+            prefixes.update(_other_minor_words[language])
+        else:
+            prefixes.update([article.capitalize() for article in _articles[language]])
+    return prefixes
+
+def _title_case(text, lower_case_words):
+    """Returns the text in titlecase.
+    
+    If a word has an apostrophe and the segment to its left is an article,
+    it capitalizes only the word on the right. Otherwise it capitalizes
+    only the word on the left.
+    For example, "let's groove" becomes "Let's Groove", but "voglio l'anima"
+    becomes "Voglio l'Anima".
+    """
+    new_text = []
+    words = re.split(r"([\w']+)", text.strip().lower().replace("’", "'"))
+    words = [word for word in words if word]
+    for word in words:
+        if "'" in word: # Apply the rule described above
+            split = word.split("'")
+            if split[0] + "'" in lower_case_words:
+                split[1] = split[1].capitalize()
+            else:
+                split[0] = split[0].capitalize()
+            word = "'".join(split)
+        elif not word in lower_case_words:
+            word = word.capitalize()
+        new_text.append(word)
+    if new_text:
+        new_text[0] = new_text[0].capitalize()
+        return "".join(new_text)
+    else:
+        return ""
+
+swapprefix_lang_documentation = N_(
+    """`$swapprefix_lang(text,language1,language2,...)`
+
+Moves the prefix to the end of the text. It uses a list prefixes
+taken from the specified languages.
+Multiple languages can be added as seperate parameters.
+If none are provided, it uses all the available ones.
+    """)
+def swapprefix_lang(parser, text, *languages):
+    return func_swapprefix(parser, text, *_create_prefixes_list(languages)) if text else ""
+
+delprefix_lang_documentation = N_(
+    """`$delprefix_lang(text,language1,language2,...)`
+    
+Deletes the prefix from the text. It uses a list prefixes
+taken from the specified languages.
+Multiple languages can be added as seperate parameters.
+If none are provided, it uses all the available ones.
+    """)
+def delprefix_lang(parser, text, *languages):
+    return func_delprefix(parser, text, *_create_prefixes_list(languages)) if text else ""
+
+title_lang_documentation = N_(
+    """`$title_lang(text,language1,language2,...)`
+    
+Makes the text title case based on the minor words of the specified languages.
+Multiple languages can be added as seperate parameters.
+If none are provided, it uses all the available ones.
+    """)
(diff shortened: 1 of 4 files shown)


plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,331 @@
...
+import re
+
+# Options.
+KEEP_ALLCAPS = "et_keep_allcaps"
+ENABLE_TAGGING = "et_enable_tagging"
+CHECK_ALBUM = "et_check_album_aliases"
+CHECK_TRACK = "et_check_track_aliases"
+
+# ISO 639-3 language codes.
+ADDED_LANGUAGES = {"eng", "spa", "ita", "fra", "deu", "por"}
```
Sophist-UK: This constant is unnecessary. The languages supported are the keys of the `_articles` and `_other minor_words` dicts.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,331 @@
...
+                ReleaseGroupHelper(album.tagger.webservice).get_release_group_by_id(
+                    metadata["musicbrainz_releasegroupid"], 
+                    handler, 
+                    inc = ["aliases"]
+                )
+            else:
+                self._swapprefix(metadata, "album")
+
+
+def _create_prefixes_list(languages = None, is_title = False):
```
Sophist-UK: IMO, this is coded incorrectly. This function is called each and every time a script function is called and simply repeats the exact same code each time.

IMO the set of words for all languages should be calculated *once, at load time* rather than each time a script function is called. Personally I would store it in the same dict as the separate languages with an empty string as the key.

Sophist-UK (changes requested):
In addition to the other comments, I would like to see some unit tests for the new script functions.

Sophist-UK:
P.S. @twodoorcoupe Giorgio - Many thanks for coding this when I don't have the time. 👍

twodoorcoupe:
Thank you @Sophist-UK  for taking the time to review this. I made the changes you requested and added some unit tests. Now, when a new combination of languages is used, the resulting list of prefixes or minor words is stored so as not to recalculate it every time.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+        For example, "The Beatles" has alias "Le Double Blanc" with sort name
+        "Double Blanc, Le", so it's not considered. But it also has alias
+        "The Beatles" with sort name "Beatles, The", so this one is chosen.
+        Another example, "The Continuing Story of Bungalow Bill" has an alias
+        with the sort name equal to the title, this is not considered because
+        it makes more sense to swap the prefix.
+        """
+        for alias in aliases:
+            sortname = alias["sort-name"]
+            if (alias["name"].casefold() == name.casefold() and
```
zas: `name.casefold()` has constant value inside the loop, better store it in a variable before `for alias in aliases:`

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+                    not sortname.casefold() == name.casefold()):
+                log.info("Enhanced Titles: sort name found for \"" + name + "\", \"" + sortname + "\".")
```
zas: `log.info()` can use % format strings (see https://github.com/metabrainz/picard/blob/90a8ee5a0d4d72d68a197195fdc9f181643fc6c8/picard/browser/browser.py#L121 for an example)

```suggestion
                log.info('Enhanced Titles: sort name found for "%s", "%s"', name, sortname)
```


That's much easier to read.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+                return sortname
+        log.info("Enhanced Titles: no proper sort name found for \"" + name + "\".")
```
zas: Same here

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+    Combinations of languages and their prefixes/minor words are stored after
+    they're used at least once.
+    """
+
+    prefixes_lists = {}
+    titles_lists = {}
+
+    def _format_languages(self, languages):
+        languages = [lang.lower()[:3] for lang in languages]
+        languages = [lang for lang in languages if lang in _articles.keys()]
```
zas: Iterator for dict defaults to keys:

```suggestion
        languages = [lang for lang in languages if lang in _articles]
```

zas: Perhaps it could be done differently (it's prolly faster) using `set.intersection`:

```python
langs = set(lang[:3].lower() for lang in languages)
articles_langs = set(_articles)
return tuple(langs.intersection(articles_langs))
```

Also better truncating and apply `lower()` on 3 letters, than lowering the full string and truncating to 3 letters.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def _response_handler(self, document, reply, error, metadata = None, field = None):
+        """Handles the response from MusicBrainz.
+
+        Args:
+            metadata (MetaData): The object that needs to be updated.
+            field (str): Either "title" or "album", depending on what is being
+                         updated.
+        """
+        sortname = ""
```
zas:
```suggestion
        sortname = None
```

Since `_select_alias()` may also return `None` it makes more sense.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+        Otherwise it uses exclusively the languages that are also available.
+        """
+        sortfield = field + "sort"
+        languages = [metadata["language"], metadata["_releaselanguage"]]
+        languages = [language for language in languages if language]
+        if not languages:
+            prefixes = _articles[""] + [article.capitalize() for article in _articles[""]]
+            metadata[sortfield] = func_swapprefix(None, metadata[field], *prefixes)
+        else:
+            languages = [language for language in languages if language in _articles.keys()]
```
zas: Isn't this equivalent to:

```suggestion
            languages = set(languages).intersection(_articles)
```

zas (changes requested):

plugins/enhanced_titles/__init__.py
```diff
@@ -61,53 +58,55 @@
     "deu" : ["der", "den", "die", "das", "dem", "des", "den"],
     "por" : ["o", "os", "a", "as", "um", "uns", "uma", "umas"]
 }
+_articles[""] = [value for values in _articles.values() for value in values]
```
Sophist-UK: I have a feeling that this can be simplified using the iterable unpacking operator `*` to:
```python
_articles[""] = [*values for values in _articles.values()]
```

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+                else:
+                    log.info("Enhanced Titles: no aliases found for \"" + metadata[field] + "\".")
+        finally:
+            if sortname:
+                sortfield = field + "sort"
+                metadata[sortfield] = sortname
+            else:
+                self._swapprefix(metadata, field)
+
+    def _swapprefix(self, metadata, field):
```
Sophist-UK, line 165: This would IMO be better of as a function taking the field value and returning a sorted value, leaving the caller to get the unsorted value and store the sorted value i.e.
```python
    def _swapprefix(self, field):
        ....
        return func_swapprefix(None, metadata[field], *prefixes)

    metadata["titlesort"] = self._swapprefix(metadata["title"])
```

Also, using return enables early return improved readability avoiding multiple nested if/else i.e.
```python
if something:
    ...
    return first_something

if something_else:
    return second_something

return third something
```

plugins/enhanced_titles/__init__.py
Sophist-UK: I would suggest that `_articles_langs = set(_articles)` would be better off done once immediately after `_articles[""]` is defined.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+    they're used at least once.
+    """
+
+    prefixes_lists = {}
+    titles_lists = {}
+
+    def _format_languages(self, languages):
+        languages = [lang.lower()[:3] for lang in languages]
+        languages = [lang for lang in languages if lang in _articles.keys()]
+        return tuple(languages)
```
Sophist-UK: I am unclear what benefit there is to converting languages into a `tuple` rather than returning it as a `list`.

plugins/enhanced_titles/__init__.py
```diff
@@ -50,9 +50,6 @@
 CHECK_ALBUM = "et_check_album_aliases"
 CHECK_TRACK = "et_check_track_aliases"
 
-# ISO 639-3 language codes.
-ADDED_LANGUAGES = {"eng", "spa", "ita", "fra", "deu", "por"}
-
 _articles = {
     "eng" : ["the", "a", "an"],
```
Sophist-UK: Would it be better for these lists of words to be `tuple`s rather than `list`s?

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+                             swapprefix and delprefix functions are case sensitive.
+                             If false, also prepositions and conjunctions are included,
+                             all in lowercase.
+
+        Returns:
+            (set): The set of prefixes or minor words.
+
+        The available languages are saved with ISO 639-3 codes.
+        """
+        prefixes = set()
```
Sophist-UK: Because `set`s are immutable, would this code be better:
```python
        prefixes = []
        for language in languages:
            prefixes.extend(_articles[language])
            if is_title:
                prefixes.extend(_other_minor_words[language])
            else:
                prefixes.extend([article.capitalize() for article in _articles[language]])
        return set(prefixes)
```

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+            return ""
+
+    def find_prefixes(self, languages):
+        """Returns the list of prefixes for the given languages.
+
+        If the same combination was previously used, it finds the values stored
+        to avoid recreating the same list twice.
+        """
+        if not languages or "" in languages:
+            return set(_articles[""] + [article.capitalize() for article in _articles[""]])
```
Sophist-UK: This should also be cached in `self.prefixes_list`.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+
+
+class LangFunctions:
+    """Provides scripting functions swapprefix_lang, delprefix_lang, and title_lang.
+
+    Combinations of languages and their prefixes/minor words are stored after
+    they're used at least once.
+    """
+
+    prefixes_lists = {}
```
Sophist-UK: Can we call this `prefixes_cache` to indicate that caching is its function?

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+    titles_lists = {}
```
Sophist-UK: Similarly this should probably be called `minor_words_cache`.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+        return prefixes
+
+    def find_minor_words(self, languages):
+        """Returns the list of minor words for the given languages.
+
+        If the same combination was previously used, it finds the values stored
+        to avoid recreating the same list twice.
+        """
+        if not languages or "" in languages:
+            return set(_articles[""] + _other_minor_words[""])
```
Sophist-UK: Similarly this should also be cached.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+        'the prefix is not in Spanish'
+        >>> lf.swapprefix_lang(None, "Il prefisso è scambiato", "ita")
+        'prefisso è scambiato, Il'
+        >>> lf.swapprefix_lang(None, "the-prefix-is-not-swapped")
+        'the-prefix-is-not-swapped'
+        >>> lf.swapprefix_lang(None, "the")
+        'the'
+        """
+        prefixes = self.find_prefixes(languages)
+        return func_swapprefix(parser, text, *prefixes) if text else ""
```
Sophist-UK: If `text` is empty, doesn't `func_swapprefix` already return ""?

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+        'the prefix is not in Spanish'
+        >>> lf.delprefix_lang(None, "Il prefisso è eliminato", "ita")
+        'prefisso è eliminato'
+        >>> lf.delprefix_lang(None, "the-prefix-is-not-deleted")
+        'the-prefix-is-not-deleted'
+        >>> lf.delprefix_lang(None, "the")
+        'the'
+        """
+        prefixes = self.find_prefixes(languages)
+        return func_delprefix(parser, text, *prefixes) if text else ""
```
Sophist-UK: Ditto

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+        "Let's Groove"
+        >>> lf.title_lang(None, "voglio l'anima")
+        "Voglio l'Anima"
+        >>> lf.title_lang(None, "casiopea, カシオペア")
+        'Casiopea, カシオペア'
+        """
+        if text.upper() == text and config.setting[KEEP_ALLCAPS]:
+            return text
+        minor_words = self.find_minor_words(languages)
+        return self._title_case(text, minor_words) if text else ""
```
Sophist-UK: Ditto

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+    "eng" : ["the", "a", "an"],
+    "spa" : ["el", "los", "la", "las", "lo", "un", "unos", "una", "unas"],
+    "ita" : ["il", "l'", "la", "i", "gli", "le", "un", "uno", "una", "un'"],
+    "fra" : ["le", "la", "les", "un", "une", "des", "l'"],
+    "deu" : ["der", "den", "die", "das", "dem", "des", "den"],
+    "por" : ["o", "os", "a", "as", "um", "uns", "uma", "umas"]
+}
+_articles[""] = [value for values in _articles.values() for value in values]
+
+# Prepositions and conjunctions with 3 letters or less.
```
Sophist-UK: Might also be helpful say that these are words which aren't title-cased.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+
+        Args:
+            aliases (list): One dictionary for each alias available.
+            name (str): Title of the album/track.
+
+        Returns:
+            (str): The sort name of the first useful alias it finds. None if
+                   none are found.
+
+        For example, "The Beatles" has alias "Le Double Blanc" with sort name
```
Sophist-UK: I got confused here as to why the 1960s' Hip Beat Combo "The Beatles" would be called "The Double White" in French, particularly considering that the group consisted of 4 white blokes not two. Then it clicked that you mean the nicknamed "White Album" which was officially entitled "The Beatles". To avoid anyone as stupid as me thinking the same, you might want to say 'For example, the album "The Beatles"...'.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,429 @@
...
+    "spa" : ["el", "los", "la", "las", "lo", "un", "unos", "una", "unas"],
+    "ita" : ["il", "l'", "la", "i", "gli", "le", "un", "uno", "una", "un'"],
+    "fra" : ["le", "la", "les", "un", "une", "des", "l'"],
+    "deu" : ["der", "den", "die", "das", "dem", "des", "den"],
+    "por" : ["o", "os", "a", "as", "um", "uns", "uma", "umas"]
+}
+_articles[""] = [value for values in _articles.values() for value in values]
+
+# Prepositions and conjunctions with 3 letters or less.
+_other_minor_words = {
```
Sophist-UK, line 68: I18n Title Casing is complicated. For example, see https://www.thoughtco.com/capitalize-french-titles-4086495 for french rules. I am not going to research and review rules for all 100+ languages, and I don't expect the author to do this either. But I do want to flag this for future reference.

Leaving aside the above, the definitive (american) English list of lower case words is apparently defined by the New York Times Manual of Style, and https://github.com/ppannuto/python-titlecase/blob/main/titlecase/__init__.py uses:
```python
SMALL = r'a|an|and|as|at|but|by|en|for|if|in|of|on|or|the|to|v\.?|via|vs\.?'
```
Your list has more than this, but seems to be missing some of these also.

I am not even going to attempt to take a position on what should and should not be in this list. But I will warn that the language nerds will likely jump on this list - and even if you research extensively, whilst you will reduce the number of comments, regardless of your diligence the language nerds will demand changes.

To avoid your own headaches, you might wish to add a comment suggesting that if anyone wants to see changes to the minor_words list they should code and submit a PR for the changes they believe are necessary together with links to web pages that give definitive language-specific title-case rules.

Sophist-UK (changes requested):
Despite for the length and number of comments here, these are relatively minor tweaks to what is already pretty good code. So thanks again for writing this.

Sophist-UK:
@zas, can you enable workflow for @twodoorcoupe so that we can see his unit tests running?

plugins/enhanced_titles/__init__.py
zas: Yes, those aren't expected to be modified anyway, so `tuple` would fit better.

plugins/enhanced_titles/__init__.py
twodoorcoupe: This is so the tuple can be used as a key for the cache dictionary, since it's immutable.

plugins/enhanced_titles/__init__.py
twodoorcoupe: I had thought about doing that, but apparently it's not allowed, as it returns a syntax error: "Unpacking cannot be used in comprehension."

twodoorcoupe:
All the tweaks you suggested should be implemented now. Many thanks for all the tips.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,445 @@
...
+            "von", "zu", "la", "so", "daß", "als", "ob", "ehe"),
+    "por": ("dem", "em", "por", "ao", "à", "aos", "às", "do", "da", "dos", "das",
+            "no", "na", "nos", "nas", "num", "dum", "e", "mas", "até", "em", "ou",
+            "que", "se", "por")
+}
+_other_minor_words[""] = tuple(value for values in _other_minor_words.values() for value in values)
+
+
+class ReleaseGroupHelper(MBAPIHelper):
+    """API Helper to retreive release group information.
```
zas:
```suggestion
    """API Helper to retrieve release group information.
```

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,445 @@
...
+
+_articles = {
+    "eng": ("the", "a", "an"),
+    "spa": ("el", "los", "la", "las", "lo", "un", "unos", "una", "unas"),
+    "ita": ("il", "l'", "la", "i", "gli", "le", "un", "uno", "una", "un'"),
+    "fra": ("le", "la", "les", "un", "une", "des", "l'"),
+    "deu": ("der", "den", "die", "das", "dem", "des", "den"),
+    "por": ("o", "os", "a", "as", "um", "uns", "uma", "umas")
+}
+_articles[""] = tuple(value for values in _articles.values() for value in values)
```
zas: We could define a method to do that (as it is used below too):

```python
def flatten_values(dictionnary):
    l = list()
    for v in dictionnary.values():
        l += v
    return tuple(l)

_articles[""] = flatten_values(_articles)
_other_minor_words[""] = flatten_values(_other_minor_words)
```


About this I wonder two things:

- is there a reason to not use `set` (over `tuple`) for dict values?
- is the extra `""` key actually of any use? I mean could we store all articles in a separate variable `all_articles`?

zas (changes requested):
Things look good to me, just few minor comments

plugins/enhanced_titles/__init__.py
Sophist-UK: > ```python
> def flatten_values(dictionnary):
> ```
Typo - single "n" in dictionary.

Sophist-UK: > * is there a reason to not use `set` (over `tuple`) for dict values?

We almost certainly should use `set` rather than `tuple` for the flattened values because this will eliminate unnecessary duplicates. And if we use `set` for this then for consistency we should probably use `set` for all of the dict entries.

Sophist-UK: > * is the extra `""` key actually of any use? I mean could we store all articles in a separate variable `all_articles`?

At the time I suggested this I was expecting to be able to use "" as an index if the user doesn't specify any languages. But the end code doesn't do this, so we could user a variable instead or we could change the code to use it as an index.

twodoorcoupe: Regarding `set`s over `tuple`s, the only reason I can think of for preferring `tuple`s is that they are immutable, and the values of the dictionary aren't meant to change at runtime.
For the second point, I think storing all the articles and minor words in a separate variable is more readable.

Sophist-UK: Sets are also immutable and they remove duplicates which makes sense in this context.

plugins/enhanced_titles/__init__.py
phw: The above is one reason why I am hesitate to add something like this to Picard's core and think such a title case function is better suited for a plugin, which more easily can be updated, changed and replaced.

For a language like German title casing does not even make any real sense at all. It's not something that is being done in  German typography and doing so makes things really confusing to read. Non-native speakers will want to do it anyway :)

phw (commented):
Code and functionality look good to me, thanks a lot for this plugin.

I wonder though, whether the script functions should also try to use the `language` from the metadata instead of using essentially all if no language is explicitly given. The same as the swapprefix in `SortTagger` already does.

Or is there some specific reason not to do this?

twodoorcoupe:
Ah that makes sense and now I remembered it was actually proposed by @Sophist-UK in [[PICARD-2779]](https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-2779) as an enhancement of $swapprefix. I moved the logic for finding the languages from `_swapprefix` in `SortTagger` to a new function `find_languages` that is then used by the script functions. Thank you @phw for the help.

plugins/enhanced_titles/__init__.py
```diff
@@ -0,0 +1,470 @@
...
+    def _swapprefix(self, metadata, field):
+        """Swaps the prefix of the title based on the album/track language.
+
+        If no language information is found, then it uses all languages available.
+        Otherwise, if none of the languages are available, it just copies the title.
+        Otherwise, it uses exclusively the languages that are also available.
+        """
+        languages = lang_functions.find_languages(metadata)
+        if languages and "" in languages:
+            return metadata[field]
```
Sophist-UK, line 174: This logic appears different to before. Shouldn't it be returning `func_swapprefix(None, metadata[field], self._all_articles)`?

Sophist-UK (changes requested):

plugins/enhanced_titles/__init__.py
twodoorcoupe: The logic is the same, I just needed a way to differentiate the case in which no language information is found in the metadata, and the one in which none of the languages found are part of the set of the languages available. I apologize for the confusion. Now I should have fixed the issue by making the comments clearer.

Also, somehow your other comments about using `set`s didn't show up for me until you reviewed this change. I followed your advice on that as well. Again, thanks for taking the time to review this.

Sophist-UK: OK

Sophist-UK:
I think this looks good to go too.

zas (approved):
LGTM

Sophist-UK:
@twodoorcoupe Thank you again for all your efforts on this - including your willingness to rework it several times to satisfy the petty demands of grumpy old coders like me.

phw (approved):
Thanks a lot. Looks good to me as well. I'll merge and the plugin will be available from the website sometime the next 12 hours.

Sophist-UK:
Sorry to revisit this over 2 years later however the ticket for this [Picard-2032](https://tickets.metabrainz.org/browse/PICARD-2032) came up in my email unexpectedly and I came to look at this plugin to remind myself what it was all about.

When doing so, I spotted a few potential opportunities for minor enhancements:

1. The description mentions twice that this plugin may slow down loading of music files, but I cannot really see why this should be the case to any noticeable extent. However the code does seem to do some massaging of prefixes and other joining words for every track when it seems to me that it could potentially be done once at startup time or the massaged values could be cached, either of which would avoid calling e.g. `_format_languages()` multiple times. How much this would speed up the plugin I am unsure.

2. The selection of sort titles from aliases is done without regard to the release language or the users preferred language or any other language indicator - so it seems to me that you might, for example, randomly get a japanese alias which has the sort flag set.

PR #370, opened 2024-01-22
PICARD-2821: Playlist: Fix crash if playlist and audio file are on different drives

phw:
Fixes crash on Windows in playlist plugin if audio file and playlist are on different drives.

See PICARD-2821

https://docs.python.org/3/library/os.path.html#os.path.commonpath
https://docs.python.org/3/library/os.path.html#os.path.relpath


phw:
I tested on Linux and Windows (there with same and different drives). Also tested by chaban, see PICARD-2821

zas (approved):

PR #372, opened 2024-02-06
Remediate / refactor wikidata plugin

Sophist-UK:
1. Reduce log verbosity, particularly INFO but also DEBUG, and provide additional information.

2. Refactor code to reduce nesting, primarily using early returns, but also simplifying and generally cleaning.

3. Fix a few small bugs whose symptoms would probably not be noticed.

I have tested this version, however I don't have not yet found any of my music that has Wikidata genres.

### DIFF
--- a/plugins/wikidata/__init__.py
+++ b/plugins/wikidata/__init__.py
@@ -11,7 +11,7 @@ PLUGIN_DESCRIPTION = '''
 Query wikidata to get genre tags.
 <br/><br/>
 Since genres can be gathered from multiple sources and the genre tag may be overwritten,
-we also set a hidden variable _wikidata_genre to the same values for use in scripts.
+we also set a hidden variable _genre_wikidata to the same values for use in scripts.
 '''
 PLUGIN_VERSION = '1.4.6'
 PLUGIN_API_VERSIONS = ["2.0", "2.1", "2.2"]
@@ -124,7 +124,7 @@ class Wikidata:
             new_genre = set(metadata.getall("genre"))
             new_genre.update(genre_list)
             #sort the new genre list so that they don't appear as new entries (not a change) next time
-            metadata["genre"] = metadata["~wikidata_genre"] = self.genre_delimiter.join(sorted(new_genre))
+            metadata["genre"] = metadata["~genre_wikidata"] = self.genre_delimiter.join(sorted(new_genre))
             return
 
         # pending requests are handled by adding the metadata object to a
@@ -233,7 +233,7 @@ class Wikidata:
                     genres = [g for gs in genres for g in gs.split(self.genre_delimiter)]
                 genres += genre_list
                 # eliminate duplicates and sort to ensure consistency
-                genres = metadata["~wikidata_genre"] = sorted(set(genres))
+                genres = metadata["~genre_wikidata"] = sorted(set(genres))
                 log.info('WIKIDATA: Setting metadata genre for %s: %s', metadata['musicbrainz_trackid'], genres)
                 if self.genre_delimiter:
                     metadata["genre"] = self.genre_delimiter.join(genres)
--- a/plugins/wikidata/__init__.py
+++ b/plugins/wikidata/__init__.py
@@ -6,9 +6,13 @@
 # as published by Sam Hocevar. See http://www.wtfpl.net/ for more details.
 
 PLUGIN_NAME = 'Wikidata Genre'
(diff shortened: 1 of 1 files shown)


Sophist-UK:
The MB data I used to test this was [release:0cc705d1-ea84-4a28-b6f5-812e295129f0](https://musicbrainz.org/release/0cc705d1-ea84-4a28-b6f5-812e295129f0).

Sophist-UK:
Ok - this is now working perfectly as far as I can tell. Review may be difficult as it is both a refactor and contains various bug fixes, most notably when wikidata network errors occur, the album will never resolve.

Additional functionality: Avoid duplicate albumartist and artist lookups

Sophist-UK:
As an aside, I am unsure how it has a codacy issue of a method being over complex when the refactoring reduced the complexity of the code substantially. However, if needed I can refactor it a little more to reduce the complexity further.

plugins/wikidata/__init__.py
```diff
@@ -93,229 +97,233 @@ def __init__(self):
...
+                        log.info('WIKIDATA: wikidata url found for RELEASE_GROUP: %s', wikidata_url)
+                        self.process_wikidata(Wikidata.RELEASE_GROUP, wikidata_url, item_id)
+        if 'artist' in response.metadata[0].children and self.use_artist_genres:
+            if 'relation_list' in response.metadata[0].artist[0].children:
+                for relation in response.metadata[0].artist[0].relation_list[0].relation:
+                    if relation.type == 'wikidata' and 'target' in relation.children:
+                        found = True
+                        wikidata_url = relation.target[0].text
+                        self.process_wikidata(Wikidata.ARTIST, wikidata_url, item_id)
+                        log.info('WIKIDATA: wikidata url found for ARTIST: %s', wikidata_url)
```
phw, line 177: This is really minor, but the other code paths have this log message before the `self.process_wikidata` call. Would be nice to have it consistent.

plugins/wikidata/__init__.py
```diff
@@ -93,229 +97,233 @@ def __init__(self):
...
+        if 'work' in response.metadata[0].children and self.use_work_genres:
+            if 'relation_list' in response.metadata[0].work[0].children:
+                for relation in response.metadata[0].work[0].relation_list[0].relation:
+                    if relation.type == 'wikidata' and 'target' in relation.children:
+                        found = True
+                        wikidata_url = relation.target[0].text
+                        log.info('WIKIDATA: wikidata url found for WORK: %s', wikidata_url)
+                        self.process_wikidata(Wikidata.WORK, wikidata_url, item_id)
```
phw, line 185: What do you think about moving the code to find the wikidata url into a helper function? Something like:

```python
def get_wikidata_url(node):
    if 'relation_list' in node.children:
        for relation in response.metadata[0].work[0].relation_list[0].relation:
            if relation.type == 'wikidata' and 'target' in relation.children:
                return relation.target[0].text
    return None
```

The use it like:

```python
wikidata_url = response.metadata[0].work[0]
if wikidata_url:
    found = True
    log.info('WIKIDATA: wikidata url found for WORK: %s', wikidata_url)
    self.process_wikidata(Wikidata.WORK, wikidata_url, item_id)
```

plugins/wikidata/__init__.py
```diff
@@ -6,9 +6,14 @@
...
-PLUGIN_DESCRIPTION = 'Query wikidata to get genre tags'
-PLUGIN_VERSION = '1.4.5'
+PLUGIN_AUTHOR = 'Daniel Sobey, Sambhav Kothari, Sophist-UK'
+PLUGIN_DESCRIPTION = '''
+Query wikidata to get genre tags.
+<br/><br/>
+Since genres can be gathered from multiple sources and the genre tag may be overwritten,
+we also set a hidden variable _genre_wikidata to the same values for use in scripts.
+'''
+PLUGIN_VERSION = '1.4.6'
```
phw, line 16: I'd suggest setting this to 1.5, after all this PR also adds a small feature by providing the `~wikidata_genre` variable.

phw (changes requested):
Thanks for this detailed refactoring. I'm finally catching up on reviewing plugins. Not much to say really about this, looks good and simplifies the code. Only some small comments.

> As an aside, I am unsure how it has a codacy issue of a method being over complex when the refactoring reduced the complexity of the code substantially. However, if needed I can refactor it a little more to reduce the complexity further.

Yes, codacy is just being silly I think. Let's ignore it.

PR #373, opened 2024-02-17
add Add to Collection plugin

dvirtz:
This plugin adds any saved release to one of the user collections


plugins/add_to_collection/override_module.py
```diff
@@ -0,0 +1,11 @@
...
+from typing import Generator
+
+
+@contextmanager
+def override_module(obj: object) -> Generator[None, None, None]:
+    # picard expects hooks to be defined at module level
+    module = obj.__module__
+    obj.__module__ = ".".join(module.split(".")[:-1])
+    yield
+    obj.__module__ = module
```
phw, line 11: I guess it's fine and I understand why you wanted to split this appart as you did. But it also feels a bit like over engineering for this plugin, to be honest.

Anyway, I'm fine keeping this.

phw (approved):
Thanks a lot for this. This is definitely useful and it is looking good. LGTM

Overall this is some functionality that eventually should end up in Picard at some point. But at this moment I think having it as a plugin and letting that mature first is a good move.

For a future idea I think it would be cool if one could define some conditions to have saved releases end up in different collections based on some criteria. E.g. I keep my physical releases in a separate collection from the downloads.

zas (approved):
LGTM

Clearly this is something Picard should handle natively, including adding one release to multiple collections.

plugins/add_to_collection/override_module.py
dvirtz: I hope this limitation can be lift off in [V3](https://github.com/phw/picard/blob/plugins-v3/PLUGINS.md)

phw: Yes, I will look into this when I get back to the reworked plugin handling

PR #374, opened 2024-02-22
PICARD-2743: Add support for custom post tagging actions

twodoorcoupe:
This plugins tries to solve [PICARD-2743](https://tickets.metabrainz.org/browse/PICARD-2743).
It adds the ability to specify "actions" that run with a context menu click. Basically, it can run external programs.
Besides what was initially proposed, I took inspiration from what other taggers offer, like [mp3tag tools](https://docs.mp3tag.de/customization/options/tools/).

The commands can be set in the plugin's option page, the UI looks like this:

ui

The commands run in the order they appear in the table. It lets you choose:
- Whether to wait for the current command to exit or run the next one immediately.
- Whether to refresh the tags after the process exits. 
- Whether to execute each command once per album or once per track highlighted.

Also, it lets you specify variables to use in the command. These are taken either from the metadata like `%title%` or `%artist%`, or from extra "special" variables like `%filepath%` or `%is_complete%`.

I uploaded a user guide [here](https://github.com/twodoorcoupe/picard-plugins/blob/user_guides/user_guides/post_tagging_actions/guide.md) where everything is hopefully explained more thoroughly, it includes a list of all the special variables as well. There's a link to it both in the options page and in the plugin's metadata.


plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+This plugin lets you set up actions that run with a context menu click. 
+An action consists in a command line executed for each album or each track along
+with a few options to tweak the behaviour. 
+This can be used to run external programs and pass some variables to it. 
+"""
+PLUGIN_VERSION = "0.1"
+PLUGIN_API_VERSIONS = ["2.10", "2.11"]
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
+PLUGIN_USER_GUIDE_URL = "https://github.com/twodoorcoupe/picard-plugins/tree/user_guides/user_guides/post_tagging_actions/guide.md"
```
Sophist-UK: I note that the documentation is **NOT** included in this PR - and IMO it is essential that it should be included in order to ensure that the availability and versioning of the documentation matches the plugin.

Also you are creating a new top-level sub-directory and have chosen `user_guides`. I would personally prefer to see this being `docs` as this seems to me to be a bit of a defacto standard - however others may think differently.

When you start to include this in the PR I think this URL will need to assume that this PR has been merged.

plugins/post_tagging_actions/options_post_tagging_actions.py
```diff
@@ -0,0 +1,175 @@
...
+
+# Form implementation generated from reading ui file 'plugins/post_tagging_actions/options_post_tagging_actions.ui'
+#
+# Created by: PyQt5 UI code generator 5.15.10
+#
+# WARNING: Any manual changes made to this file will be lost when pyuic5 is
+# run again.  Do not edit this file unless you know what you are doing.
+
+
+from PyQt5 import QtCore, QtGui, QtWidgets
```
Sophist-UK: Codacy error - unused QtGui reference.

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+            self._create_action(priority, commands, album, options)
+
+    def load_actions(self):
+        """Loads the information from the options and saves it in the action options list.
+
+        This gets called when the plugin is loaded or when the user saves the options.
+        """
+        self.action_options = []
+        loaded_options = zip(*[config.setting[name] for name in OPTIONS])
+        for options in loaded_options:
```
Sophist-UK: Why not iterate on `for name in OPTIONS` rather than create a new object?

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+    "get_num_unmatched_files",
+    "get_num_total_files",
+    "get_num_unsaved_files",
+    "is_complete",
+    "is_modified"
+}
+
+# Settings.
+CANCEL = "pta_cancel"
+OPTIONS = ["pta_command", "pta_wait_for_exit", "pta_execute_for_tracks", "pta_refresh_tags"]
```
Sophist-UK: Would this be better as an immutable `set` (i.e. curly brackets) like the ALBUM_SPECIAL_VARIABLES above?

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+from queue import PriorityQueue
+from threading import Thread
+from concurrent import futures
+from os import path
+import re
+import shlex
+import subprocess
+
+# Additional special variables.
+TRACK_SPECIAL_VARIABLES = {
```
Sophist-UK, line 56: I like this use of lambdas, however if you can find a way to annotate and avoid the Codacy warnings that would be good.

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+        config.setting[CANCEL] = self.ui.cancel.isChecked()
+        action_loader.load_actions()
+
+
+action_loader = ActionLoader()
+action_runner = ActionRunner()
+register_album_action(ExecuteAlbumActions())
+register_track_action(ExecuteTrackActions())
+register_options_page(PostTaggingActionsOptions)
+QObject.tagger.register_cleanup(action_runner.stop)
```
Sophist-UK: I don't know what this `register_cleanup` line is for or why it is an initialisation statement.

Can you please annotate this with a comment explaining it?

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+        futures.wait(future_objects, return_when = futures.ALL_COMPLETED)
+        for file in album.iterfiles():
+            file.set_pending()
+            file.load(lambda file: None)
+        thread.to_main(album.load, priority = True, refresh = True)
+
+    def _run_process(self, command):
+        """Runs the process and waits for it to finish.
+        """
+        process = subprocess.Popen(command, text = True, stdout = subprocess.PIPE, stderr = subprocess.PIPE)
```
Sophist-UK: The command here is user provided, and is explicitly visible in the UI, so I see no issue of untrusted input here.

If you are able to annotate this to avoid this Codacy warning that would be helpful.

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+    def load_actions(self):
+        """Loads the information from the options and saves it in the action options list.
+
+        This gets called when the plugin is loaded or when the user saves the options.
+        """
+        self.action_options = []
+        loaded_options = zip(*[config.setting[name] for name in OPTIONS])
+        for options in loaded_options:
+            command = options[0]
+            other_options = [eval(option) for option in options[1:]]
```
Sophist-UK: The options here seem to be booleans, so avoiding use of `eval` and the consequent Codacy waarning shouldn't be difficult.

But if use of `eval` cannot be avoided, if you are able to annotate this to avoid this Codacy warning that would be helpful.

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+        self.action_options = []
+        self.action_counter = 0
+        self.load_actions()
+
+    def _create_options(self, command, *other_options):
+        """Finds the variables in the command and adds the options to the action options list.
+        """
+        variables = [variable[1:-1] for variable in re.findall(r'%.*?%', command)]
+        variables = [parser.normalize_tagname(variable) for variable in variables]
+        command = re.sub(r'%.*?%', '{}', command)
```
Sophist-UK: Would it make more sense to create a pre-compiled regex as a pre-initialised object?

plugins/post_tagging_actions/__init__.py
Sophist-UK: On this documentation page, I am not sure that the COV integration tool is a good example for this plugin even though it is a valid example - because any Cover Art integration would ideally be done by creating a Cover Art plugin.

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+    def __init__(self):
+        self.action_options = []
+        self.action_counter = 0
+        self.load_actions()
+
+    def _create_options(self, command, *other_options):
+        """Finds the variables in the command and adds the options to the action options list.
+        """
+        variables = [variable[1:-1] for variable in re.findall(r'%.*?%', command)]
+        variables = [parser.normalize_tagname(variable) for variable in variables]
```
Sophist-UK: ` variables = [parser.normalize_tagname(variable[1:-1]) for variable in re.findall(r'%.*?%', command)]`?

plugins/post_tagging_actions/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+    """Runs actions in the execution queue.
+
+    Attributes:
+        action_thread_pool (ThreadPoolExecutor): Pool used to run processes with the subprocess module.
+        refresh_tags_pool (ThreadPoolExecutor): Pool used to reload tags from files and refresh albums.
+        worker (Thread): Worker thread that picks actions from the execution queue.
+    """
+
+    def __init__(self):
+        self.action_thread_pool = futures.ThreadPoolExecutor()
```
Sophist-UK: According to the [Python `futures` documentation](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor):
> If `max_workers` is None or not given, it will default to the number of processors on the machine, multiplied by 5

So this is going to be the number of subprocesses that are run - and depending on the commands the user is choosing to run and the specifications of their PC this may be fine or it may cause major issues.

I think there needs to be a UI option with a sensible default to set the `max-workers` for this thread-pool.

Sophist-UK (changes requested):
This appears to be a potentially useful plugin - so congrats on creating it.

In addition to the comments on existing code, I would also like to propose the the code increments and decrements the Picard UI `Pending Requests` number in the status bar.

plugins/post_tagging_actions/__init__.py
twodoorcoupe: That makes sense. I also saw other plugins like Additional Artist Details and Format Performer Tags have their docs included in the plugin's folder, so I did the same rather than creating a new top-level sub-directory.

plugins/post_tagging_actions/__init__.py
twodoorcoupe: The order of the options in this list is the same as the order of the columns of the table in the options page. I use the order when saving or loading the options. This way I can just iterate over the columns without having to explicitly save or load the values in each column.

plugins/post_tagging_actions/__init__.py
twodoorcoupe: I think I used a confusing name for `options`  and `loaded_options` here. I  iterate over the options for each "action" rather than each option.  Each `options` is a tuple made of all the options for that action (command, wait_for_exit, etc...), so I need to use the zip function to do this. I changed the name of both `options`  and `loaded_options` to be hopefully less confusing.

twodoorcoupe:
Thank you @Sophist-UK! I made the number of worker threads an option and made it so that the number of pending requests is displayed in the status bar like you asked.

plugins/post_tagging_actions/__init__.py
Sophist-UK: If `set` doesn't preserve order, this is a sound reason.

plugins/post_tagging_actions/__init__.py
```diff
@@ -201,12 +207,17 @@ def _execute(self):
         """
         while True:
             priority_action = action_queue.get()
+            QObject.tagger.window.set_statusbar_message(
+                N_("Post Tagging Actions: number of pending requests is %(pending_requests)d"),
+                {"pending_requests": action_queue.qsize()},
+                timeout = 3000
+            )
+
```
Sophist-UK: This wasn't what I had in mind (which was to increment/decrement Picard's existing outstanding requests value), and I have no idea whether it will cause any issues (threading, excessive overhead etc.).

plugins/post_tagging_actions/__init__.py
```diff
@@ -259,7 +270,11 @@ class PostTaggingActionsOptions(OptionsPage):
     PARENT = "plugins"
 
     action_options = [config.ListOption("setting", name, []) for name in OPTIONS]
-    options = [config.BoolOption("setting", CANCEL, True), *action_options]
+    options = [
+        config.BoolOption("setting", CANCEL, True),
+        config.IntOption("setting", MAX_WORKERS, 4),
```
Sophist-UK: Why 4?

Shouldn't this at least be based on the number of cores/hyperthreads that the CPU has?

plugins/post_tagging_actions/__init__.py
Sophist-UK: Can this be moved into the initialisation of the `ActionRunner` object?

Sophist-UK (changes requested):
Thank you for taking the time to make all these changes.

Only a few follow-on comments.

I haven't tested this - nor am I likely to have time to do so. And in any case I can't approve this PR. So perhaps @zas or @outsidecontext could test this?

plugins/post_tagging_actions/__init__.py
twodoorcoupe: Ah, sorry for misinterpreting. I think that value is used for webservice requests. Would it make sense to use it also for pending actions? For example, if I have some action running in the background and I load a new album, then the number of pending requests would include both?

Sophist-UK: Its not a problem - you got the overall intent (that the user needs some sort of indicator about progress).

Yes - that value is intended for web-service requests. Maybe we need another similar indicator for local requests that can also be used for `acoustid` scans or volume levelling scans or re-encoding. Or perhaps you could look see what the which indicator is used by the native acoustid scan code and use the same indicator?

twodoorcoupe: The `acoustid` scans use the `Pending Files` indicator right next to the `Pending Requests` one. I could make it so that the files being "used" by an action are put in the pending state, meaning the counter is updated and the tracks are greyed out until the action finishes. This way, however, `Pending Files` would indicate the number of files being "used" by an action, rather than the number of actions that still need to be completed. I don't know if this is what you meant. 
In the meantime, I made the other changes you suggested.

plugins/post_tagging_actions/__init__.py
twodoorcoupe: I made it so that the default value is the same as the one used by `futures` module.

plugins/post_tagging_actions/__init__.py
Sophist-UK: I cannot be certain because I haven't tried to use the plugin, but this description seems to make sense.

plugins/post_tagging_actions/__init__.py
Sophist-UK: That sounds as good a default as we can get given that the tasks for a file could be heavy (reencoding the file) or light (doing some sort of web lookup), and the number of parallel tasks that can be supported would be very different for these two situations.

Sophist-UK:
@twodoorcoupe Giorgio - We are getting pretty close as far as I am concerned. Keep going....

plugins/post_tagging_actions/__init__.py
phw: But it could be a tuple :)

phw (commented):
Thanks a lot for this. This definitely looks good and is something I wanted to have for a long time myself.

I'll give this a local test run, but from reading the PR this already looks really good.

And also thanks a lot @Sophist-UK for all the review work here. Much appreciated.

twodoorcoupe:
@Sophist-UK, I tried using the `Pending files` indicator in the status bar, but I ran into a few issues with concurrency. For example: refreshing the album, saving the files, or running `acoustid` scans would all change the files' states.

Therefore, I opted for adding another indicator in the status bar to keep track of the pending actions. It look like this:

pending_actions

It is only visible when there are actions pending, otherwise it stays hidden. The icon is the same as the one used for `Plugins` and `Run scripts` in the context menu.

I think this solution is better since, besides avoiding issues with other functions, it also makes the meaning of the indicator more clear.

twodoorcoupe:
Thanks @phw! I saw your comment a minute too late. I changed it now.

Sophist-UK:
> Therefore, I opted for adding another indicator in the status bar to keep track of the pending actions. 
It is only visible when there are actions pending, otherwise it stays hidden. The icon is the same as the one used for `Plugins` and `Run scripts` in the context menu.
> 
> I think this solution is better since, besides avoiding issues with other functions, it also makes the meaning of the indicator more clear.

I think this is a great solution.

Does this new status icon have a tooltip? If not can you add one please.

twodoorcoupe:
> Does this new status icon have a tooltip? If not can you add one please.

Yes, it says "Pending actions", keeping the "style" of the other icons' tooltips.

Sophist-UK:
Great! As far as my desk review goes, this seems like it is good to go.

phw (changes requested):
@twodoorcoupe I actually used this now and it is great. This will be useful for me, I already integrated it into my workflow. There are a couple of things that can be improved for the UI, e.g. editing existing actions. But all that can be done in a future update, I'd rather see that we get this merged first.

But also encountered some actual issues:

1. I could not see the status bar icon show up. I haven't digged deeper, yet,and I don't see any error message,  but something is no working for me here.
2. When closing Picard without an action currently running there is an exception:


``` 
  File "/usr/bin/picard", line 5, in <module>
    main('/usr/share/locale', False)
  File "/usr/lib/picard/picard/tagger.py", line 1583, in main
    exit_code = tagger.run()
                ^^^^^^^^^^^^
  File "/usr/lib/picard/picard/tagger.py", line 751, in run
    self.exit()
  File "/usr/lib/picard/picard/tagger.py", line 738, in exit
    self.run_cleanup()
  File "/usr/lib/picard/picard/tagger.py", line 635, in run_cleanup
    f()
  File "/home/phw/.config/MusicBrainz/Picard/plugins/post_tagging_actions/__init__.py", line 284, in stop
    self.update_widget.join()
  File "/usr/lib/python3.11/threading.py", line 1114, in join
    raise RuntimeError("cannot join thread before it is started")
RuntimeError: cannot join thread before it is started
```

Actually this might be related to the first issue (widget not being instantiated)

plugins/post_tagging_actions/docs/guide.md
```diff
@@ -0,0 +1,52 @@
+# Post Tagging Actions
+This plugin lets you set up actions that run with a context menu click.
+An action consists in a command line executed for each album or each track along with a few options to tweak the behaviour. This can be used to run external programs and pass some variables to it. Environment variables do not work.
+
+To run the actions,
+- First move the albums or tracks you are interested in to the right hand pane.
+- Then highlight all the items you want the actions to run for.
+- Right click, go to plugins, then click "Run Actions for highlighted albums/tracks".
+### Adding an action
+In the options page, you will find "Post Tagging Actions" under "Plguins". You will be greeted by this:
```
zas:
```suggestion
In the options page, you will find "Post Tagging Actions" under "Plugins". You will be greeted by this:
```

zas (changes requested):

twodoorcoupe:
Thanks for testing it @phw. 
I've been trying to reproduce the issue you are having, but I'm out of luck. I've tried with both Picard 2.10 and 2.11 on both windows and linux, and for me it works out of the box.
So I wanted to ask you, does the icon never show up at all? Even after you restart Picard? I ask this because the widget is instantiated differently whether the plugin is loaded before or after the main window is created.

phw:
@twodoorcoupe Found the issue, not yet sure how to fix. This happens if the plugin is already installed, like it was for me, since I checked out your branch and placed a symlink in `~/.config/MusicBrainz/Picard/plugins`. On the run where you load Picard with the plugin disabled but then enable and use it the UI widget does not get set up properly.

That's because registering all the hooks happens when Picard is started and loads the plugins, regardless on whether the plugin is enabled or not. That means the code where it decided whether to initialize the widget immediately or use the ` register_ui_init`  hook (https://github.com/metabrainz/picard-plugins/pull/374/files#diff-dff6ecca533b5cccfc28ffd3c93229fc405c7f00c781d451b1c7766dfa8f6713R194-R198) it uses the UI init hook. But because the plugin is disabled, it does not actually run. And this hook only runs on start, not later if a plugin actually gets enabled (which essentially implies the use of `register_ui_init`  always requires a restart).

The `register_ui_init`  is kind of a kludge really, I'm not too happy about this.

But for now I think the best solution would be to live with the fact that showing the status widget requires a restart and just make sure that the call `tagger.register_cleanup(self.stop)` only happens after `self.update_widget.start()` got called in `_create_widget`. That will ensure there is no crash on exit.

It's all really a limitation of the plugin system currently, something we want to improve with Picard 3 by having plugins provide explicit hooks for lifecycle events like "enable" and not running anything if a plugin is disabled (more on that soon).

twodoorcoupe:
Thank you so much for finding the issue @phw! I guess I have the habit of always enabling plugins as soon as I install them, so I would have never figured it out without your help.

phw (approved):
Thanks again. From my side this is good to go. Code looks good and I have used it for my recent tagging without a hitch.

zas (approved):
LGTM

PR #375, opened 2024-03-18
Added losslessfuncs plugin

phw:
This adds two helper functions `$is_lossless()`  and `$is_lossy()` to help tagging files depending on whether they are in a lossy or lossless format.

The detection is not actually based on the audio codec but uses formats and some heuristics. While not perfect and does the right thing in the majority of cases and it is an essential part of my file naming script for years. The limitation would be a reason for me not to include it in Picard itself, but as a plugin I think it is fine.

The plugin is already in active use by others as well, see the discussions in the forums:

https://community.metabrainz.org/t/categorizing-lossless-music/258530/
https://community.metabrainz.org/t/is-lossless-plugin-no-longer-working-with-m4a/469429

### DIFF
--- a/plugins/losslessfuncs/__init__.py
+++ b/plugins/losslessfuncs/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2014, 2017, 2021 Philipp Wolfer
+# Copyright (C) 2014, 2017, 2021, 2023-2024 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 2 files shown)


plugins/losslessfuncs/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+PLUGIN_NAME = 'Tagger script functions $is_lossless() and $is_lossy()'
+PLUGIN_AUTHOR = 'Philipp Wolfer'
+PLUGIN_DESCRIPTION = 'Tagger script functions to detect if a file is lossless or lossy'
+PLUGIN_VERSION = "0.3"
+PLUGIN_API_VERSIONS = ["2.0", "2.1", "2.2", "2.3", "2.4", "2.5", "2.6", "2.7", "2.8", "2.9", "2.10", "2.11"]
+PLUGIN_LICENSE = "GPL-2.0"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
+
+
+LOSSLESS_EXTENSIONS = ['flac', 'oggflac', 'ape', 'ofr', 'tak', 'wv', 'tta',
```
zas: Maybe use a set instead of a list here,

zas (approved):
LGTM

PR #377, opened 2024-05-11
Use iso-3166-2-codes if iso-3166-1-codes is not available

rdswift:
In reviewing the discussion at https://community.metabrainz.org/t/possibility-to-tag-by-artist-area-artist-country-of-origin/596808/19 it was discovered that sometimes the country code was returned as iso-3166-2-codes rather than iso-3166-1-codes, and these were not being processed by the plugin.  This PR adds the code to process these iso-3166-2-codes if there are no iso-3166-1-codes provided by the api.

### DIFF
--- a/plugins/additional_artists_details/__init__.py
+++ b/plugins/additional_artists_details/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Additional Artists Details
 """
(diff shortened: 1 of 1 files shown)


phw (approved):
That's interesting, I didn't know this could happen.

Changes look good, let's merge it

PR #379, opened 2024-06-25
deezerart: improve error handling

phw:
Do not crash if the response does not include a "data" object or is an error

### DIFF
--- a/plugins/deezerart/deezer/client.py
+++ b/plugins/deezerart/deezer/client.py
@@ -46,7 +46,15 @@ class Client:
             except json.JSONDecodeError:
                 callback([], error)
             else:
-                callback([obj.parse_json(dct) for dct in parsed_doc['data']], error)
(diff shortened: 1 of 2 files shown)


plugins/deezerart/deezer/client.py
```diff
@@ -43,10 +43,19 @@ def advanced_search(self, options: SearchOptions, callback: SearchCallback[obj.A
         def handler(document: QByteArray, _: QNetworkReply, error: Optional[QNetworkReply.NetworkError]):
             try:
                 parsed_doc = json.loads(str(document, 'utf-8'))
+                print(parsed_doc)
```
zas: Debug leftover?

phw: oops, fixed

zas (approved):
LGTM

PR #381, opened 2024-09-08
replaygain2: support writing both standard and R128 tags to Opus files

phw:
Currently the replaygain2 plugin supports writing either the standard ReplayGain tags or the R128 tags for Opus files, but not both.

I discovered that Navidrome only handles the standard tags, but I also need the the R128 tags for another tool. Having both gives compatibility with both.

See also https://github.com/complexlogic/rsgain/issues/119 , but the plugin can do this independently has it is Picard writing the final tags anyway.


zas (approved):
LGTM

PR #383, opened 2024-10-06
replaygain2: handle subclasses of supported file formats

phw:
This fixes file format detection when using e.g. the compatible_TXXX plugin, see PICARD-2986


zas (approved):
LGTM

PR #386, opened 2024-10-31
Added polish diacritics in non_ascii_equivalents.py

finem4n:
As in title I've added some polish diacritics and also extended filter tags list


phw (changes requested):
Thanks. This is somewhat similar to #387

Extending the tag list that way makes sense to me. The new tags seem to be in line with how the plugin originally was conceived. It probably would be even better to have some configuration for this, but in absence of this I think the extensions as presented here is useful.

What also applies here is my comment on #387 about using Picard's `picard.util.textencoding.unaccent` function (please see my detailed comment there). This would allow to get rid of the explicit mapping of most accented characters. As I see it the first mapping section can be completely removed then, except for the two letters "Ł" and "ł" (which could be placed under "Misc letters" then.

In the future this will then avoid the need to add additional accented characters, likely there are a few that we still miss.

finem4n:
Hi. I implemented `picard.util.textencoding.unaccent`. I did some testing and at first glance it handled more than letters, e.g. ≠ and the L shaped ones, but results were disappointing:  
≠ changed to =  
Ls changed back to 「 instead of |-  
So I left them in CHAR_TABLE as they were before.  
As per your suggestion in #387  I renamed function `ascii` to `to_ascii`. I couldn't find a better name.
I also bumped version and appended my name in authors section, if you don't mind.  
If I have more free time, I'd be willing to step in and add scripting functionality.

Sophist-UK:
In # 387 Echelon666 has said:
> You have to finish it yourself.

@finem4n Konrad Would you be willing to include the extra characters from the other PR taking into account @phw Philipp's comments?

finem4n:
@Sophist-UK Yeah, sure.  
According to [wiki](https://en.wikipedia.org/wiki/Thorn_(letter)) one of the transliterations of `þ` (thorn) is `th` not `p`, so I've changed that. I also went with ascii representations of `♥ → ・` instead of minus.

phw (approved):
Thanks for the update, this looks good to me.

plugins/non_ascii_equivalents/non_ascii_equivalents.py
```diff
@@ -29,38 +30,24 @@
...
-    "Ý": "Y",  "Ỳ": "Y",  "Ÿ": "Y",  "Ŷ": "Y",
-    "á": "a",  "à": "a",  "ä": "a",  "â": "a",
-    "é": "e",  "è": "e",  "ë": "e",  "ê": "e",
-    "í": "i",  "ì": "i",  "ï": "i",  "î": "i",
-    "ó": "o",  "ò": "o",  "ö": "o",  "ô": "o",
-    "ú": "u",  "ù": "u",  "ü": "u",  "û": "u",
-    "ý": "y",  "ỳ": "y",  "ÿ": "y",  "ŷ": "y",
-
     # Misc Letters
-    "Å": "AA",
```
zas, line 50: If I understand correctly this is now handled by `unaccent()` but it's a change here, since `unaccent("Å")` returns `A` not `AA`.
Is that expected?

phw: As of https://en.wikipedia.org/wiki/%C3%85#International_transcription the typical replacement for "å" is "aa", as "å" represents a long pronounced "a" sound.

At least the plugin originally made the choice to replace the character like this, hence I suggested to keep this functionality the same (which happens by applying the replacement first, then doing the unaccent).

On the flip side replacing uppercase "Å" with "AA" is not fully fine. In regular writing with upper and lowercase letters you would replace it with "Aa" instead. But this then looks wrong if the word it is part of is otherwise spelled in all caps. Also it seems to be language dependent, Wikipedia states that this replacement is not used in Swedish. But I'm no expert on any language using Å, so I can't say for sure.

Also there is another similar case with Umlauts ä, ö, ü. Currently they get just unaccented to a, o, u. But in German you would usually replace them with ae, oe or ue instead when constrained to ASCII. But this is I think not universally true for other languages using those characters.

So in the end I think trying to preserve the existing functionality for Å makes sense for now, unless someone with better knowledge of affected languages tells otherwise.

zas: @phw: So, revert changes for those 2 lines, right?

I have no strong opinion about it, just wanted to be sure this change was noticed.

phw: Yes, you are right. I missed that these get removed.

@finem4n We should keep the existing conversion of Å -> AA and å -> aa

finem4n: Sure.
Pushing changes right now

ghost:
What about this:

"č": "c",
"š": "s",
"ș": "s",

unaccent performs this?

zas:
> What about this:
> 
> "č": "c", "š": "s", "ș": "s",
> 
> unaccent performs this?

Yes.

```python
>>> unaccent("čšș")
'css'
```

phw (approved):
Thanks a lot

ghost:
I found two similar signs in my collection: 

µ - https://symbl.cc/en/00B5/    (in plugin)

μ - https://symbl.cc/en/03BC/   (song Bjork - Hunter (μ-Ziq remix) )

finem4n:
@Echelon666 as @phw said, in current state of this plugin we will almost always miss some characters. That's why he proposed to add scripting functionality, so users'll be able to define missing characters or change transliterations themselves. I'd be willing to add scripting functionality, but as for now I should focus on defending my thesis.  
If you really need to deal with these two characters now, then I believe that you already know how to append them in `CHAR_TABLE` and install plugin from local .py file.

ghost:
@finem4n 

Sure. ;)

I thought this was important. ;)

PR #388, opened 2025-01-07
create plugin 'remove_[data_track]s' (formerly skip_data_tracks_and_fix_track_numbers)

crnkv:
I wrote a plugin that is intended to present the track list in line with online digital music platforms. It automatically skips data tracks and silent tracks once the album metadata is loaded / retrieved.


plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+PLUGIN_LICENSE_URL = 'https://www.gnu.org/licenses/gpl-2.0.html'
+
+from picard import log
+from picard.metadata import (register_album_metadata_processor, register_track_metadata_processor)
+from picard.plugin import PluginPriority
+
+
+def remove_datatracks_from_release(album, metadata, release):
+    try:
+        log.info("{0}: Infomation: {1}".format(PLUGIN_NAME, "Removing data tracks / silence tracks from release "+release['title'],))
```
Sophist-UK: Given that this is going to be a public plugin, I think this should only be shown IF the plugin is actually going to remove tracks,

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+                   or track['recording']['artist-credit'][0]['artist']['name'] == '[data]' \
+                   or track['title'] == '[silence]' \
+                   or track['recording']['title'] == '[silence]' :
+                    # The track['number'] field usually conforms with the %_musicbrainz_tracknumber% metadata, which can be vinyl numbering (A1, A2…). This is not to be used.
+                    # Assuming that the track['position'] field coincides with the usual %tracknumber% metadata, which are integers starting from 1, even when a pregap track exists in the medium (which is in the disc['pregap'] field with a disc['pregap']['position'] field equal to 0).
+                    datatrack_positions.append(track['position'])
+                    del disc['tracks'][i]
+            disc['track-count'] = len(disc['tracks'])
+            if len(datatrack_positions) != 0:
+                log.info("{0}: Infomation: {1}".format(PLUGIN_NAME, "Removed data tracks / silence tracks of positions: "+str(sorted(datatrack_positions)),))
```
Sophist-UK: 1. Personally I would like this to say "Removed X data / silence tracks, original positions: 2, 3, 5, 8.

2. `log.info("{0}: Information: Removed {1} data / silence tracks, original positions: {2}", [PLUGIN_NAME, len(datatrack_positions), datatrack_positions])`. This is better style, and better suited for future language translations.

3. Perhaps better shown after the tracks have been renumbered.

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+<ul>
+<li>Track #00: [pregap] audio track</li>
+<li>Track #01: audio track</li>
+<li>Track #02: audio track</li>
+<li>Track #03: [crowd noise] yet still an audio track</li>
+<li>Track #04: hidden audio track</li>
+<li> -- totaltracks: 5 -- </li>
+</ul>
+This is automatic once the album is loaded. Users wouldn't be aware of the existence of data tracks and silence tracks.
+<br /><br />
```
Sophist-UK: Can you please include the MBID of a real release that exhibits as many of these examples in real-life as possible (so that we can have an example to use in testing)?

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+                track = disc['tracks'][i]
+                # This is a data track if:
+                # Track Title / Recording Title is [data track], or Track Artist / Recording Artist is [data]
+                # Bonus: also skip when Track Title / Recording Title is [silence]
+                if track['title'] == '[data track]' \
+                   or track['recording']['title'] == '[data track]' \
+                   or track['artist-credit'][0]['artist']['name'] == '[data]' \
+                   or track['recording']['artist-credit'][0]['artist']['name'] == '[data]' \
+                   or track['title'] == '[silence]' \
+                   or track['recording']['title'] == '[silence]' :
```
Sophist-UK: I think you should also include:
```python
or track['~datatrack']
```
in this list of triggers and avoid the need for the Track Metadata Processor to double check for this below.

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
```diff
@@ -0,0 +1,104 @@
...
+                log.info("{0}: Infomation: {1}".format(PLUGIN_NAME, "Removed data tracks / silence tracks of positions: "+str(sorted(datatrack_positions)),))
+            for track in disc['tracks']:
+                position = track['position']
+                number_to_skip = len([p for p in datatrack_positions if p < position])
+                track['position'] = position - number_to_skip
+            # Assuming that after the remove_datatracks_from_release function call, MusicBrainz Picard will infer / calculate the metadata %tracknumber% %totaltracks% %_absolutetracknumber% %_totalalbumtracks% from track['position'] fields of the modified release Object, that nothing else is needed to care about.
+    except Exception as ex:
+        log.error("{0}: Error: {1}".format(PLUGIN_NAME, ex,))
+
+def doublecheck_metadata(album, metadata, track, release):
```
Sophist-UK: See above.

Sophist-UK (commented):
This looks useful and generally well written.

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
crnkv: done

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
crnkv: done

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
crnkv: done

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
crnkv: This is not possible because the `%_datatrack%` metadata (MusicBrainz variable) is not part of the album metadata or track metadata retrieved from the MusicBrainz JSON web service.
It seems that none of the arguments passed to the album_metadata_processor contains this `_datatrack` or `~datatrack` boolean flag, while only the arguments passed to the track_metadata_processor gets populated with such flag(s) by MusicBrainz Picard through some logical inferences.
So, logically speaking, MusicBrainz Picard couldn't see no more information than what's already in the album_metadata_processor. There shouldn't be cases where the album JSON says no datatrack yet Picard still deduce the `%_datatrack%` flag. This really is just a double check.
If it happens, something very wierd must be happening, and I want to stay away from it.
Plus, if I continue to remove tracks within the track_metadata_processor based on the `~datatrack` flag (yes it is possible), a lot more work has to be done: `tracknumber`, `totaltracks`, `~absolutetracknumber`, `~musicbrainz_tracknumber`, `~totalalbumtracks` should be all updated asyncly, and these tracks are mutually dependent, that's quite a nightmare...

Sophist-UK:
Sorry, one more comment which is the length of the filename - can we have a shorter name please?

Also I feel the plugin name could be snappier - yes, it does what it says on the can, but how about "Remove [data] tracks" instead?

plugins/skip_data_tracks_and_fix_track_numbers/skip_data_tracks_and_fix_track_numbers.py
Sophist-UK: Yes - I think it is as per [these lines of code in Picard's album.py](https://github.com/metabrainz/picard/blob/master/picard/album.py#L447-L448).

And I am NOT suggesting that you do anything in the Track Metadata Processor - I am suggesting that you eliminate it entirely.

crnkv: I spent some time looking into the codes. It seems that the lines you pointed out are for another scenario (that I missed out in my script), which is when the album JSON contains a `json['media'][0]['data-tracks']` node (which is not seen in a real album example)
The `~datatrack` flags per track are actually set in [these lines](https://github.com/metabrainz/picard/blob/master/picard/track.py#L315). And, yes, as long as the `if` condition in those lines won't change in the future, my `if` conditions here would be more than sufficient - I'll remove the double check

crnkv:
> Sorry, one more comment which is the length of the filename - can we have a shorter name please?
> 
> Also I feel the plugin name could be snappier - yes, it does what it says on the can, but how about "Remove [data] tracks" instead?

all suggestions accepted. any other problems? can we proceed?

PR #390, opened 2025-03-08
Update test.yml

rdswift:
Remove python 3.7 and 3.8 test platforms which are no longer supported.


phw (approved):
Yes, we need to move on here as well.

PR #391, opened 2025-05-11
albumartist_website: handle URLs that have ended

inco-cc:
This modifies the `albumartist_website` plugin to prevent it from including URLs that have their relationship marked as “ended” (i.e. no longer active / applicable to the artist)

### DIFF
--- a/plugins/albumartist_website/albumartist_website.py
+++ b/plugins/albumartist_website/albumartist_website.py
@@ -4,7 +4,7 @@ PLUGIN_NAME = 'Album Artist Website'
 PLUGIN_AUTHOR = 'Sophist, Sambhav Kothari, Philipp Wolfer'
 PLUGIN_DESCRIPTION = '''Add's the album artist(s) Official Homepage(s)
 (if they are defined in the MusicBrainz database).'''
-PLUGIN_VERSION = '1.1'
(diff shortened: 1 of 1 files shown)


plugins/albumartist_website/albumartist_website.py
```diff
@@ -4,7 +4,7 @@
 PLUGIN_AUTHOR = 'Sophist, Sambhav Kothari, Philipp Wolfer'
 PLUGIN_DESCRIPTION = '''Add's the album artist(s) Official Homepage(s)
 (if they are defined in the MusicBrainz database).'''
-PLUGIN_VERSION = '1.1'
+PLUGIN_VERSION = '1.1.1'
```
Sophist-UK: 1.2?

Sophist-UK (approved):
As the original author, I support this change.

plugins/albumartist_website/albumartist_website.py
phw: I'd also bump it to 1.2, as this is a bit more than a fix.

phw (changes requested):
Thanks. As suggested I would also be for bumping the version to 1.2. Otherwise this looks good to merge to me.

phw (approved):
Thanks for the update

Issue #344, opened 2023-02-05
Replaygain 2.0 only usable in "releases" pane


tacuna:
As per https://community.metabrainz.org/t/how-do-i-run-the-replay-gain-2-0-plugin-in-picard/611697/3 i noticed you can only use the Replaygain 2.0 plugin when your song is part of a release. This is a deviation from the original plug-in, which i could on local files in the "right" pane (see screenshots in linked issue, i'm not the OP).

It might be my ignorance but the Replaygain 2.0  does not seem to work well for my use-case, selecting files > calc Replaygain + calc BPM from the Plugins context menu. I manage tags manually.

Can the Replaygain 2.0 plugin please also be enabled on the "right" pane?

phw:
This was actually by design, but this can be reconsidered. Please see the following discussion for some background:

https://github.com/metabrainz/picard-plugins/pull/336#discussion_r980509606

@complexlogic What's your take on this?

complexlogic:
@phw I remember we discussed whether or not to support the file mode before the initial release of the plugin. I thought it wasn't necessary, but I may have underestimated the number of Picard users that manually tag files because several people have now requested this feature. @tacuna is right that it is a loss of functionality, because the original ReplayGain plugin *did* support the file mode. 

Perhaps it would be worthwhile to add support for the file and cluster plugin actions. To do that, we need to refactor the plugin to operate on the Picard `File` object, instead of `Track`. This would require a moderate amount of work.

I'm currently busy working on another project, but I may revisit this in a month or two. Of course, anybody else that wants to work on adding this feature is welcome to do so.

Xerpocalypse:
Hello, 

I know this is an older issue, but as someone who recently discovered Picard and who doesn't necessarily want to immediately commit to Picard's tags, I'd greatly appreciate the ability to use it only to apply ReplayGain.

tacuna:
This issue is fixed in https://github.com/metabrainz/picard-plugins/pull/392 and available, thanks!

@phw  I do not know what the process is in this repo, should i close this issue or will it be added to some release and closed?

phw:
Can be closed. Thanks for the reminder

PR #392, opened 2025-06-07
Add support for calculating ReplayGain on unidentified clusters

emersonrp:
This commit allows unidentified clusters in the "left pane" to be run through the plugin and have their ReplayGain calculated.  Previously the plugin could only be used on identified tracks and albums in the "right pane."

With this commit, one or more clusters can be scanned as albums, and the album and track ReplayGain added to the files.  There is no change to the existing / previous functionality for identified albums and tracks.

### DIFF
--- a/plugins/replaygain2/__init__.py
+++ b/plugins/replaygain2/__init__.py
@@ -8,9 +8,9 @@ This plugin depends on the ReplayGain utility [rsgain](https://github.com/comple
 are required to install rsgain and set its path in the plugin settings before use.
 
 #### Usage
-Select one or more tracks or albums, then right click and select Plugin->Calculate ReplayGain. The plugin
-will calculate ReplayGain information for the selected items and display the results in the metadata
-window. Click the save button to write the tags to file.
+Select one or more tracks, albums, or clusters then right click and select Plugin->Calculate ReplayGain.
+The plugin will calculate ReplayGain information for the selected items and display the results in the
+metadata window. Click the save button to write the tags to file.
 
 The following file formats are supported:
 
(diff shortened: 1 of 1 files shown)


emersonrp:
I see that Codacy is saying that the "calculate_replaygain" method is too complex.  I think it's not appreciably more complex than it used to be, though it of course is a bit longer.  I can refactor this to get around the Codacy issue, but that expands the scope of the change, and has more possible impact on the existing functionality.  Happy to do the right thing, be it refactoring or just living with the Codacy issue, as y'all see fit.  Thanks.

emersonrp:
Also I have just noticed that this is actually an open issue, #344

phw:
Thanks a lot, this is very helpful.

> I see that Codacy is saying that the "calculate_replaygain" method is too complex. I think it's not appreciably more complex than it used to be, though it of course is a bit longer. I can refactor this to get around the Codacy issue, but that expands the scope of the change, and has more possible impact on the existing functionality.

Yes, I think the codacy check is a bit too strict here. It calculates complexity by considering the different possible code paths in some way. And because your change introduces several "if" conditions it now got above the threshold.

We could address this by splitting the function `calculate_replaygain` further up into separate functions, e.g. extracting building the file list or applying the metadata to files, but I don't think it does help much for the code readability in this case.

I think we can raise the codacy complexity limit or ignore the complaint for this case.

Otherwise the code looks fine to me. I'll test run this and then I think we can merge this change.

> Also I have just noticed that this is actually an open issue, https://github.com/metabrainz/picard-plugins/issues/344

Yes, exactly, this is the fix for this issue. As mentioned it was initially left out intentionally, but has been requested frequently since then. Great you provided this fix.

phw (approved):
Everything working as expected. Thanks again for this contribution

PR #393, opened 2025-08-09
Adjust plugin priority

rdswift:
Adjust the priority of the plugin to ensure that other plugins that modify the performer information, such as `performer_tag_replace`, are executed first.


rdswift:
Replaced with https://github.com/metabrainz/picard-plugins/pull/394 to avoid merge conflicts with other PRs.

PR #395, opened 2025-08-09
Add `performer_tag_replace` plugin

rdswift:
Adds the `performer_tag_replace` plugin.

## Overview

This plugin provides the ability to replace text in performer tags. Once installed a settings page will be added to Picard's options, which is where the plugin is configured.

---

## Settings

The settings panel allows the user to provide a list of the original/replacement pairs used to modify the keys for the performer tags. Each pair must be entered on a separate line in the form:

```
original character string=replacement character string
```

Blank lines and lines beginning with an equals sign (=) will be ignored. Replacements will be made in the order they are found in the list.

There is also a setting which allows the user to choose whether or not to apply the replacements to the artist values in addition to the instrument / vocal keys.  By default, only the keys are processed.

---

### Example 1:

Suppose that you don't like instruments listed as "instrument family", but would rather have it shown as the plural of the instrument.  This could be done using the following:

```
==========================================
= Remove "family" from instrument groups =
==========================================
s family=ses
 family=s
```

Note that the last line begins with a single space.

This would cause the performer credit `Performer [guitar family]: Chet Atkins` to be changed to `Performer [guitars]: Chet Atkins`.

Note that the `s family=ses` would match an instrument family where the instrument ended in "s" and add "es" rather than just an "s".  For example if there was a "bass family", this would change it to "basses" rather than "basss".  (As of this writing, I don't think there are any cases in the database where the instrument in an instrument family ends with an "s", so that replacement pair should never be triggered.)

### Example 2:

Suppose that you don't like the term "additional", but would rather have it shown as "extra".  This could be done using the following:

```
==================================
= Change "additional" to "extra" =
==================================
additional=extra
```

This would cause the performer credit `Performer [additional background vocals]: Jeen` to be changed to `Performer [extra background vocals]: Jeen`.

### Example 3:

Suppose that you want to show the instrument as "drums" rather than "drums (drum set)", or "membranophone". This could be done using the following:

```
=====================================
= Standardize instrument to "drums" =
=====================================
drums (drum set)=drums
drum set=drums
membranophone=drums
```

This would cause the performer credit `Performer [membranophone]: Ringo Starr` to be changed to `Performer [drums]: Ringo Starr`.


phw:
Cool, that's useful. I'll do a proper review later, but it looks good overall.

> Suppose that you want to show the instrument as "drums" rather than "drums (drum set)", or "membranophone".

Just slightly related, but I use this script for this specific purpose:

```
$set(performer:drums,$if2($get(performer:drums \(drum set\)),%performer:membranophone%))
$delete(performer:drums \(drum set\))
$delete(performer:membranophone)

$set(performer:guitar,$if2(%performer:guitar%,$get(performer:guitar family)))
$delete(performer:guitar family)
```

plugins/performer_tag_replace/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+        self.ui.performer_tag_replace_performers.setChecked(config.setting["performer_tag_replace_performers"])
+
+    def save(self):
+        # Replacement settings
+        config.setting["performer_tag_replacement_pairs"] = self.ui.performer_tag_replacement_pairs.toPlainText()
+        config.setting["performer_tag_replace_performers"] = self.ui.performer_tag_replace_performers.isChecked()
+
+
+# Register the plugin to run at a LOW priority.
+register_track_metadata_processor(performer_tag_replace, priority=10)
```
phw, line 165: Priority 10 is actually higher then the default 0. That makes the comment misleading. When using the `PluginPriority` constants `LOW` would be -100.

I guess the value you chose is intentional to work well together with the other performer related plugins. But I think the comment should reflect this.

phw (approved):
Looks good to me. Minor comment only

plugins/performer_tag_replace/__init__.py
rdswift: I did, and I agree that the comment should be changed to reflect that.  I'll push a change.  Good catch!

phw:
Let's release this into the wild then :) Thanks

Sophist-UK:
Nice.

PR #396, opened 2025-08-25
fix(bpm): Only update file in callback

kgarner7:
When using the `bpm` plugin, I've observed a bug where after using this plugin, where Picard would freeze after updates. This PR apparently resolves it, by moving the `update` call into the callback instead of in the initial call


phw (changes requested):
Interesting. But yes, makes sense. `.update` triggers a UI update, and that should run on the main thread anyway.

Can you also bump the version? Then we can directly merge this.

phw (approved):
Thanks, LGTM

PR #397, opened 2025-09-17
fix: lowercase ident

nrth3rnlb:
The variable `ident` contains a string with upper and lower case letters. When saving in Picard, this causes everything in the resulting tag to be converted to lower case. The next time the album or track is refreshed, a difference is now detected because the plugin returns upper and lower case. Picard compares with the existing tag, which is only written in lower case. The result is that the lower case tag is deleted and replaced with a mixed case tag.

### DIFF
--- a/plugins/critiquebrainz/critiquebrainz.py
+++ b/plugins/critiquebrainz/critiquebrainz.py
@@ -59,7 +59,7 @@ def result_review(album, metadata, data, reply, error):
         if reviews:
             for review in reviews:
                 if "last_revision" in review:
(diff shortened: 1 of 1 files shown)


phw:
Good catch. Yes, having this lowercase is better, it's how Picard internally handles all the known tags.

Can you bump `PLUGIN_VERSION`? Then I can have this merged and the changed plugin will be available a few hours later in the official list.

phw (approved):

phw:
Thanks!

PR #398, opened 2025-10-18
Fix: Make `ident` case insensitive.

nrth3rnlb:
Write `ident` entirely in lower case. Why?

MP converts the name of the tag to lower case when saving. When the data is retrieved again, a case sensitive comparison is performed, which always leads to an update in version 1.0.1. This fix resolves the issue.


Sophist-UK:
Sorry - but this PR now seems to be doing two things:

1. A fix to the critiquebrainz plugin for ident;
2. A completely new Remove Soundtrack (Indicator from Release Title) plugin.

These really need to be in separate PRs. 

(And as an aside, the Remove Soundtrack plugin does not actually remove a sound track file - it "Remove(s) OST indicators from Release titles". Can we rename it when you split the PR?)

Sophist-UK:
Also seems to duplicate #397.

nrth3rnlb:
Two fundamentally different changes in one pull request. Sorry.

PR #406, opened 2025-11-04
Sort collection list alphabetically

chaban-mb:
The dropdown list of user collections on the plugin's options
page was previously unsorted, making it difficult to find
a specific collection in a long list.

This change sorts the list case-insensitively by collection
name to improve usability.

_Note: Description and code AI-written_


phw (changes requested):
Thanks for the contribution. I totally agree on that change.

Can you also bump the `PLUGIN_VERSION` to 0.1.2? Then we can merge this change

phw (approved):
Thanks

PR #407, opened 2025-11-07
feat(plugin): Replace Unwanted Characters

nrth3rnlb:
# Replace Unwanted Characters (Picard plugin)

Replace Unwanted Characters is a MusicBrainz Picard plugin that replaces characters often problematic in filenames or filesystems (for example `/`, `*`, `?`, `"` , `.`) with visually similar Unicode alternatives.

## Features

- Replace specified characters in configurable tags (default: `album`, `artist`, `title`, `albumartist`, `releasetype`, `label`).
- Configurable default character mapping table.
- Per-tag mappings: enable either the default mapping or a custom selection for each tag.
- Tagger script function: `$replace_unwanted()` for use in Picard scripts.


plugins/replace_unwanted_characters/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+    <alex_rustler@rambler.ru>
+'''
+
+def get_config_settings():
+    """Load settings from config: returns (filter_tags, default_table, per_tag_tables)"""
+
+    try:
+        filter_tags = config.setting["replace_unwanted_characters_filter_tags"]
+    except KeyError:
+        filter_tags = DEFAULT_TAGS
```
phw: This custom settings handling should all not be needed if options are properly registered (by declaring the `Option` instances as done below.

For registered options it is safe to do e.g. `config.setting["replace_unwanted_characters_filter_tags"]` and it will use the default value also,

plugins/replace_unwanted_characters/__init__.py
```diff
@@ -0,0 +1,621 @@
...
+            table = {k: v for k, v in default_table.items() if k in set(keys_list)}
+
+        metadata[name] = _replace_with_table(value, table)
+
+def script_replace_unwanted(parser, value):
+    # Tagger function: use configured default mapping
+    try:
+        default_table = config.setting["replace_unwanted_characters_char_table"]
+    except KeyError:
+        default_table = DEFAULT_CHAR_MAPPING
```
phw: The exception handling is not needed here

phw (changes requested):
Not sure who invited copilot here, or is Github now just forcing this everywhere?

Anyway, thanks for the plugin. Such a replacement plugin is definitely useful.

I especially like that the list of tag names to process automatically is configurable and that the plugin also provides a tagger function.

The code looks good so far. A few comments about settings handling. We don't need so much boilerplate code for handling settings, Picard already handles things like default values.

Also I haven't looked very close at the UI yet. I'll test this separately.

> Replace Unwanted Characters is a MusicBrainz Picard plugin that replaces characters often problematic in filenames or filesystems (for example `/`, `*`, `?`, `"` , `.`) with visually similar Unicode alternatives.

To be honest I don't fully understand that use case. Picard is already handling this for filenames and has options to configure the replacements.

nrth3rnlb:
"We don't need so much boilerplate code for handling settings, Picard already handles things like default values."

Oh, I didn't know that. 😳

Do you have a small example for me? Or a link?

---

I actually only wrote the plugin because I wanted to replace the dot with another character that doesn't serve as a separator between the prefix and suffix in the file name. The built-in functionality in Picard can't do that.

There is already a plugin called "Replace Forbidden Symbols" that does something like this, but it cannot be configured, and since I wanted to offer something more than just a simple configuration option, I added support for defining a separate replacement for each configured Tag.

plugins/replace_unwanted_characters/__init__.py
```diff
@@ -0,0 +1,139 @@
+# -*- coding: utf-8 -*-
+
+from PyQt5 import QtWidgets, QtCore
+from picard import config, log
+from picard import metadata
+from picard.script import register_script_function
+from picard.ui.options import register_options_page
+
+__version__ = '1.1.2'
```
Sophist-UK: Since this is a first version of the script, it should be numbered either "0.1" or "1.0" and the version number only gets incremented from this once for each subsequent PR.

nrth3rnlb: I have a repository for each plugin and adhere to [Semantic Versioning 2.0.0](https://semver.org/). Would it be better to use two versions now? One for my repository per plugin and one for your [repository](https://github.com/metabrainz/picard-plugins/)?

Sophist-UK: What you do with your own repository is entirely up to you.

nrth3rnlb:
I am closing this pull request because the avalanche of AI references, tests that are not described anywhere, and processes that are not transparent are taking up too much of my time.

Sophist-UK:
I am sorry that you are finding this so difficult - because it really isn't. 

The AI is giving you a code review - and this is really no different to a human code review from e.g. me or Zas or Philipp. All you need to do is to consider whether each of the suggestions is valid and reasonable or not and either follow the suggestion, or do something similar or add a comment stating why you are rejecting the review comment.

These review comments are teamwork contributions to help you to make a better plugin that has better code, is more robust and gives a better experience to users. They are not harsh or personal criticisms - so please take them in the helpful way that they are intended.

The tests are repo wide tests that run on every plugin to check the basics. The tests are self documenting and they are in  the repo, and so you can read them yourself to work out why your plugin(s) are not passing the tests. (And it may not be your plugin that is failing - it could be a different plugin.

As for processes which are not transparent, I am really not sure what you are referring to here. 

The process is a simple one:

1. You submit a PR - when you have finished development and have no more changes to make yourself, you mark it as final rather than draft and request reviews from the project's maintainers.

2. You mark the PR as draft again, respond to any review comments as described above, mark it as final again and request reviews again.

But as stated elsewhere you need to be patient because both Picard and Picard Plugins repos are managed by volunteers who have day jobs and lives and so cannot necessarily get to reviewing your submissions that quickly.  

P.S. When you lose patience and close the PR, you are not only trashing the time you have invested in the plugin, but you are also trashing the time ***I*** have invested in trying to help you, and it feels to me like a slap in the face. Is this really the impression you want to give?

phw:
I'm actually sorry for the AI reviews. I don't think those are intentional. I didn't enable anything for this. Maybe someone else here did, not sure. But I also can't also find some setting to disable it. It looks like Github is forcing this right now.

But to be honest I don't fully understand the reaction. You actually addressed my review comments. And my review actually was that the code looks good, just some minor cleanup for settings handling. Which you addressed, then closed the PR.

If there is still some interest in getting this plugin merged feel free to reopen.

nrth3rnlb:
reopen

nrth3rnlb:
reopen

PR #419, opened 2025-11-23
Add LRCLIB Lyrics External Save plugin

Harmonese:
This plugin fetches lyrics from LRCLIB based on track metadata after saving files and writes synced or plain lyrics as external sidecar files without modifying audio tags.

### DIFF
--- a/plugins/lrclib_lyrics_external_save/__init__.py
+++ b/plugins/lrclib_lyrics_external_save/__init__.py
@@ -1,0 +1,180 @@
+# -*- coding: utf-8 -*-
+
+PLUGIN_NAME = "LRCLIB Lyrics External Save"
+PLUGIN_AUTHOR = "Harmonese"
+PLUGIN_DESCRIPTION = (
+    "Fetch lyrics from LRCLIB based on track metadata after saving files, "
+    "and write synced lyrics (.lrc) or plain lyrics (.txt) as external sidecar "
+    "files in the same folder as the audio file. Does not modify audio tags."
+)
+PLUGIN_VERSION = "1.0.0"
+PLUGIN_API_VERSIONS = ["2.0", "2.1", "2.2", "2.3", "2.4", "2.5", "2.6"]
+PLUGIN_LICENSE = "MIT"
+PLUGIN_LICENSE_URL = "https://opensource.org/licenses/MIT"
+
+import os
+import json
+import ssl
+from urllib.parse import urlencode
+from urllib.request import Request, urlopen
+
+from picard import log
+from picard.file import register_file_post_save_processor
+
+
+LRCLIB_URL = "https://lrclib.net/api/get"
+
+
+# ============================================================
+# Build LRCLIB query from saved file metadata
+# ============================================================
+
+def _build_query_from_file(file):
+    """
+    Extract metadata from file.metadata (after the file is saved) and
+    build a LRCLIB query dictionary.
+    """
+    md = file.metadata
+
+    title = md.get("title")
+    artist = md.get("artist")
+    album = md.get("album")
+    duration = None
+
+    # Use ~length (mm:ss) as duration information
+    length_str = md.get("~length")
+    if length_str:
+        try:
+            parts = length_str.split(":")
+            if len(parts) == 2:
+                minutes = int(parts[0])
+                seconds = int(parts[1])
+                duration = minutes * 60 + seconds
+        except Exception as exc:
+            log.info(
+                "LRCLIB SIMPLE: Failed to parse ~length %r: %r",
+                length_str, exc,
+            )
+
+    log.info(
+        "LRCLIB SIMPLE: Query → title=%r, artist=%r, album=%r, duration=%r",
+        title, artist, album, duration,
+    )
+
+    if not (title and artist and duration):
+        return None
+
+    return {
+        "track_name": title,
+        "artist_name": artist,
+        "album_name": album or "",
+        "duration": duration,
+    }
+
+
+# ============================================================
+# Request LRCLIB (SSL verification disabled for compatibility)
+# ============================================================
+
+def _fetch_lyrics_from_lrclib(query):
+    """
+    Send HTTP request to LRCLIB using urllib with SSL verification disabled
+    (macOS Python inside Picard often lacks a certificate store).
+    """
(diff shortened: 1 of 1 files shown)


plugins/lrclib_lyrics_external_save/__init__.py
```diff
@@ -0,0 +1,180 @@
...
+    length_str = md.get("~length")
+    if length_str:
+        try:
+            parts = length_str.split(":")
+            if len(parts) == 2:
+                minutes = int(parts[0])
+                seconds = int(parts[1])
+                duration = minutes * 60 + seconds
+        except Exception as exc:
+            log.info(
```
Sophist-UK: log.error?

plugins/lrclib_lyrics_external_save/__init__.py
```diff
@@ -0,0 +1,180 @@
...
+                length_str, exc,
+            )
+
+    log.info(
+        "LRCLIB SIMPLE: Query → title=%r, artist=%r, album=%r, duration=%r",
+        title, artist, album, duration,
+    )
+
+    if not (title and artist and duration):
+        return None
```
Sophist-UK, line 98: I think that the check for title, artist and duration should be before the previous log.info and should include a log.warning about data not being available.

plugins/lrclib_lyrics_external_save/__init__.py
```diff
@@ -0,0 +1,180 @@
...
+# ============================================================
+# Request LRCLIB (SSL verification disabled for compatibility)
+# ============================================================
+
+def _fetch_lyrics_from_lrclib(query):
+    """
+    Send HTTP request to LRCLIB using urllib with SSL verification disabled
+    (macOS Python inside Picard often lacks a certificate store).
+    """
+    url = LRCLIB_URL + "?" + urlencode(query)
```
Sophist-UK: A very minor point, but you could include the "?" in LRCLIB_URL.

plugins/lrclib_lyrics_external_save/__init__.py
```diff
@@ -0,0 +1,180 @@
...
+    log.info("LRCLIB SIMPLE: Requesting %s", url)
+
+    ctx = ssl._create_unverified_context()
+
+    try:
+        req = Request(
+            url,
+            headers={"User-Agent": "Picard-LRCLIB-External-Save"},
+        )
+        with urlopen(req, context=ctx, timeout=10) as resp:
```
Sophist-UK: I am guessing that this is a synchronous web service call - and if so this could cause the UI to become erratically unresponsive. It is "normal" in Picard to queue web calls using Picard's standard queueing mechanism, and then use a call-back to process the results (and this also allows you to specify a rate limit to avoid web sites from blocking your IP address due to excessive calls).

But if you are going to stick with a synchronous web call and want to avoid the UI being unresponsive, you can  make the call from a worker thread.

plugins/lrclib_lyrics_external_save/__init__.py
```diff
@@ -0,0 +1,180 @@
...
+    plain = obj.get("plainLyrics")
+
+    if synced:
+        log.info("LRCLIB SIMPLE: Synced lyrics found (%d chars)", len(synced))
+        return synced, True
+    if plain:
+        log.info("LRCLIB SIMPLE: Plain lyrics found (%d chars)", len(plain))
+        return plain, False
+
+    return None, None
```
Sophist-UK, line 165: Should the message about lyrics not found be here rather than after the call to this method?

Sophist-UK (changes requested):
Lyrics is a something that is a very common requirement so any plugins that help with this are a great idea IMO.

On the whole this seems very well written - my only major concern is whether using synchronous web calls rather than asynchronous ones might cause erratic UI behaviour.

Some people want lyrics to be stored as e.g. ID3 tags inside the music files. Once this PR is approved, are you planning to make a similar one (using a track metadata hook rather than a post file save hook) for getting lyrics and storing them as tags?

Sophist-UK (changes requested):
Extra comment because I forgot to select Request Changes on the last review.

Harmonese:
> Lyrics is a something that is a very common requirement so any plugins that help with this are a great idea IMO.
> 
> On the whole this seems very well written - my only major concern is whether using synchronous web calls rather than asynchronous ones might cause erratic UI behaviour.
> 
> Some people want lyrics to be stored as e.g. ID3 tags inside the music files. Once this PR is approved, are you planning to make a similar one (using a track metadata hook rather than a post file save hook) for getting lyrics and storing them as tags?

I thought the existing plugin "lrclib lyrics" has this function. It does not use synced lyrics as tags though, which could be improved

plugins/lrclib_lyrics_external_save/__init__.py
```diff
@@ -0,0 +1,262 @@
...
+    snapshot is a plain dict with at least:
+        title, artist, album, ~length
+    """
+    title = snapshot.get("title")
+    artist = snapshot.get("artist")
+    album = snapshot.get("album")
+    duration = None
+
+    # parse "~length" safely
+    length_str = snapshot.get("~length")
```
phw, line 66: Use `metadata.length` (integer value in milliseconds) instead of paring the formatted `~length`

plugins/lrclib_lyrics_external_save/__init__.py
```diff
@@ -0,0 +1,262 @@
...
+PLUGIN_API_VERSIONS = ["2.0", "2.1", "2.2", "2.3", "2.4", "2.5", "2.6"]
+PLUGIN_LICENSE = "MIT"
+PLUGIN_LICENSE_URL = "https://opensource.org/licenses/MIT"
+
+import json
+import os
+import ssl
+import threading
+from urllib.parse import urlencode
+from urllib.request import Request, urlopen
```
phw, line 39: Instead of urllib Picard's WebService class should be used.

phw (changes requested):
We already have a lrclib plugin, which also can save lyrics as external files.

If the goal is to explicitly not write any tags, but only save as external file, than I think we should rather update the existing plugin to have this option.

The plugin has some other problematic aspects, like loading lyrics only after saving and not using Picard's WevService class.

Harmonese:
The existing LRCLIB plugin already have the function to save the lyrics as external files, this is no longer needed.

PR #420, opened 2025-11-27
Lookup in Harmony

JadedBlueEyes:
Adds a plugin to look up a release in harmony. Makes adding a new release or updating a release slightly more convenient.

### DIFF
--- a/plugins/lookup_harmony/__init__.py
+++ b/plugins/lookup_harmony/__init__.py
@@ -1,0 +1,99 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.
+
+PLUGIN_NAME = 'Look Up with Harmony'
+PLUGIN_AUTHOR = 'Jade'
+PLUGIN_DESCRIPTION = '''
+Adds a right click option to look up releases on Harmony using the GTIN code.
+If a MusicBrainz release ID is available, it will be included in the lookup URL.
+'''
+PLUGIN_VERSION = '0.1.0'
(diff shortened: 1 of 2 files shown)


plugins/lookup_harmony/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+            )
+        )
+
+        if isinstance(mbid_value, list):
+            musicbrainz_albumid = str(mbid_value[0]).strip() if mbid_value else ""
+        else:
+            musicbrainz_albumid = str(mbid_value).strip() if mbid_value else ""
+
+        # Build the Harmony URL
+        url = "https://harmony.pulsewidth.org.uk/release?gtin={0}&deezer=&itunes=&spotify=&tidal=&musicbrainz={1}".format(
```
kellnerd, line 95: Two quick thoughts here:
1. Providers shouldn't be hard-coded, there are provider categories (currently allowed values: default, all, preferred) which can be used instead:
```suggestion
        url = "https://harmony.pulsewidth.org.uk/release?gtin={0}&category=preferred&musicbrainz={1}".format(
```
2. It might make sense to make the base URL configurable as well. Although most people use the pulsewidth server, alternative instances could then be used.

plugins/lookup_harmony/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+    try:
+        barcode_value = metadata.get("barcode", "")
+        log.debug(
+            "{0}: barcode tag raw content: {1} (type: {2})".format(
+                PLUGIN_NAME, repr(barcode_value), type(barcode_value).__name__
+            )
+        )
+
+        # Convert to string and strip (handle lists, etc.)
+        if isinstance(barcode_value, list):
```
phw, line 55: This currently doesn't work as expected. `metadata.get` never returns a list, but always a string (with multiple values joined by "; "). Also metadata values are always strings, which simplifies things here a bit.

You should instead use `metadata.getall` which will always return a list. Something like:

```python
if barcode_values := metadata.get("barcode"):
    gtin = barcode_value[0].strip()
if not gtin:
    if upc_values := metadata.get("upc"):
        gtin = upc_values[0].strip()
```

plugins/lookup_harmony/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if not gtin:
+            log.warning(
+                "{0}: No GTIN/UPC code found for this release.".format(PLUGIN_NAME)
+            )
+            return
+
+        log.info("{0}: Using GTIN: {1}".format(PLUGIN_NAME, gtin))
+
+        mbid_value = metadata.get("musicbrainz_albumid", "")
```
phw, line 82: Same as above, use `getall`

phw (changes requested):
Than you Jade for this plugin. Since we focus on Picard 3 currently we lack behind with reviewing Picard 2 plugins.

Overall this looks fine. I have some code comments for metadata handling, and I think kellnerd's comment should also be addressed.

PR #421, opened 2025-11-29
Search Engine Lookup: Additional words setting not saving

chaban-mb:
Previously, text entered into the 'Additional Search Words' field was not updating the internal setting variable. This meant changes were ignored during 'Save' and 'Test'.

This commit connects the textChanged signal of the input field to a handler that updates the self.additional_words variable immediately.

_Note: Description and code AI-written_

**Testing:**
- Ensured the changes are applied immediately when using "Test" button
- Ensured changes are only saved permanently when using "Make It So!" button and are reverted when using "Cancel"
- Ensured changes are picked up by "Search engine lookup" menu in cluster pane.

### DIFF
--- a/plugins/search_engine_lookup/__init__.py
+++ b/plugins/search_engine_lookup/__init__.py
@@ -24,7 +24,7 @@ Adds a right click option on a cluster to look up album information using a sear
 
 Adds a right click option on an album or track to look up cover art for the selected album or title.
 '''
-PLUGIN_VERSION = '2.1.0'
+PLUGIN_VERSION = '2.1.1'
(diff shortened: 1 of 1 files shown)


phw (approved):
Thanks for your fix. Since we focus on Picard 3 currently we lack behind with reviewing Picard 2 plugins.

Changes look good, LGTM

rdswift:
Just so that you know, I already included this functionality in the Picard 3 version of the plugin based on this request.  I handled it slightly differently so that it strips leading / trailing spaces before saving to `self.additional_words`.

PR #417, opened 2025-11-22
feat(plugin): Shelves

nrth3rnlb:
This pull request introduces the new "Shelves" plugin for MusicBrainz Picard, providing a comprehensive system for managing and organizing music files by virtual "shelves" (top-level folders). The implementation includes shelf detection, assignment, workflow automation, and integration with Picard’s UI and scripting system. The main changes are grouped by new feature implementation, plugin architecture, and supporting utilities/constants.

**Major new features and architecture:**

*Initial implementation of the Shelves plugin:*

- Added `plugins/shelves/__init__.py` to register the plugin with Picard, define plugin metadata, and set up all integration points (file/track processors, context menu actions, options page, and script functions). Introduced a global, thread-safe `ShelfManager` class for tracking and resolving shelf assignments per album, including conflict detection and logging.

*User interface and actions:*

- Implemented `plugins/shelves/actions.py` with context menu actions: "Set shelf name..." (allows manual assignment via dialog with validation and known shelf suggestions) and "Determine shelf" (automatically detects shelf from file paths, updates metadata, and resolves conflicts). Includes UI dialog for shelf selection and validation feedback.

*Constants and configuration:*

- Added `plugins/shelves/constants.py` to centralize all plugin constants, such as tag keys, default shelf names, validation rules, config keys, and album indicators. This improves maintainability and ensures consistency across the plugin.

*Shelf management logic:*

- Added `plugins/shelves/manager.py` defining a `ShelfManager` class for managing album-to-shelf mappings, voting logic when files suggest different shelves, and conflict logging. This supports robust and maintainable shelf assignment logic.

PR #422, opened 2026-01-01
Fix invalid escape sequence for Python 3.12

shifanatk:
### Summary
This PR fixes `SyntaxWarning` issues regarding invalid escape sequences that occur when running the plugin with Python 3.12.

### Changes
* Added the `r` (raw string) prefix to regular expression patterns in `plugins/classical_extras/__init__.py`.
* Specifically addressed the `\W` escape sequences that were causing syntax warnings.

### Related Issues
Fixes the Python 3.12 syntax errors mentioned in Issue #417.


phw:
Thanks for your changes. Since we focus on Picard 3 currently we lack behind with reviewing plugins.

This fix looks good. The plugin version needs to be bumped to 2.0.15 before merging, though.

phw (changes requested):
Version bump

PR #423, opened 2026-03-29
Add plugin: Pseudo-Release Title Sort

alydevs:
This plugin checks if a track title in a pseudo-release is different to the recording title, and applies the track title to the `titlesort` tag if so. This is useful in cases such as [Kessoku Band](https://musicbrainz.org/release/dc4e94fb-eeb9-46e1-ac40-98dbdd772b0e), where the English track names for this pseudo-release are shown in Picard, but the original kanji/kana script is used for titles. There are many instances of people wondering about this behaviour going back a couple years: https://redd.it/13fi3ur https://redd.it/1hud7g1 https://community.metabrainz.org/t/japanese-music-titles-into-their-english-names/557050/2

Before: English titles are shown in track list when adding pseudo-release, but tags only have Japanese titles, and there is no way to write English titles into tags
image
image

After: English titles are written to titlesort tag, this can be used in file naming script to set consistent filenames across a library. Original title is currently retained in track Title tag.
image

### DIFF
--- a/plugins/pseudo-release-titlesort/pseudo-release-titlesort.py
+++ b/plugins/pseudo-release-titlesort/pseudo-release-titlesort.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2026 alydevs <aly@aly.pet>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Disclosure: Claude Sonnet 4.6 was used in the creation of pseudo-release-titlesort.py
+
+# Plugin metadata
+# =============================================================================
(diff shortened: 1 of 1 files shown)


phw:
Thanks for plugin. Since we focus on Picard 3 currently we lack behind with reviewing Picard 2 plugins, hence reviews currently take longer.

I'm not yet fully sure I understand the purpose of the plugin, though. If you tag against e.g. the pseudo release https://musicbrainz.org/release/dc4e94fb-eeb9-46e1-ac40-98dbdd772b0e directly, then the plugin would set the `titlesort` to the track title. But this then is the same as the `title`, since that is also the track title. The original Japanese title would not be used.

If you tag against the original Japanese release instead of the pseudo-release then of course you get Japanese track titles, but in this case the plugin would not apply and not `titlesort` would be set.

I would understand if the plugin would also set the `title` to the recording title.

That would lossely be the same as using this script:

```taggerscript
$if($eq(%releasestatus%,pseudo-release),
$set(titlesort,%title%)
$set(title,%_recordingtitle%)
)
```

phw (changes requested):
See above

alydevs:
Hi, thanks for reviewing my PR.

I had no luck getting Picard to perform the behaviour you described without making this plugin. As described in the first post, all track tags in the pseudo release were written by Picard as the kana/kanji script instead of English. They were shown with the English titles in the right-pane album listing, but not in the track tags pane at the bottom.

This difference was frustrating when trying to troubleshoot the problem, because as far as I could tell, the track tags being written via the pseudo release *should* have been the English names, but this was not happening.

phw:
Then something else is at play. Do you have any script running that might change the titles? Or is the title tag maybe set as preserved tag in Options > Tags? 

Your screenshot shows only  the original value, which is what is already in the file. If you have the title marked as a preserved tag Picard would not overwrite that value and keep it as is.

If the correct title shows on the right is shows the proper data is loaded. That view uses the exact same metadata as what then also used for writing the tags.

PR #424, opened 2026-04-20
Ensure new provider key is a string

rdswift:
An issue was raised in https://tickets.metabrainz.org/browse/PICARD-3253 whereby the plugin would generate an exception when selecting a newly created provider.  The problem was that the new provider id was being created as a UUID and not converted to a string.


rdswift:
Looks like we need to update the tests to remove Python 3.9 from the test platforms.

phw (approved):

phw:
Thanks. I'll take care of the CI issue

zas (approved):
LGTM

PR #425, opened 2026-04-25
Add plugin: matroska_tagger

BenMcLean:
Adds support for reading and writing tags in Matroska (.mkv, .mka) files via MKVToolNix.

I'm sure the ideal solution would be to fix mutagen to support Matroska but progress on that has stalled. My solution, a plugin that uses MKVToolNix, already immediately works right now today.

### DIFF
--- a/plugins/matroska_tagger/matroska_tagger.py
+++ b/plugins/matroska_tagger/matroska_tagger.py
@@ -28,8 +28,6 @@ import os
 import subprocess
 import sys
 import tempfile
-import xml.etree.ElementTree as ET
-
 from picard import config, log
 from picard.config import BoolOption, TextOption
 from picard.file import File
@@ -37,34 +35,26 @@ from picard.formats import register_format
 from picard.metadata import Metadata
 from picard.ui.options import register_options_page, OptionsPage
 
-try:
-	from PyQt6.QtWidgets import (
-		QCheckBox,
-		QFileDialog,
-		QFormLayout,
-		QHBoxLayout,
-		QLabel,
-		QLineEdit,
-		QPushButton,
-		QVBoxLayout,
-	)
-	from PyQt6.QtCore import Qt
+from PyQt5.QtWidgets import (
+	QCheckBox,
+	QFileDialog,
+	QFormLayout,
+	QHBoxLayout,
+	QLabel,
+	QLineEdit,
+	QPushButton,
+	QVBoxLayout,
+)
+from PyQt5.QtCore import (
+	QBuffer,
+	QFile,
+	QIODevice,
+	QXmlStreamReader,
+	QXmlStreamWriter,
+	Qt,
+)
 
-	_AlignTop = Qt.AlignmentFlag.AlignTop
-except ImportError:
-	from PyQt5.QtWidgets import (
-		QCheckBox,
-		QFileDialog,
-		QFormLayout,
-		QHBoxLayout,
-		QLabel,
-		QLineEdit,
-		QPushButton,
-		QVBoxLayout,
-	)
-	from PyQt5.QtCore import Qt
-
-	_AlignTop = Qt.AlignTop
+_AlignTop = Qt.AlignTop
 
 
 # ---------------------------------------------------------------------------
@@ -198,16 +188,11 @@ def _get_tooldir():
 # ---------------------------------------------------------------------------
 
 
-def _add_simple(parent, name, value):
-	simple = ET.SubElement(parent, "Simple")
-	ET.SubElement(simple, "Name").text = name
-	ET.SubElement(simple, "String").text = str(value)
-
-
-def _xml_doc(root):
-	return '<?xml version="1.0" encoding="UTF-8"?>\n' + ET.tostring(
-		root, encoding="unicode"
-	)
+def _write_simple(w, name, value):
+	w.writeStartElement("Simple")
+	w.writeTextElement("Name", name)
+	w.writeTextElement("String", str(value))
+	w.writeEndElement()
 
 
 def _tags_xml(metadata):
@@ -230,44 +215,57 @@ def _tags_xml(metadata):
 	on global tags (e.g. "ALBUM/TITLE"), which breaks Kodi's exact-match lookup.
 	TargetTypeValue is still written for spec-compliant readers.
 	"""
-	root = ET.Element("Tags")
+	buf = QBuffer()
+	buf.open(QIODevice.WriteOnly)
+	w = QXmlStreamWriter(buf)
+	w.setAutoFormatting(True)
+	w.writeStartDocument()
+
+	w.writeStartElement("Tags")
 
 	# --- Level 50: album ---
-	alb = ET.SubElement(root, "Tag")
-	alb_targets = ET.SubElement(alb, "Targets")
-	ET.SubElement(alb_targets, "TargetTypeValue").text = "50"
+	w.writeStartElement("Tag")
+	w.writeStartElement("Targets")
+	w.writeTextElement("TargetTypeValue", "50")
+	w.writeEndElement()  # Targets
 	for picard_name, mkv_name in _ALBUM_TAGS.items():
 		for value in metadata.getall(picard_name):
 			if value:
-				_add_simple(alb, mkv_name, value)
+				_write_simple(w, mkv_name, value)
 	for value in metadata.getall("totaltracks"):
 		if value:
-			_add_simple(alb, "TOTAL_PARTS", value)
+			_write_simple(w, "TOTAL_PARTS", value)
 	# Kodi compat aliases: no level-30 equivalents so they persist in fctx->metadata
 	for value in metadata.getall("album"):
 		if value:
-			_add_simple(alb, "ALBUM", value)
+			_write_simple(w, "ALBUM", value)
 	for value in metadata.getall("albumartist"):
 		if value:
-			_add_simple(alb, "ALBUM_ARTIST", value)
+			_write_simple(w, "ALBUM_ARTIST", value)
 	_date_field = "originaldate" if config.setting["mkv_use_original_date"] else "date"
 	for value in metadata.getall(_date_field):
 		if value:
-			_add_simple(alb, "DATE", value)
+			_write_simple(w, "DATE", value)
+	w.writeEndElement()  # Tag (album)
 
 	# --- Level 30: track ---
-	trk = ET.SubElement(root, "Tag")
-	trk_targets = ET.SubElement(trk, "Targets")
-	ET.SubElement(trk_targets, "TargetTypeValue").text = "30"
+	w.writeStartElement("Tag")
+	w.writeStartElement("Targets")
+	w.writeTextElement("TargetTypeValue", "30")
+	w.writeEndElement()  # Targets
 	for picard_name, mkv_name in _TRACK_TAGS.items():
 		for value in metadata.getall(picard_name):
 			if value:
-				_add_simple(trk, mkv_name, value)
+				_write_simple(w, mkv_name, value)
 	for value in metadata.getall("tracknumber"):
 		if value:
-			_add_simple(trk, "PART_NUMBER", value)
+			_write_simple(w, "PART_NUMBER", value)
+	w.writeEndElement()  # Tag (track)
 
-	return _xml_doc(root)
+	w.writeEndElement()  # Tags
+	w.writeEndDocument()
+	buf.close()
+	return bytes(buf.data()).decode("utf-8")
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+PLUGIN_DESCRIPTION = "Adds support for reading and writing tags in Matroska (.mkv, .mka) files via MKVToolNix."
+PLUGIN_LICENSE = "GPL-2.0-or-later"
+PLUGIN_LICENSE_URL = "https://www.gnu.org/licenses/gpl-2.0.html"
+
+import json
+import os
+import subprocess
+import sys
+import tempfile
+import xml.etree.ElementTree as ET
```
phw: I'm not fully sure whether this is available in the packaged versions for macOS and Windows. Those contain only Python modules that are actually being used, and Picard doesn't use the `xml` package directly. It might be pulled in indirectly, though.

Have you tested this plugin with the Picard packages on one of those OS?

If it doesn't work Picard offers a `parse_xml` helper in `picard.util.xml`, which is based on `QXmlStreamReader`. Picard itself currently does no longer use it for anything, but it is kept around for plugins to use.

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+		QFormLayout,
+		QHBoxLayout,
+		QLabel,
+		QLineEdit,
+		QPushButton,
+		QVBoxLayout,
+	)
+	from PyQt6.QtCore import Qt
+
+	_AlignTop = Qt.AlignmentFlag.AlignTop
```
phw: Importing from PyQt6 will cause the plugin to error during initialization if PyQt6 is available. Picard 2 uses Qt5, and trying to mix Qt6 and Qt5 does not work.

Import only from PyQt5

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"mixer": "MIXED_BY",
+	"djmixer": "DJMIXED_BY",
+	"producer": "PRODUCER",
+	"work": "WORK",
+	"mood": "MOOD",
+	"artists": "ARTISTS",
+	"grouping": "GROUPING",
+	"titlesort": "TITLESORT",
+	"artistsort": "ARTISTSORT",
+	"composersort": "COMPOSERSORT",
```
phw: The Matroska documentation indicates the sort values should be stored as a nested `SORT_WITH` tag on the corresponding main tag, see https://www.matroska.org/technical/tagging.html#nested-information . This is also what MP3Tag does (https://docs.mp3tag.de/mapping/)

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+# Picard internal name → Matroska tag name, for track-level tags (TargetTypeValue=30)
+_TRACK_TAGS = {
+	"title": "TITLE",
+	"artist": "ARTIST",
+	"comment": "COMMENT",
+	"lyrics": "LYRICS",
+	"isrc": "ISRC",
+	"bpm": "BPM",
+	"key": "INITIAL_KEY",
+	"language": "LANGUAGE",
```
phw: Side note: `LANGUAGE` is not defined at https://www.matroska.org/technical/tagging.html , but used the same in MP3Tag (https://www.matroska.org/technical/tagging.html)

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"subtitle": "SUBTITLE",
+	"remixer": "REMIXED_BY",
+	"conductor": "CONDUCTOR",
+	"lyricist": "LYRICIST",
+	"composer": "COMPOSER",
+	"engineer": "SOUND_ENGINEER",
+	"mixer": "MIXED_BY",
+	"djmixer": "DJMIXED_BY",
+	"producer": "PRODUCER",
+	"work": "WORK",
```
phw: Not sure we should include the `WORK` in the mapping for now. This is not officially defined and it feels like it should be put somewhere else in Matroska tags. I would omit this for now.

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"mood": "MOOD",
+	"artists": "ARTISTS",
+	"grouping": "GROUPING",
+	"titlesort": "TITLESORT",
+	"artistsort": "ARTISTSORT",
+	"composersort": "COMPOSERSORT",
+	"license": "LICENSE",
+	"copyright": "COPYRIGHT",
+	"website": "URL_OFFICIAL_ARTIST",
```
phw: I wonder if this is better set as a nested `URL` tag on the `ARTIST` or probably rather `ALBUMARTIST` tag.

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"originalartist": "ORIGINALARTIST",
```
phw: Here also I think we need to work with nested tags. Matroska spec describes an `ORIGINAL` tag that then contains other tags. So IMHO it would be correct to have:

- ORIGINAL
    - ARTIST

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"musicbrainz_recordingid": "MUSICBRAINZ_TRACKID",
```
phw: Just thinking aloud: Should we maybe call this `MUSICBRAINZ_RECORDINGID`? We haven't yet defined those tags for Matroska, and it would be better describe the content. On the other hand it would be inconsistent with the historic tag names used in other formats, but we have those tags different depending on the format anyway.

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"musicbrainz_albumid": "MUSICBRAINZ_ALBUMID",
+	"musicbrainz_releasegroupid": "MUSICBRAINZ_RELEASEGROUPID",
+	"musicbrainz_albumartistid": "MUSICBRAINZ_ALBUMARTISTID",
+	"musicbrainz_discid": "MUSICBRAINZ_DISCID",
+	"musicbrainz_originalalbumid": "MUSICBRAINZ_ORIGINALALBUMID",
+	"replaygain_album_gain": "REPLAYGAIN_GAIN",
+	"replaygain_album_peak": "REPLAYGAIN_PEAK",
+	"replaygain_album_range": "REPLAYGAIN_ALBUM_RANGE",
+	"replaygain_reference_loudness": "REPLAYGAIN_REFERENCE_LOUDNESS",
+}
```
phw, line 133: Likewise to the track tags we could also use `ORIGINAL` with nested tags here for `originalartist` and `originaldate`:

- ORIGINAL
  - ARTIST
  - DATE_RELEASED

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"replaygain_track_peak": "REPLAYGAIN_PEAK",
+	"replaygain_track_range": "REPLAYGAIN_TRACK_RANGE",
+}
+
+# Picard internal name → Matroska tag name, for album-level tags (TargetTypeValue=50)
+_ALBUM_TAGS = {
+	"album": "TITLE",
+	"albumartist": "ARTIST",
+	"date": "DATE_RELEASED",
+	"genre": "GENRE",
```
phw: Genres can be different per track, hence I think this tag should go to `_TRACK_TAGS`.

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"barcode": "BARCODE",
+	"script": "SCRIPT",
+	"media": "ORIGINAL_MEDIA_TYPE",
+	"releasestatus": "RELEASE_STATUS",
+	"releasecountry": "RELEASE_COUNTRY",
+	"releasetype": "RELEASE_TYPE",
+	"asin": "ASIN",
+	"compilation": "COMPILATION",
+	"discnumber": "DISC_NUMBER",
+	"totaldiscs": "DISC_TOTAL",
```
phw: MP3Tag uses `PART_NUMBER` for the disc number. I think that makes sense. But there would be no tag for `totaldiscs`, as this would need to be `TOTAL_PARTS` on a different level.

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,586 @@
...
+	"albumsort": "ALBUMSORT",
+	"albumartistsort": "ALBUMARTISTSORT",
```
phw: Same as for tracks, using `SORT_WITH` nesting.

phw (changes requested):
Thank you for this, having a plugin for MKV is very welcomed. The code overall looks fine, but I have some comments.  Specifically we need to check whether the XML reading/writing works with the Windows and macOS packages of Picard, otherwise this would need to be refactored. Also any attempt to load from PyQt6 must be removed.

I think the most tricky part with Matroska is getting the tag mapping right. The plugin already handles this pretty well, but I have a couple of comments for discussion.

For reference, we should consider:

- The Matroska tagging spec: https://www.matroska.org/technical/tagging.html
- Mapping already used by MP3Tag: https://docs.mp3tag.de/mapping/

> I'm sure the ideal solution would be to fix mutagen to support Matroska but progress on that has stalled. My solution, a plugin that uses MKVToolNix, already immediately works right now today.

That's a fair solution, especially for a plugin. I would probably add a link to the MKVToolNix download site to the plugin description. But wow, at least on mobile https://mkvtoolnix.download/source.html is horribly ad-infested. There is also https://mkvtoolnix.com/#download which looks nicer, but not sure that's official. But at least it offers the downloads from the official website.

BenMcLean:
There is some great feedback here. I have only tested on Windows. I could potentially try testing on Linux. I do not plan to ever support Apple, so while I won't go out of my way to actively sabotage the software from working on macOS, I will not personally be testing macOS.

I think that we need to be careful not to let the perfect become the enemy of the good. People have been trying to make this happen since 2014 and we don't want to still have it sitting around unused in 2034 from not being quite perfect. The issues cited need to be sorted into two separate categories:

1. Issues that must be resolved for the initial version to become accepted as a plugin
2. Issues that can be addressed in a future version down the road. (quite possibly in a v3 plugin since I undertsand Picard 3 is on the horizon)

In the meantime, "Import only from PyQt5" and "picard.util.xml" will get addressed first, then I will assess the rest.

(later edit: After that assessment, it turned out I had overestimated the required effort when I made this post, based solely on the sheer number of issues. Almost all of them were trivial.)

phw:
> I do not plan to ever support Apple, so while I won't go out of my way to actively sabotage the software from working on macOS, I will not personally be testing macOS.

It should be fine. The only concern is really the issue of only using Python packages that come bundled with Picard. If `xml` is available in the installer or portable package for Windows it should also be present in the macOS package. We can of course test this once, but there should not be much OS specific in this kind of plugin.

> I think that we need to be careful not to let the perfect become the enemy of the good. People have been trying to make this happen since 2014 and we don't want to still have it sitting around unused in 2034 from not being quite perfect.

I totally agree with that.

> In the meantime, "Import only from PyQt5" and "picard.util.xml" will get addressed first, then I will assess the rest.

I think the PyQt5 thing is the mandatory part. `picard.util.xml` only provides a helper for reading XML, not writing it. The writing part would be needed to be implemented separately. So I would only make this a real concern in case the Python `xml` package would not be available in the PyInstaller packaged versions of Picard. If it is working there, we can keep it being used in this package.

The tagging part is what I would think we can improve in future updates. But if in doubt I would prefer to omit writing a certain tag and rather implement it properly with a future update.

rdswift:
> So I would only make this a real concern in case the Python `xml` package would not be available in the PyInstaller packaged versions of Picard. If it is working there, we can keep it being used in this package.

Could we include the import in a `try:` block and log an error (and disable the plugin) if it fails?

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: You're right.
The solution is to replace xml.etree.ElementTree with PyQt5.QtCore's QXmlStreamReader and QXmlStreamWriter which are guaranteed to ship with Picard 2 on all platforms.
Fixed in edf94ee556ebac521d248449c88510f447de5314

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Fixed in edf94ee556ebac521d248449c88510f447de5314

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Fixed in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Added comment about this rationale in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Removed in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: The [MusicBrainz Picard tag mapping for Vorbis](https://picard-docs.musicbrainz.org/en/latest/appendices/tag_mapping.html#id39) uses `MUSICBRAINZ_TRACKID` so I believe my code is already correct on this point.

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Fixed in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Fixed in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Fixed in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Resolved in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Addressed in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: Fixed in 7c68c6ef9e47f0db5c77d2debd2509e39168d10d

BenMcLean:
@phw I believe I have resolved all the remaining issues with the latest commit 7c68c6ef9e47f0db5c77d2debd2509e39168d10d and had a successful test on WIndows 11.

phw (changes requested):
Thanks for the big update, from checking the code this is looking good. I'll give this a test run, but I think we are mostly ready to merge.

One last thing, though: `_MatroskaFileBase` should implement the `supports_tag` static method. This takes an (internal) tag name and returns a boolean depending on whether the tag is supported by the format or not.

This is used inside Picard at several places and e.g. avoids unsupported tags being shown as changed / missing.

See for example https://github.com/metabrainz/picard/blob/master/picard/formats/mp4.py#L358-L367

In our case it's mostly checking whether the tag is in the mappings and maybe some special cases.

plugins/matroska_tagger/matroska_tagger.py
BenMcLean: I thought about this a little more and maybe we should be blazing a new path instead of following a precedent from the past which may no longer be relevant. What do you think?

phw: That was my thinking, yes. The current naming is historic, at the time the mapping was defined what is stored in `MUSICBRAINZ_TRACKID` was a track ID. But those track IDs turned into recording IDs when recordings got introduced, only later actual track IDs got introduced again (now stored a bit awkwardly as "release track ID").

It leads to all kinds of confusion, but changing the naming would also. But on the other side we are just about to define the mapping for Matroska, and the exact naming differs for formats like MP3 or ASF as well.

BenMcLean: Changed in fb1fa0490e7de2978353ee972812e9b4aa63d364

plugins/matroska_tagger/matroska_tagger.py
```diff
@@ -0,0 +1,725 @@
...
+	"mood": "MOOD",
+	"artists": "ARTISTS",
+	"grouping": "GROUPING",
+	# titlesort/artistsort/composersort emitted as SORT_WITH nested inside TITLE/ARTIST/COMPOSER
+	"website": "WEBSITE",
+	"license": "LICENSE",
+	"copyright": "COPYRIGHT",
+	# originalartist emitted as ORIGINAL/ARTIST nested structure
+	"musicbrainz_recordingid": "MUSICBRAINZ_RECORDINGID",
+	"musicbrainz_trackid": "MUSICBRAINZ_TRACKID",
```
phw: Let's keep this one as `MUSICBRAINZ_RELEASETRACKID`. I fear otherwise some tools will definitely read the wrong ID when trying to get the recording ID.

BenMcLean: 66d1b76e4d4a2f75670cd04f266c5b86489e5ccd

phw (approved):
Thanks for plugin and for going through the review with it. Much appreciated.

Code looks good and it worked well in my testing. LGTM

BenMcLean:
@phw When will it show up in the list of available plugins in Picard? I had assumed merging would be enough but apparently there's an additional step?

phw:
@BenMcLean It just takes a while. The website gets updated on a schedule, not sure, but I think every 4 hours or so.

phw:
This doesn't work as it should right now. I checked, the implemented default is to sync plugins every 12 hours (the configured value on the website deployment could be different, though, I can't see that config myself). But anyway, by now the plugin should show up, but doesn't. We'll investigate.

phw:
@BenMcLean There was indeed an issue, the scheduler did not run. We fixed the issue and the plugin list is now up-to-date, I clouding this Matroska plugin

PR #426, opened 2026-05-06
plugins/haikuattrs.py: `sys.platform` returns just 'haiku' now.

OscarL:
Could alternatively use: `if sys.platform[:5] == 'haiku':`.

Issue is that all current Python interpreters on Haiku now return 'haiku' on `sys.platform`.

Companion change to https://github.com/metabrainz/picard/pull/3149.

### DIFF
--- a/plugins/haikuattrs/haikuattrs.py
+++ b/plugins/haikuattrs/haikuattrs.py
@@ -20,7 +20,7 @@
 PLUGIN_NAME = 'Haiku BFS Attributes'
 PLUGIN_AUTHOR = 'Philipp Wolfer'
 PLUGIN_DESCRIPTION = 'Save and load metadata to/from Haiku BFS attributes.'
-PLUGIN_VERSION = "1.2"
+PLUGIN_VERSION = "1.3"
 PLUGIN_API_VERSIONS = ["2.2", "2.3", "2.4", "2.5", "2.6"]
 PLUGIN_LICENSE = "GPL-2.0-or-later"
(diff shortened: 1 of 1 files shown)


phw:
> Could alternatively use: `if sys.platform[:5] == 'haiku':`.

Just for me to understand, since I always ran the nightly versions: Will users of the current beta 5 also get the latest Python updates from depot, and hence also see "haiku"? If so I see no problem in changing the check to just "haiku". But otherwise a more generic check  (`sys.platform[:5] == 'haiku'` or `sys.platform in {'haiku', 'haiku1'}`) might be better.

OscarL:
Hello. Users that install beta5, if they do not do any system update, will get an older version of Python 3.10 where `sys.platform` will return 'haiku1'.

If they run SystemUpdate, or `pkgman update`, they will get Python 3.10.20 (both on beta5 and on nightlies), where 'haiku' will be the return of `sys.platform`.

So, for now, the safest option (at least until Haiku both switches to a newer Python default and releases a new beta) would be indeed using `sys.platform[:5] == 'haiku'` (or the "equivalent" `sys.platform.startswith('haiku')`). In fact, we used to use both of those in our patches of the Python interpreters themselves previous to these changes.

In any case, sorry for the inconvenience, and thanks for having Haiku support on your project!

phw (changes requested):
As discussed on matrix, let's change it to `sys.platform[:5] == 'haiku'` for the plugin, since this will be offered as an update.

Please also bump the plugin version.

phw (approved):
Thanks, LGTM

PR #436, opened 2026-09-15
Add clapback: look up, name and contribute CLAP embeddings; ask what sounds like a track

jeffcrouse:
Adds the **Clapback** plugin (MIT; Picard 2.6–2.13).

Picard fingerprints every file it scans and knows its MusicBrainz recording id — exactly the pair the [clapback](https://clapback.seethroughlab.com) commons keys on: a public corpus of CLAP audio embeddings (512-float vectors describing what a recording sounds like), keyed on the SHA256 of the AcoustID fingerprint, with recording ids attached by the contributors who hold the audio. It is the fetch-and-submit pair the AcousticBrainz plugins were, for a payload that can be reconciled between contributors.

Two context-menu actions on tracks and files:
- **Clapback: look up in the commons…** — per file, whether the commons holds the recording (`~clapback_status` in the metadata panel). With *Contribute* on, a held row is **named** with the recording id Picard already has; an absent one is embedded and contributed only if `clapback-embed` is installed.
- **Clapback: what sounds like this…** — the nearest recordings across every library that has plugged in, with MusicBrainz links for the named and hashes for the rest.

**Lookup-only by default.** A bundled Picard cannot `pip install`, so the plugin carries a verbatim copy of the stdlib-only `clapback-client` and computes nothing without an embedder — it says so in its options page. Files without a fingerprint go through Picard's own fpcalc.

Nothing leaves the machine until *Contribute* is turned on: then a one-way fingerprint hash, a vector if one was computed here, and the MusicBrainz recording id. Never audio, paths or other tags. The options page says this where the switch is.

Tested against Picard 2.13.3 headless (registrations, options round-trip, installing the zip through `PluginManager`); source and tests: <https://github.com/seethroughlab/clapback/tree/main/packages/picard-clapback>.


Sophist-UK:
Wow - this sounds like a VERY clever way to do automated DJing.

Perhaps MB itself should start to store the ClapBack values.

jeffcrouse:
Thanks! Automated DJing is one thing you could build on it. The plugin itself just does three things: checks whether the commons already has a recording's embedding, attaches the MusicBrainz recording id Picard already knows, and answers "what sounds like this" across everything that's been contributed. Playlists and the like are for clients to build on top.

On MB storing the values: MetaBrainz did run something close to this as AcousticBrainz, and (my understanding) shut it down because the stored estimates (BPM, key, mood) weren't worth keeping. The thing I've tried to do differently is store the embedding itself, keyed on the fingerprint hash plus a string naming the exact pipeline that produced it, so two contributors can be checked against each other and a pipeline change doesn't silently poison the pool. If MB ever wanted to ingest it, the key is designed so they could, but that's a conversation for once more than one library has contributed, not for this PR.

